Issue #29, opened 2020-02-13
Extension fails to build with musl libc


adamyeats:
As mentioned in https://github.com/appsignal/appsignal-agent/pull/484 (private issue), the extension is currently unable to build against musl libc.

This affects all users installing in an Alpine Linux environment.

xadamy:
As a workaround, users can add the following to their `Dockerfile`, which will allow the extension to build on Alpine Linux:

```
FROM node:alpine

RUN apk add --no-cache --virtual .gyp python make g++ \
    && npm install [ your npm dependencies here ] \
    && apk del .gyp
```

See https://github.com/nodejs/docker-node/blob/master/docs/BestPractices.md#node-gyp-alpine.

A proper fix for this is incoming.

tomaspinho:
Hi @xadamy . Just encountered this issue. The `appsignal-agent` shared library cannot be loaded in an Alpine Linux container because there are missing functions from glibc. Could we have a version of the agent that works with musl?

xadamy:
@tomaspinho The agent already has a `musl` build, but we currently have an issue with the extension (the C++ bridge between the integration and the agent) that prevents it from compiling with `musl`. We're looking into it but I can't give a timeframe currently, however it should still work if you follow the instructions [above](https://github.com/appsignal/appsignal-nodejs/issues/29#issuecomment-586186959). Does this work for you?

tomaspinho:
The extension builds correctly, however for some reason the binary downloaded from your servers is built against glibc, not musl. So the extension fails loading it as a shared library on runtime.

xadamy:
@tomaspinho Ah! In which case, there might be something wrong with our installation script. Could you try setting an environment variable for me: can you set `APPSIGNAL_BUILD_FOR_MUSL` to `true` before the package is installed, and let me know whether this gets you the right binary?

tomaspinho:
Just debugged this, yes, there's a bug in `getMetadataForTarget` in `packages/nodejs/scripts/extension.js`: the key `musl_override` should actually be `muslOverride`, as defined in `packages/nodejs/scripts/report.js`.

tomaspinho:
@xadamy please see https://github.com/appsignal/appsignal-nodejs/pull/91 for a fix

tomaspinho:
Also, I would suggest considering https://github.com/appsignal/appsignal-nodejs/pull/92/files to take care of musl installations automatically instead of relying on user override

[...]

PR #91, opened 2020-04-10
Fix for bug reported in #29

tomaspinho:
Fixes bad key lookup 'musl_override' in a report while installing the extension. Standardizes keys in report to be all camelcase.


xadamy:
👍 Thanks @tomaspinho !

PR #92, opened 2020-04-10
Figure out if using musl libc instead of depending on APPSIGNAL_BUILD_FOR_MUSL

tomaspinho:


packages/nodejs-ext/scripts/report.js
```diff
@@ -11,8 +12,8 @@ function createReport() {
       packagePath: path.join(__dirname, "/../ext/"),
       architecture: process.arch,
       target: process.platform,
-      muslOverride: Boolean(process.env["APPSIGNAL_BUILD_FOR_MUSL"]),
-      library_type: "static"
+      muslOverride: hasMusl(),
```
xadamy, line 15:
```suggestion
      muslOverride: (process.env["APPSIGNAL_BUILD_FOR_MUSL"] === "true") || hasMusl(),
```

- Saw a bug in my own code here, `Boolean(process.env["APPSIGNAL_BUILD_FOR_MUSL"])` converts `"false"` to `true`, so now this will give the correct result.
- I think it's a good idea that we still have the `"APPSIGNAL_BUILD_FOR_MUSL"` env variable as an override just in case, to keep it consistent with the other integrations. This change will ensure `hasMusl()` is called when `"APPSIGNAL_BUILD_FOR_MUSL"` is not set. Does that seem reasonable to you?

packages/nodejs-ext/scripts/extension/helpers.js
```diff
@@ -28,4 +28,13 @@ function hasSupportedArchitecture(report) {
...
 }
 
-module.exports = { hasLocalBuild, hasSupportedArchitecture }
+/**
+ * Returns `true` if the current linux system is using musl as its libc and vice versa
+ *
+ * @return  {boolean}
+ */
+function hasMusl() {
+  return /musl/.test(child_process.spawnSync("ldd", ["--version"]).stderr);
```
xadamy, line 37: This currently throws as `child_process` is not imported from anywhere. Can you please add the necessary import to the top of the file?

tomaspinho:
Addressed all of your points and rebased on top of your master :)

xadamy:
Thanks @tomaspinho ! Merging 🚀

PR #98, opened 2020-04-14
Move CI to Semaphore

### LINKED ISSUE #83: Move CI from Travis to Semaphore
We should keep this consistent with all the other current integrations.

xadamy:
Closes #83.


xadamy:
Pipeline currently fails as build artefacts are not carried over to the next "block" by default.

tombruijn:
I don't see a link to Semaphore in this PR. When I look it up manually I see the build has failed: https://appsignal.semaphoreci.com/workflows/185f2c9d-6e00-4370-9c4b-6dc561f6cbc1

xadamy:
@tombruijn That's because I turned builds off for this repo in Semaphore's control panel so it wouldn't prevent other things being merged (see above).

Do you know how to set the following up:
- Carrying build artefacts over to the next "block"?
- Running the build and tests on every major version of `node`? e.g. latest `10`, latest `11`, latest `12`, latest `13`?

tombruijn:
> Carrying build artefacts over to the next "block"?

if it's limited to the same build, use the cache instead. Artifacts are for things to use after the build is done. Which I haven't used myself yet. If you can't access artifacts in the same build, then also store it in the cache so you can access it through the cache within the build, and through artifacts after.

> Running the build and tests on every major version of node? e.g. latest 10, latest 11, latest 12, latest 13?

You can set up a matrix for that: https://docs.semaphoreci.com/essentials/build-matrix/

I've written my own matrix for the Ruby gem build here because it was a bit too limited for all the exceptions that occur in the Ruby gem build (Ruby, Rubygems and Ruby gem versions that don't work well together): https://github.com/appsignal/appsignal-ruby/blob/5a4a10343552fa2652cd45ab8b67f83cdc3a784a/build_matrix.yml#L77-L166
which generates: https://github.com/appsignal/appsignal-ruby/blob/5a4a10343552fa2652cd45ab8b67f83cdc3a784a/.semaphore/semaphore.yml#L46-L964

xadamy:
This is now ready for review!

.semaphore/semaphore.yml
```diff
@@ -0,0 +1,52 @@
...
+agent:
+  machine:
+    type: e1-standard-2
+    os_image: ubuntu1804
+auto_cancel:
+  running:
+    when: branch != 'master' AND branch != 'develop'
+global_job_config:
+  env_vars:
+  - name: RUNNING_IN_CI
```
tombruijn, line 12: if we don't use this env var we can remove it.

PR #123, opened 2020-05-03
Postgres integration

### LINKED ISSUE #51: Postgres integration
https://github.com/brianc/node-postgres

xadamy:
Closes #51.


xadamy:
> would not having a function like appsignal_record_event and wrapping the event suffice, or would you switch to a function where you could just set the event time if it's measured in the Postgres library and the function exists in the agent?

Apologies, my conceptual model of this was different, so I continued this in the [other thread](https://github.com/appsignal/appsignal-agent/issues/511).

> Also, I'd love to see a test that makes sure a basic example of a query is handled correctly, to keep an eye on possible regressions in the future.

Yes! Let me add an issue in for this - I think we can get a really nice integration test working for this on Semaphore, but this might take a bit of time to get set up.

PR #130, opened 2020-05-06
Add Next.js integration

### LINKED ISSUE #100: Next.js integration
I've been doing some work on getting this to integrate with Next.js.

We were using a custom server with Node's default HTTP server implementation `http` as described in https://nextjs.org/docs/advanced-features/custom-server . However, this wasn't correctly capturing the route paths, naming all spans as `GET [unknown route]`. Moved to using a custom `express` server instead.

Had to change the default express instrumentation a bit, so made our own middleware, like so:

```
const { parse } = require('url');

function expressMiddleware(appsignal) {
  return function (req, res, next) {
    const tracer = appsignal.tracer();
    const rootSpan = tracer.currentSpan();

    if (!rootSpan) {
      return next();
    }

    if (req.params.password) {
      rootSpan.setSampleData('params', {
        ...req.params,
        password: '[FILTERED]',
      });
    } else {
      rootSpan.setSampleData('params', req.params);
    }

    return tracer.withSpan(rootSpan, (span) => {
      const originalEnd = res.end;

      tracer.wrapEmitter(req);
      tracer.wrapEmitter(res);

      res.end = function (...args) {
        res.end = originalEnd;

        const { method = 'GET' } = req;

        // if there is no error passed to `next()`, the span name will
        // be updated to match the current path
        span.setName(`${method} ${parse(req.originalUrl).pathname}`);

        span.set('method', method).set('status_code', res.statusCode);
        span.set('qs', req.query);

        return res.end.apply(this, args);
      };

      return next();
    });
  };
}

function expressErrorHandler(appsignal) {
  return function (err, req, res, next) {
    const span = appsignal.tracer().currentSpan();

    if (!span) {
      return next();
    }

    // if there's no `status` property, forward the error
    // we also ignore client errors here
    if (err && (!err.status || (err.status && err.status >= 500))) {
      span.addError(err);
    }

    return next(err);
  };
}

module.exports = {
  expressMiddleware,
  expressErrorHandler,
};
```

Relevant changes are the insertion of the querystring arguments as part of the span and usage of `parse(req.originalUrl).pathname` instead of `req.route.path`. `req.route.path` is actually the path of the route that matches, not the path portion of the original URL. Because Next.js implements its own handler function, we must match on all `*` routes and pass that on to the default handler function. See https://github.com/zeit/next.js/blob/canary/examples/custom-server-express/server.js for an example.

Right now, we have spans in AppSignal with request duration, however can't find the span metadata anywhere. Also, the span itself has no drilldown available, listing a single `unknown` event.

xadamy:
Closes #100.

cc:/ @tomaspinho


tomaspinho:
Wow this is great! Thanks a ton @xadamy. Will test this out in our app shortly.

PR #148, opened 2020-05-14
Add experimental Web Vitals support for Next.js

xadamy:
Adds experimental support for Next.js v9.4 Web Vitals reporting. This works by providing an Express middleware which will mount a route to `/__appsignal-web-vitals` and `POST` the metrics to that, and we’ll add custom metrics for these metrics on the server.

This works by exporting a `reportWebVitals` function from `pages/_app.js` as described [here](https://nextjs.org/blog/next-9-4#integrated-web-vitals-reporting).   

```
export function reportWebVitals(metric) {
  const body = JSON.stringify(metric)
  const url = '/__appsignal-web-vitals'

  // Use `navigator.sendBeacon()` if available, falling back to `fetch()`.
  ;(navigator.sendBeacon && navigator.sendBeacon(url, body)) ||
    fetch(url, { body, method: 'POST', keepalive: true })
}
```

On successful setup, two new magic dashboards will be created for the user - Next.js and Next.js Web Vitals.


xadamy:
See https://github.com/appsignal/public_config/pull/18 for magic dashboard config.

PR #159, opened 2020-05-18
Refactor span, accept context via span args

xadamy:
This is a ⚠️ breaking change.

This PR refactors some of the Span logic to move it closer to the OpenTelemetry behaviour as well as allowing https://github.com/appsignal/appsignal-agent/pull/524 to be implemented.

- Instead of taking a `namespace` argument in the first position, the `Span` now takes a `SpanOptions` object with optional `name` and `startTime` properties.
- `Span`s can now take an optional `SpanContext` as the second parameter,  ready for https://github.com/appsignal/appsignal-agent/issues/531
- `currentSpan()`  now returns a `NoopSpan` instead of `undefined` if no current `Span` is available.
- Remove incorrect implementation of `public setSQL(keyOrValue: string, value?: string)`, this should only be unary.

### DIFF
--- a/packages/nodejs/src/__tests__/scope.test.ts
+++ b/packages/nodejs/src/__tests__/scope.test.ts
@@ -40,7 +40,7 @@ describe("ScopeManager", () => {
     })
 
     it("should run the callback (object as target)", done => {
-      const test = new RootSpan("test")
+      const test = new RootSpan({ namespace: "test" })
 
       scopeManager.withContext(test, () => {
         expect(scopeManager.active()).toStrictEqual(test)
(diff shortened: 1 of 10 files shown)


xadamy:
@jeffkreeftmeijer If you ware still looking to implement appsignal/appsignal-agent#524, I think the best thing to do would be to work from this branch and commit directly to it. Please ping me if you need anything from me, or would like me to do it on your behalf.

jeffkreeftmeijer:
Great, thanks. As discussed in the integrations call; let's see where we're at at the beginning of next week.

PR #251, opened 2020-08-31
Diagnose tool

### LINKED ISSUE #24: Diagnose function
This was skipped in the initial implementation as the integration as the log output would always output an error even though the agent had started successfully.

xadamy:
Closes #24.

This adds the long-awaited diagnose CLI. In a new app, this can now be used by running `npx appsignal-diagnose` from your project directory after `@appsignal/nodejs` is installed.

As `@appsignal/nodejs` doesn't use a config file, it relies on the correct environment variables being set at the time. If unset, you can use the command like this: `APPSIGNAL_PUSH_API_KEY="<key>" npx appsignal-diagnose`

At the moment, the `paths` key has been omitted as this might take a little extra time.


thijsc:
Looking good so far! I see it's still in draft state, do you expect me to do a review already?

xadamy:
@thijsc This is now ready for review.

packages/nodejs/bin/diagnose
```diff
@@ -0,0 +1,52 @@
...
+const tool = new DiagnoseTool({})
+
+console.log(`
+🔧 AppSignal Diagnose Tool
+
+Use this information to debug your configuration.
+More information is available on the documentation site.
+https://docs.appsignal.com/
+
+Send this output to support@appsignal.com if you need help.
```
thijsc, line 18: Since we post the result I suggest this text:
```suggestion
This diagnose output was sent to AppSignal, contact us at support@appsignal.com if you need help.
```

thijsc (approved):
Made a suggestion on the text, otherwise good to go.

PR #254, opened 2020-09-02
Move dependency management from yarn to npm

### LINKED ISSUE #40: Remove further references to yarn
As mentioned in #32, there are several hardcoded references to `yarn` which will fail if the user doesn't have it installed. These should be made more generic to cover both `npm` and `yarn`.

xadamy:
Fixes #40.

During development of the diagnose function, I noticed that `yarn` was missing a lot of functionality needed to test nodejs CLI binaries. As this also fixes #40, I have decided to migrate the project from `yarn` to `npm`. 

This gives us some better developer ergonomics and makes us more able to accept contributions, at the cost of losing the yarn workspaces feature, which we will regain again once `npm` v7 is out. This means that each package needs a lockfile of it's own going forward.

(@thijsc Yes, I have checked, and you need to check the lockfiles into git.)


thijsc:
> (@thijsc Yes, I have checked, and you need to check the lockfiles into git.)

This was indeed the first question on my mind :-). I still don't understand what benefit checking them in has, can you explain why it's needed?

xadamy:
@thijsc 

> This file **is intended to be committed into source repositories**, and serves various purposes:
>- Describe a single representation of a dependency tree such that teammates, deployments, and continuous integration are guaranteed to install exactly the same dependencies.
> - Provide a facility for users to “time-travel” to previous states of node_modules without having to commit the directory itself.
> - To facilitate greater visibility of tree changes through readable source control diffs.
> - And optimize the installation process by allowing npm to skip repeated metadata resolutions for previously-installed packages.

https://docs.npmjs.com/configuring-npm/package-lock-json.html

thijsc:
These arguments hold a lot of water for applications, which is why for applications we always check in lock files. I don't see the argument for libraries though.

In the case of a library we generally don't want to specify exact dependencies, but a range of dependencies that we expect to be compatible with. We don't want to be in a position where a very specific version lock actually prevents security updates from being installed when a user runs npm update, but we haven't shipped a new version with a specific version update yet.

In addition to this, the lock file isn't used for dependency resolution. It's more of an internal implementation detail on what we run our tests against. I do think checking in the file has downsides, we have an endless stream of dependency updates in our commit log that make it harder to keep an overview of changes.

Overall I don't see a clear and cut case why checking in the lock file is necessary, and given the downsides it's not my preference to do so.

xadamy:
Reverted the change as removing the lockfiles broke CI. Merging this once the test suite has passed. @thijsc You can open an issue on this repo if you want to discuss this further.

thijsc:
I think the depfu configuration change you made might well work to make the pull requests list less cluttered. If that's about one per week I think no further changes are necessary.

PR #272, opened 2020-09-15
Set Express route params as `params` sample data

xadamy:
Fixes a recent support query.

This appends the parsed dynamic route parameters (`req.params`) and query string (`req.query`, if it exists) from Express to the root span as sample data. 

This required a small refactor to the data model - rather than exposing two separate classes for `DataMap` and `DataArray`, I have opted to more closely match the Ruby implementation. My thinking at the time was that it might be useful if those objects could be introspect-able, but on reflection, I don't see much need for this.

This also closes https://github.com/appsignal/appsignal-agent/issues/619 by using a filtered map for sample data, allowing data to be filtered by the agent, and adding associated config for this.


packages/express/src/middleware/index.ts
```diff
@@ -33,14 +34,21 @@ export function expressMiddleware(appsignal: Appsignal): RequestHandler {
...
         // if there is no error passed to `next()`, the span name will
         // be updated to match the current path
         if (req.route?.path) {
           span.setName(`${method} ${req.route.path}`)
         }
 
+        // set route params (if parsed by express correctly)
+        span.setSampleData("params", {
+          ...params,
+          query: query || ""
```
jeffkreeftmeijer, line 49: Would this include the whole, unfiltered query string? If so, this could include sensitive or identifiable information we can't easily filter out. Do we get the GET params in the main `params` variable as well? Since that's filtered by the agent, that might be the safer place for it.

jeffkreeftmeijer (changes requested):

packages/express/src/middleware/index.ts
thijsc: We should indeed use the filtered data variant here. Do we?

xadamy: Thanks @jeffkreeftmeijer @jeffkreeftmeijer, forgot about the data sanitization here. I have now changed this to combine `req.params` and `req.query` together as params, which needed a small refactor. Please re-review.

thijsc (approved):

packages/express/src/middleware/index.ts
jeffkreeftmeijer: Awesome, looking good!

jeffkreeftmeijer (approved):

PR #283, opened 2020-10-02
Add ignoreInstrumentation option

### LINKED ISSUE #60: Ignoring a core Node instrumentation
The client should be able to turn instrumentation off for internal Node.js modules if the user doesn't require it. This could be handled via a config option.

```js
const { Appsignal } = require("@appsignal/nodejs")

  active: true,
  name: "<YOUR APPLICATION NAME>"
  apiKey: "<YOUR API KEY>",
  ignoreInstrumentation: ["http"]
})
```

xadamy:
Closes #60. Adds a `ignoreInstrumentation` config option to ignore an automatic instrumentation. It can be used like this:

```
const { Appsignal } = require("@appsignal/nodejs")

  active: true,
  name: "<YOUR APPLICATION NAME>"
  apiKey: "<YOUR API KEY>",
  ignoreInstrumentation: ["http"]
})
```

...which will ignore the `http` instrumentation. The string must match the name of the module as passed to `require()`. This also fixes a bug I noticed where active instrumentation was not added to the `instrumentation.active` array.

### DIFF
--- a/packages/nodejs/src/instrument.ts
+++ b/packages/nodejs/src/instrument.ts
@@ -6,14 +6,14 @@ import { Plugin } from "./interfaces/plugin"
 import { Tracer } from "./interfaces/tracer"
 import { Metrics } from "./interfaces/metrics"
 
-type InstrumentedModule<T> = { name: string; plugin: Plugin<T>; hook: Hook }
+type InstrumentedModule = { name: string; hook: Hook }
 
 /**
  * The Instrumentation class.
  * @class
  */
 export class Instrumentation {
-  public active: InstrumentedModule<any>[]
+  public active: InstrumentedModule[]
 
   private _tracer: Tracer
   private _meter: Metrics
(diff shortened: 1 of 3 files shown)


packages/nodejs/src/instrument.ts
```diff
@@ -53,55 +51,45 @@ export class Instrumentation {
...
    */
-  public unload(name: string): InstrumentedModule<any>[] {
+  public unload(name: string): void {
     this.active = this.active.filter(active => {
       if (active.name !== name) {
         return true
       } else {
-        const { plugin, hook } = active
-
-        plugin.uninstall()
```
xadamy, line 82: Removed this call as `hook.unhook()` removes the instrumentation altogether from the `require()` cache, making unpatching it unnecessary.

jeffkreeftmeijer (approved):

matsimitsu (approved):

thijsc (approved):

Issue #317, opened 2020-11-05
Next.js hot-reload routes are not ignored


afsanefda:
I have the following code for my nextjs application which has a custom serverjs file:

* server.js - some parts are removed. if something important is missing let me know!

```
/* Appsignal should be initialized before anything else */
const { Appsignal } = require('@appsignal/nodejs');
const { expressErrorHandler , expressMiddleware  } = require('@appsignal/express');

const appsignalssr =  new Appsignal({
  active: true,
  name: process.env.APPSIGNAL_APP_NAME,
  environment: process.env.APPSIGNAL_ENV,
  apiKey: process.env.APPSIGNAL_SSR_KEY,
  log: 'stdout'
});

(async () => {
  await app.prepare();
  const server = express();

if (appsignalssr) {
    /* for setting request route and method in appsignal */
    server.use((req, res, next) => {
      const tracer = appsignalssr.tracer();
      const span = tracer.createSpan({
        namespace: 'web',
        startTime: Date.now()
      });
      span.setName(`${req.method} ${req.url}`);
      span.close();

      next();
    });
  }

  server.use(handler);


  server.get('*', (req, res) => handle(req, res));

  if (appsignalssr) {
    /* should use at the end */
    server.use(expressMiddleware(appsignalssr));
    server.use(expressErrorHandler(appsignalssr));
  }

  server.use((error, req, res, next) => {
    /* eslint-disable */
    if (!error.statusCode) error.statusCode = 500;
    if (!error.name) error.name = 'Express server error!';
    if (!error.message) error.message = 'An Error occurred on express server!';
    /* eslint-enable */

    const tracer = appsignalssr.tracer();
    const span = tracer.createSpan({
      namespace: 'web',
      startTime: Date.now()
    });

    span.addError(error);
    span.close();

    next();
  });

  await server.listen(port);
  console.log(`> Ready on http://localhost:${port}`); // eslint-disable-line no-console
})
```

The problem is when I run this locally I get the route and method but when it is deployed on QA the routes are unknown:

**localhost:** 

<img width="552" alt="Screen Shot 2020-11-05 at 21 42 36" src="https://user-images.githubusercontent.com/37764909/98281602-3217ea80-1fb2-11eb-903c-ca990da870b4.png">

**QA:**

[...]

PR #320, opened 2020-11-09
Add option to disable calling setSampleData

TheMightyPenguin:
Adds an option to `expressMiddleware` to disable calling `setSampleData` and settings `"params"`.

## Context

In our NextJS application, we're using both the NextJS middleware and the ExpressJS middleware for Appsignal. In our app, we're manually calling `span.setSampleData("params", ...)` and this calls gets overriden by the ExpressJS middleware as it does a monkey patch over `res.end`, and in that way its call to `span.setSampleData` always win and we end up with empty Parameters on the Appsignal dashboard.

I tried our code with the library call to `setSampleData` disabled, and then we were able to get our parameters in the dashboard.

Not sure if this extra option is the right solution for the problem we have, but happy to discuss this or other approaches 😄  Thanks in advance!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1780,6 +1780,30 @@
         }
       }
     },
+    "@jest/create-cache-key-function": {
+      "version": "26.6.2",
+      "resolved": "https://registry.npmjs.org/@jest/create-cache-key-function/-/create-cache-key-function-26.6.2.tgz",
+      "integrity": "sha512-LgEuqU1f/7WEIPYqwLPIvvHuc1sB6gMVbT6zWhin3txYUNYK/kGQrC1F2WR4gR34YlI9bBtViTm5z98RqVZAaw==",
+      "dev": true,
+      "requires": {
+        "@jest/types": "^26.6.2"
+      },
+      "dependencies": {
+        "@jest/types": {
+          "version": "26.6.2",
+          "resolved": "https://registry.npmjs.org/@jest/types/-/types-26.6.2.tgz",
+          "integrity": "sha512-fC6QCp7Sc5sX6g8Tvbmj4XUTbyrik0akgRy03yjXbQaBWWNWGE7SGtJk98m0N8nzegD/7SggrUlivxo5ax4KWQ==",
+          "dev": true,
+          "requires": {
+            "@types/istanbul-lib-coverage": "^2.0.0",
+            "@types/istanbul-reports": "^3.0.0",
+            "@types/node": "*",
+            "@types/yargs": "^15.0.0",
+            "chalk": "^4.0.0"
+          }
+        }
+      }
+    },
     "@jest/environment": {
       "version": "26.5.2",
       "resolved": "https://registry.npmjs.org/@jest/environment/-/environment-26.5.2.tgz",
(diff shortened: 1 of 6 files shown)


xadamy:
@TheMightyPenguin Thanks for the PR! It would help me to know more about what you're trying to do here. When you call `span.setSampleData("params", ...)` again in your code, it should overwrite any keys that are in the hashmap already with the same name, but any new keys should remain. It should not matter who "wins" here if you are using new keys. If that's not the case, that is a bug in the agent.

It would help me to know:

- Are you overwriting any keys to the `params` hash that are present already?
- What data are you trying to write to `params`?

TheMightyPenguin:
> Are you overwriting any keys to the params hash that are present already?
Don't think so actually, as when the line I added the parameter for runs both of those values seem to be empty.

> What data are you trying to write to params?
The URL dynamic params (e.g.: we have a route like `magazine/article/:slug`, and we want to record this dynamic `slug`), and also any query paramer. I know this library already records those, but we're using a third-party library for routing, and our parameters aren't available as normally with Express 😅 .

For extra context, this is the custom middleware we have at the moment:

```js
const nextJsAppsignalName = (req, res, next) => {
  const tracer = appsignal.tracer();
  const rootSpan = tracer.currentSpan();
  const { pathname } = url.parse(req.url || '/', true);

  if (rootSpan) {
    const routes = getRoutes();
    const matched = routes.filter((el) => el.match(pathname))[0];

    if (matched) {
      const params = matched.match(pathname);
      rootSpan.setName(`${req.method} ${matched.page}`);
      rootSpan.setSampleData('params', { ...params, ...req.query });
    } else if (!pathname.startsWith('/_next')) {
      rootSpan.setName(`${req.method} ${pathname}`);
    }
    rootSpan.set('pathname', pathname);
  }

  return next();
};
```

Not sure if this is actually missing anything, or if we should actually be using `tracer.withSpan` 🤔

xadamy:
@TheMightyPenguin Thank you for this context! Give me a little while to take a look at this, I think you may have uncovered an interesting bug here but I'll need a little time to reproduce this locally. I'll get back to you on this ASAP!

FYI: also, I see that you are calling `pathname.startsWith('/_next')`, which leads me to think you're seeing another bug as outlined [here](https://github.com/appsignal/appsignal-nodejs/issues/317), which is something I'm looking into separately.

TheMightyPenguin:
@xadamy got it, thank you 👍  and yes, callign `pathname.startsWith('/_next')` this is related to the bug you mentioned!

xadamy:
Update: I've spent some time today looking at this, but it looks like an upstream change in `next-routes-with-locale@1.8.5` has broken things in my test setup, which is currently running `next@10.0.3`.

@TheMightyPenguin If you catch this before it gets fixed, I'm wondering if you could provide me with your current `next` and `next-routes-with-locale` version numbers? Thanks!

TheMightyPenguin:
@xadamy we're using `next-routes-with-locale@1.8.5` with `next@9.4.0`! Sorry for the late response, just saw this

TheMightyPenguin:
@xadamy any update here? We're still having this bug and it's a bit cumbersome to not be able to see the params we record in the dashboard 😅

thijsc:
We've created a very basic Next.js test app: https://github.com/appsignal/test-setups/tree/main/nodejs/next-js

I'm looking into what the best way forward is here. Could you give me an example of how you're using Next here? Seems like it goes beyond basic static page rendering?

thijsc:
I'm closing this one for now. We will make a change in our API in the future to accommodate this scenario.

PR #321, opened 2020-11-09
Extract function to set span name out of middleware

TheMightyPenguin:
Extracts the logic of setting the span name to a separate function, and exports it for library consumers.

## Context

We have a NextJS application, and we faced 2 challenges when integrating Appsignal for monitoring.

First, we use a third party library for defining routes: https://github.com/vonschau/next-routes-with-locale, we use this to internationalize routes, and once we integrated Appsignal for performance monitoring, all requests were labeled as:
```
${METHOD} [unknown route]
```

After some digging around, we found out that in our case `app.router.dynamicRoutes` was an empty array, and this is most likely as we use the library I mentioned, so we had the routes in our app, but just in a different object, so in our case reading from `app.router.dynamicRoutes` won't work.

Second, we could not use the NextJS middleware (`getRequestHandler`) as `next-routes-with-locale` also relies on calling the request handler from NextJS: https://github.com/vonschau/next-routes-with-locale/blob/56ada8d2b71fb63f74fc5b8e0a07e911f09f39fa/src/index.js#L131-L150, so this made this middleware (and the other one too), quite hard to compose together. Ideally for us, this wouldn't call the NextJS request handler, but I don't know if there's any specific reason why it does.

To solve both of these, we extracted the logic to set the span name into a separate function, and this function takes the `routes` as a parameter, this way we can use this function inside a middleware in our app and get the expected result.

Again, I don't know if this is the best solution, but happy to discuss more approaches! and if more context is needed pleast let me know, thanks in advance!

### DIFF
--- a/packages/nextjs/src/handler.ts
+++ b/packages/nextjs/src/handler.ts
@@ -2,20 +2,56 @@ import url from "url"
 import { Appsignal } from "@appsignal/nodejs"
 import { ServerResponse, IncomingMessage } from "http"
 
+interface Route {
+  page: string
+  match: (
+    pathname: string | null | undefined
+  ) => false | { [param: string]: any }
+}
+
 interface NextServer {
   router: {
-    dynamicRoutes?: {
-      page: string
-      match: (
-        pathname: string | null | undefined
-      ) => false | { [param: string]: any }
-    }[]
+    dynamicRoutes?: Route[]
   }
   getRequestHandler(): (
     req: IncomingMessage,
     res: ServerResponse,
     ...rest: any[]
   ) => Promise<void>
+}
+
+export function setSpanName(
+  appsignal: Appsignal,
+  req: IncomingMessage,
+  routes: Route[] = []
+) {
+  const span = appsignal.tracer().currentSpan()
+  const { pathname } = url.parse(req.url || "/", true)
+
+  if (span) {
+    const matched = routes.filter(el => el.match(pathname))[0]
+
+    // passing { debug: true } to the `Appsignal` constructor will log
+    // data about the current route to the console. don't rely on this
+    // working in future!
+    if (appsignal.config.debug) {
+      console.log("[APPSIGNAL]: Next.js debug data", {
+        routes,
+        matched,
+        pathname
+      })
+    }
+
+    if (matched) {
+      // matched to a dynamic route
+      span.setName(`${req.method} ${matched.page}`)
+    } else if (!matched && pathname === "/") {
+      // the root
+      span.setName(`${req.method} ${pathname}`)
+    } else {
+      span.setName(`${req.method} [unknown route]`)
+    }
+  }
 }
 
 /**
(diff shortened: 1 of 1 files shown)


xadamy:
@TheMightyPenguin Thanks for this PR too! :) Again, with this one, it would help to know more about your use case, as it's not clear to me what this would solve that calling `tracer.currectSpan().setName()` wouldn't solve. Can you tell me why this approach wouldn't work for you?

TheMightyPenguin:
@xadamy thanks for the fast response 😄  at the moment this middleware is not useful for us as it always resolves to:

```
GET [unknown route]
```

This is because in our case `app.router.dynamicRoutes` always resolves to an empty array, as we use a third-party library for managing routes.

As you mentioned, yes we could totally write some custom middleware using `tracer.currentSpan().setName()` and that would solve our issue, but we wanted to try and use the library and also make a PR as maybe other teams had similar issues, and to contribute back to the community (in case it makes sense and more people can benefit from this).

We also understand if this change doesn't make sense for you, as we can also go with the custom middleware approach.

xadamy:
@TheMightyPenguin Understood! 👍 

I think, for this particular PR, it wouldn't make a lot of sense to expose this functionality in this way. Using `tracer.currentSpan().setName()` in a custom middleware is more what I had in mind for doing things like this - I think that exposing this functionality as a public function makes some internals public that I'd prefer to keep private, and making that logic public might break forwards-compatibility in some interesting ways further down the road. I think it's best to leave this "as-is" for now.

But! I thank you for the effort ❤️ I very much appreciate your thoughtful reasoning here! Closing this PR for now.

TheMightyPenguin:
@xadamy Got it, that makes sense, thank you! 😄  For now we won't use the NextJS plugin, and make a custom middleware as we do need to get the routes from a different object.

Just one more question, does this middleware need to call `app.getRequestHandler()`? just wondering as that specific aspect made it a bit complicated to compose, why cannot it call `next()` like other middlewares do? (asking out of curiosity)

xadamy:
@TheMightyPenguin 

> does this middleware need to call app.getRequestHandler()?

If you're creating a custom Next.js middleware and not using ours, then you'll have to call the handler function returned by `const handler = app.getRequestHandler()` at the end of the middleware. Feel free to take a look at how I did this in our middleware [here](https://github.com/appsignal/appsignal-nodejs/blob/main/packages/nextjs/src/handler.ts).

> why cannot it call next() like other middlewares do?

That's a good question! We do call `next()` inside the Express middleware, but not inside the Next.js middleware, indeed. I think this is the only way I could get this to work for users that weren't using Express and are using, for example, just the default `http` library (The function `next()` is, of course, an Express thing, which is a bit confusing when we're talking about Next.js too!). 

However, if you can think of a nicer way for us to make this more composable, then I'm keen to hear! Open a new issue and I'll be happy to consider the idea, I'm always happy to add a more composable API method here if it makes sense! 👍

PR #345, opened 2020-12-15
Add minutely probe for collecting heap stats

### LINKED ISSUE #23: General GC metrics from Node/V8
`appsignal_start_transaction` requires `long` parameter `gc_duration_ms`, which is documented as "the current garbage collection duration in milliseconds". 

Node.js does not expose this number in its public API, which means we may have to write further code in C/C++ to get those values from V8, and in fact this seems [like what others have done](https://github.com/NodeFly/node-nodefly-gcinfo/blob/master/src/gc.cc).

We can import `v8.h` from the same place as we import `node.h`, however, per the docs, we will likely lose ABI stability. This may or may not matter to us.

V8 also has [separate types of GC](http://jayconrod.com/posts/55/a-tour-of-v8-garbage-collection). There are [3 types of GC events](https://v8docs.nodesource.com/node-0.8/d2/dc3/namespacev8.html#a83be2143eef5fa2b17127fd69686496b) in V8:

- `kGCTypeMarkSweepCompact`
- `kGCTypeScavenge`
- `kGCTypeAll`

We will need to decide which of these events are relevant to our use case.

xadamy:
This PR adds basic heap stats support for V8, tuning a minutely probe to poll for changes to `v8.getHeapStatistics()`. Also includes a bug fix for an issue that I came across during dev, where if an `ignoreInstrumentation` array was passed, all instrumentation would be ignored.

Closes #23.


packages/nodejs/src/bootstrap.ts
```diff
@@ -0,0 +1,57 @@
...
+
+import { Instrumentation } from "./instrument"
+import { httpPlugin, httpsPlugin } from "./instrumentation/http"
+import * as pgPlugin from "./instrumentation/pg"
+import * as redisPlugin from "./instrumentation/redis"
+
+import * as gcProbe from "./probes/v8"
+
+/**
+ * Initialises all the avaiable core instrumentation.
```
tombruijn, line 11:
```suggestion
 * Initialises all the available core instrumentation.
```

tombruijn (approved):

jeffkreeftmeijer (approved):

PR #367, opened 2021-02-17
Use npm workspace

thijsc:
Use npm workspace. Adds a Makefile with some helpers.


tombruijn:
Unassigning my review for now. We're looking at the tooling.

thijsc:
I rebased and used this branch today because I could not get `main` work otherwise on my computer. Is it an option to merge this, and then adopt the new tooling right after?

tombruijn:
If we merge this we cannot release new versions of the libraries, so I'd rather not. It's annoying how it works now, but it's better than it not working at all. Yesterday I tried again to make this solution work alongside lerna but I could not make it work. Work on mono is next on my list, probably picking it up again tomorrow.

thijsc:
> It's annoying how it works now, but it's better than it not working at all.

Can't you still publish packages though? Just individually?

tombruijn:
Not without changing now how our dependencies depend on one another now. They are locked down to the patch release currently. It would require you to release all packages currently to update only one package. It's not possible to release a package individually currently.

thijsc:
>  It would require you to release all packages currently to update only one package

Right, yeah I think that's not needed with this approach. But if you're continuing with mono soon let's do that. I'll keep this branch rebased until then so we can actually have a working dev environment.

thijsc:
I just pushed cd084fbfd98b6749a0b9b65bf2d7d0af945294ff. It switches the repo to use an NPM workspace. I've now managed to have an actually working dev setup where you can change things in one package, and have that be picked up in the other ones.

packages/nodejs/package.json
```diff
@@ -26,7 +26,7 @@
     "typescript": "~> 4.1"
   },
   "optionalDependencies": {
-    "@appsignal/nodejs-ext": "~> 1.2"
+    "@appsignal/nodejs-ext": "=1.2.3-beta.0"
```
thijsc, line 29: I think it's best to specify the exact version we want here from now on. If there is a mismatch, for example the package is in beta, the linking files and it falls back to the version from the registry.

thijsc:
We need NPM 7 for this, asked Semaphore support what the best way to run this is.

thijsc:
The npm workspace works well for me locally, but giving some issues here. Think it's related to caching.

tombruijn:
Removing myself as a reviewer, I'm working on integrating mono using workspaces as well.

tombruijn:
Closing this in favor of #394

Issue #370, opened 2021-03-03
Diagnose report musl_override always returns true on musl systems


tombruijn:
The `musl_override` value of the report is only to indicate that a user _manually_ specified the installation of a musl build on the host system with `APPSIGNAL_BUILD_FOR_MUSL`.

It currently also checks if musl was detected by the system. This needs to be removed.

https://github.com/appsignal/appsignal-nodejs/blob/57c32683792e443f3345b838a3064a27e0dc5624/packages/nodejs-ext/scripts/report.js#L29-L30

Remove: ` || hasMusl()`

See if we can test this.

tombruijn:
I can't simply remove the `hasMusl` because the `getMetadataForTarget` function in `scripts/extension.js` uses the `musl_override` key to determine if it's musl or not.

PR #374, opened 2021-03-03
Fix musl target and override reporting

tombruijn:
The musl_override value in the diagnose reports if the user has manually
chosen the linux-musl build of the agent and extension, by specifying
`APPSIGNAL_BUILD_FOR_MUSL=true` as the environment variable. (Also
accepts the `1` value on other integrations, see
appsignal/integration-guide#26 .)

In this Node.js integration it currently is set to true if the host
system is detected as a musl system, not just when the user specifies
the `APPSIGNAL_BUILD_FOR_MUSL` environment variable.

This change fixes that, and moves the `-musl` suffix to the `target`
key, which is used to determine which file to download. The `-musl`
suffix should be part of the `target` key value, like it's done in other
integrations.

Fixes https://github.com/appsignal/appsignal-nodejs/issues/370

I've not implemented the fallback for musl builds on old libc versions
like is part of the other integrations, as that's outside of the scope
of this change.

### DIFF
--- a/packages/nodejs-ext/scripts/extension.js
+++ b/packages/nodejs-ext/scripts/extension.js
@@ -78,13 +78,11 @@ function dumpReport(report) {
 
 function getMetadataForTarget({
   architecture,
-  target,
-  musl_override: muslOverride
+  target
 }) {
   const triple = [
     architecture === "x64" ? "x86_64" : "i686",
-    `-${target}`,
-    muslOverride ? "-musl" : ""
+    `-${target}`
   ]
 
   return TRIPLES[triple.join("")]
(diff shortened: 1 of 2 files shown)


tombruijn:
Still a draft because I need to test it more in the test setups.

tombruijn:
Tested this works (with a lot of fiddling in the test setups for an alpine linux build, everything is bash and there's no bash on alpine linux). The package itself does not work on alpine linux, as per #29, but now it reports the target and musl override correctly.

thijsc (approved):

jeffkreeftmeijer (approved):

PR #383, opened 2021-04-26
Handle all non-dynamic paths as static routes

### LINKED ISSUE #337: Next.js integration is not compatile with API Routes
[Next.js API Routes](https://nextjs.org/docs/api-routes/introduction) create a Span correctly but do not update the spans name with the correct route name.

```
[APPSIGNAL]: Next.js debug data { routes: [], matched: undefined, pathname: '/api/hello' }
```

This means that the span name will always be `GET [unknown route]` for an API Route.

jeffkreeftmeijer:
In previous versions, we only used the full path name as in the sample
name if it matched "/". This patch removes that constraint, leaving two
types of routes:

1. Dyanmic routes, matched from the dynamicRoutes
2. Static routes, including everything else

This is based on the assumption that all non-dynamic routes are in fact
static. I haven't found any examples where this might fail, but I'm open
to suggestions on that.

Closes #337.

### DIFF
--- a/packages/nextjs/src/handler.ts
+++ b/packages/nextjs/src/handler.ts
@@ -61,11 +61,8 @@ export function getRequestHandler<T extends NextServer>(
       if (matched) {
         // matched to a dynamic route
         span.setName(`${req.method} ${matched.page}`)
-      } else if (!matched && pathname === "/") {
-        // the root
-        span.setName(`${req.method} ${pathname}`)
       } else {
-        span.setName(`${req.method} [unknown route]`)
+        span.setName(`${req.method} ${pathname}`)
       }
     }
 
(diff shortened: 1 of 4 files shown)


tombruijn:
GitHub doesn't show it, but the build on the last commit failed: https://appsignal.semaphoreci.com/jobs/8f696bab-2b5c-4841-ae63-f72ee9a98210

tombruijn (approved):
I've tested in the test-setups this works

jeffkreeftmeijer:
Tests fixed: https://appsignal.semaphoreci.com/workflows/2410cc36-c923-49e0-8acd-7d9296382e54?pipeline_id=6c854f06-dbcd-4fee-90f0-014f14b6c703

shairyar (approved):

thijsc:
If you rebase and force-push again it will trigger CI, that was off for a while before.

thijsc (approved):

Issue #375, opened 2021-03-03
Diagnose tool should not submit report by default


tombruijn:
The behavior of the other integrations is not to send the report by default, to give users the choice instead with a prompt, so they can decide if they want to send the data (which is shown above the prompt) to our servers.

- [x] Switch from `--no-report` to `--send-report` and `--no-send-report`, and don’t send the report by default. (https://github.com/appsignal/appsignal-nodejs/pull/392)
- [x] Update the documentation to mention the command line option (https://github.com/appsignal/appsignal-docs/pull/457)
- [x] Ask to send the report when running diagnose (https://github.com/appsignal/appsignal-nodejs/pull/392)

tombruijn:
Change has been published. Closing.

PR #392, opened 2021-05-12
Add --[no-]send-report options to diagnose

jeffkreeftmeijer:
Replace the `--no-report` option--which turned sending the diagnose report
to AppSignal's servers off--with `--send-report` and `--no-send-report`.

By default, the report is not sent if you don't pass `--send-report`.

Part of https://github.com/appsignal/appsignal-nodejs/issues/375.

### DIFF
--- a/packages/nodejs/bin/diagnose
+++ b/packages/nodejs/bin/diagnose
@@ -11,9 +11,9 @@ const { DiagnoseTool } = require("../dist/diagnose")
 // enable diagnose mode
 process.env["_APPSIGNAL_DIAGNOSE"] = "true"
 
-// providing --no-report at runtime stops the report from bieng sent
+// providing --send-report sends the report. --no-send-report stops the report from bieng sent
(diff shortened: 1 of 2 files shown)


packages/nodejs/bin/diagnose
```diff
@@ -11,9 +11,9 @@ const { DiagnoseTool } = require("../dist/diagnose")
 // enable diagnose mode
 process.env["_APPSIGNAL_DIAGNOSE"] = "true"
 
-// providing --no-report at runtime stops the report from bieng sent
+// providing --send-report sends the report. --no-send-report stops the report from bieng sent
```
tombruijn, line 14:
```suggestion
// providing --send-report sends the report. --no-send-report stops the report from being sent
```

tombruijn (approved):
I tested this locally by running:

```
cd packages/nodejs
bin/diagnose --send-report
bin/diagnose --no-send-report
```

PR #394, opened 2021-05-13
Add mono to project

tombruijn:
Replace lerna with mono, our own tool for managing mono repos.
https://github.com/appsignal/mono/

Follow the instructions on the mono README to install it locally.

Mono is also installed on the CI using the `scripts/setup_ci` script
which is run before every job. The mono download is cached between
jobs so we don't download it for every job fresh. Update the version
variable `MONO_VERSION` to update the mono install on the CI.

Mono depends on npm workspaces, which are available since npm 7. npm 7
also uses lockfileVersion "2" which is a slightly different format, but
causes a lot of unrelated changes in this diff.


tombruijn:
This replaces #367

tombruijn:
Waiting on anything that may come out of https://github.com/appsignal/test-setups/pull/24

jeffkreeftmeijer (approved):

thijsc (approved):
Nice!

PR #401, opened 2021-06-24
Print diagnose report from CLI

jeffkreeftmeijer:
In order to get the CLI diagnose command up to speed with the output from the Ruby and Elixir integrations, we created https://github.com/appsignal/diagnose_tests, which is an specification that describes the desired output to use as a guide and check for the diagnose output. It's modeled after the diagnose command in Ruby, and will eventually be used to test and develop diagnose commands for all integrations. 

This patch introduces diagnose_tests to the Node.js library, and implements a new diagnose command that matches the specification. Because of differences in the generated JSON reports, the diagnose output in the Node.js library is not an exact match with the Ruby integration. For those cases, the diagnose_tests specification has specific assertions for the Node.js. The goal is to eventually get rid of those, but we’re cutting this project here to be able to ship the new diagnose now, and have it checked on CI going forward.

*Note: this patch depends on #402*


thijsc (approved):
Nice!

packages/nodejs/src/config/configmap.ts
```diff
@@ -61,6 +61,7 @@ export const JS_TO_RUBY_MAPPING: { [key: string]: string } = {
   debug: "debug",
   log: "log",
   logPath: "log_path",
+  logFilePath: "log_file_path",
```
tombruijn, line 64: Is this line what adds it to the diagnose report as a config option? Because I don't think we should include it as a config option in the sent report. It's part of the paths section already and that contains more details.

What I see in a sent report:

image

tombruijn (approved):
Great on the human friendly output! Could you add a separate changeset for the updated output?

I notice some duplicate commits from #402 merged into main in this PR as well. Does it need a rebase?

Will the diagnose_tests repo also test the JSON payload in the future?

.semaphore/semaphore.yml
```diff
@@ -38,6 +38,22 @@ blocks:
...
+      env_vars:
+        - name: NODE_VERSION
+          value: "14"
+      jobs:
+        - name: Diagnose
+          commands:
+            - cache restore packages-$SEMAPHORE_GIT_SHA-$SEMAPHORE_GIT_BRANCH-v$NODE_VERSION
+            - git submodule init
+            - git submodule update
+            - LANGUAGE=nodejs test/integration/diagnose/bin/test
```
tombruijn, line 56: Can we disable the debug mode in the diagnose tests now? I see a lot of debug output there.

tombruijn:
@jeffkreeftmeijer we'll need to rebase this on the build matrix. Let me know if you need help with that.

jeffkreeftmeijer:
@tombruijn I’m working on some support issues today, so it would be nice if you could pick this up. If you don’t have time, I’ll do it early next week.

build_matrix.yml
```diff
@@ -55,6 +55,11 @@ matrix:
       path: "packages/nodejs"
       variations:
         - name: "nodejs"
+      extra_tests:
+        diagnose:
+          - git submodule init
+          - git submodule update
+          - LANGUAGE=nodejs test/integration/diagnose/bin/test
```
tombruijn, line 62: @jeffkreeftmeijer I've rebased this branch on the build matrix. I've added the diagnose tests as an "extra test" for the node.js package so it gets run for every node.js version as well.

packages/nodejs/src/config/configmap.ts
jeffkreeftmeijer: This line was added because the diagnose command uses the `JS_TO_RUBY_MAPPING` object to fetch the names for each field.
Because there was no name for the `logFilePath` field, it printed `undefined: '/tmp/appsignal.log'` before this change.
I'm pretty sure that'd mean this field was called `undefined` in the diagnose parser before, then.

In Ruby, the log_path and log_file_path aren't printed in the configuration at all.
I'll update the test in diagnose_tests to make sure this isn't the case in Node.js either, so we can remove this line again.

jeffkreeftmeijer:
I’ve updated this pull request to remove the `logFilePath` field and remove the debug output on CI.

packages/nodejs/src/cli/diagnose.ts
```diff
@@ -0,0 +1,249 @@
...
+
+    console.log(`Read more about how the diagnose config output is rendered`)
+    console.log(`https://docs.appsignal.com/nodejs/command-line/diagnose.html`)
+
+    this.print_newline()
+
+    console.log(`Validation`)
+    console.log(
+      `  Validating Push API key: ${this.colorize(
+        data["validation"]["push_api_key"]
```
tombruijn, line 120: Not sure if it's in scope but I noticed that this output isn't accurate.

Given this config output (with an undefined push_api_key, which I also don't know if we should print it as `undefined`):

```
Configuration
  ...
  push_api_key: undefined
```

I get this validation:

```
Validation
  Validating Push API key: valid
```

See also the admin view of this report: https://appsignal.com/admin/diagnose_reports/60e41eb35ac13f2d186f711f

packages/nodejs/src/cli/diagnose.ts
```diff
@@ -0,0 +1,249 @@
...
+        input: process.stdin,
+        output: process.stdout
+      })
+
+      let self = this
+      rl.question(
+        `  Send diagnostics report to AppSignal? (Y/n): `,
+        function (answer: String) {
+          switch (answer) {
+            case "":
```
tombruijn, line 177: Not sure I like a non breaking switch statement. We can also make the default value "y" in the switch?

```suggestion
          switch (answer || "y") {
```


```js
> "" || "y"
'y'
```

tombruijn (changes requested):
See previous comments, GitHub still shows me I have approved it 🤷‍♂️

packages/nodejs/src/cli/diagnose.ts
jeffkreeftmeijer: That should be caused by https://github.com/appsignal/appsignal-nodejs/issues/408, which causes the result to return “valid” for any input. I’ve created a separate issue to fix that to make sure this moves along.

tombruijn (approved):

tombruijn:
Can you add a changeset for this PR? Something about diagnose report improvements.

PR #402, opened 2021-06-24
Return the full log path in the diagnose report

jeffkreeftmeijer:
Currently, the diagnose command expects the logPath to include the
filename for the `appsignal.log` file. To get the `log_dir_path`, the
filename is stripped off. The `LogPath` is then used to return the full
path to the file.

In reality, it should be the other way around. The `LogPath` does not
include the filename, so it can be used as the log_dir_path in the
diagnose. Then, to get the full path to the `appsignal.log` file, this
patch appends the filename to the log path.

### DIFF
--- a/packages/nodejs/.changesets/fix-diagnose-log-paths.md
+++ b/packages/nodejs/.changesets/fix-diagnose-log-paths.md
@@ -1,0 +1,6 @@
+---
+bump: "patch"
+---
+
+Fix the log path reported when running the diagnose command to include the
+filename.
--- a/packages/nodejs/src/diagnose.ts
+++ b/packages/nodejs/src/diagnose.ts
@@ -93,7 +93,8 @@ export class DiagnoseTool {
 
     // we want to fall over if this value isn't present
     // (it should be)
-    const logPath = this.#config.data.logPath!
+    const logDirPath = this.#config.data.logPath!
+    const logPath = logDirPath + "/appsignal.log"
 
     // add any paths we want to check to this object!
     const files = {
@@ -101,7 +102,7 @@ export class DiagnoseTool {
         path: process.cwd()
       },
       log_dir_path: {
-        path: logPath.replace("/appsignal.log", "")
+        path: logDirPath
       },
       "appsignal.log": {
         path: logPath,
--- a/packages/nodejs/src/diagnose.ts
+++ b/packages/nodejs/src/diagnose.ts
@@ -93,8 +93,7 @@ export class DiagnoseTool {
 
     // we want to fall over if this value isn't present
     // (it should be)
-    const logDirPath = this.#config.data.logPath!
-    const logPath = this.#config.data.logFilePath!
+    const logFilePath = <string>this.#config.data.logFilePath!
 
     // add any paths we want to check to this object!
     const files = {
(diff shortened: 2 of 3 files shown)


thijsc (approved):

packages/nodejs/src/diagnose.ts
```diff
@@ -93,15 +93,16 @@ export class DiagnoseTool {
...
+    const logDirPath = this.#config.data.logPath!
+    const logPath = logDirPath + "/appsignal.log"
 
     // add any paths we want to check to this object!
     const files = {
       working_dir: {
         path: process.cwd()
       },
       log_dir_path: {
-        path: logPath.replace("/appsignal.log", "")
```
tombruijn, line 104: Removing this line could break with existing configurations. Our docs now say you need to specify the whole path: https://docs.appsignal.com/nodejs/configuration/options.html#option-logpath
Note to us: update that.

If someone specifies the "appsignal.log" file, the output would become `/tmp/appsignal.log/appsignal.log` (note the "appsignal.log" is mentioned twice).

Maybe we should keep this behavior (for a couple versions), but move it up to the other line changes in this PR where we fetch the `logPath` config value?

packages/nodejs/src/diagnose.ts
```diff
@@ -93,15 +93,16 @@ export class DiagnoseTool {
 
     // we want to fall over if this value isn't present
     // (it should be)
-    const logPath = this.#config.data.logPath!
+    const logDirPath = this.#config.data.logPath!
+    const logPath = logDirPath + "/appsignal.log"
```
tombruijn, line 97: We also have an (internal?) `logFilePath` config value: https://github.com/appsignal/appsignal-nodejs/blob/8a26df7974dbb2751acf561890118e4fb6d1812e/packages/nodejs/src/config.ts#L82-L86

Could we use that instead? I tried the code below but got a TypeScript error I don't quite understand yet.

```suggestion
    const logPath = this.#config.data.logFilePath!
```

```
src/diagnose.ts:135:9 - error TS2322: Type '{ exists: false; path: string; } | { exists: false; path: string; } | { exists: false; path: string | boolean | string[]; content: string[]; }' is not assignable to type 'FileMetadata'.
  Type '{ exists: false; path: string | boolean | string[]; content: string[]; }' is not assignable to type 'FileMetadata'.
    Types of property 'path' are incompatible.
      Type 'string | boolean | string[]' is not assignable to type 'string | undefined'.
        Type 'false' is not assignable to type 'string | undefined'.

135         paths[key] = {
```

tombruijn (changes requested):

packages/nodejs/.changesets/fix-diagnose-log-paths.md
```diff
@@ -0,0 +1,6 @@
+---
+bump: "patch"
+---
+
+Fix the log path reported when running the diagnose command to include the
+filename.
```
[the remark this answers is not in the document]
jeffkreeftmeijer: Noted, thanks. Let me know if you’d like me to update that here. 👍

packages/nodejs/src/diagnose.ts
jeffkreeftmeijer: I’ll add a test case for this. 👍

packages/nodejs/.changesets/fix-diagnose-log-paths.md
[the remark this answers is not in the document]
tombruijn: It's not needed now.

tombruijn (approved):

Issue #372, opened 2021-03-03
Diagnose install report location is easily overwritable


tombruijn:
The diagnose report location is stored in `/tmp/appsignal-install-report.json` which can be easily overwritable by other versions of the appsignal npm package. If there are other apps on the system also using the appsignal npm package they will overwrite this file during the installation of the packages. Only the last install's report will be readable and it might get reported for other apps that it does no longer apply to.

We should look if we can store the report file in the package directory (in `node_modules`) itself, or in some versioned and app path prefixed location.

tombruijn:
We can't store it in the `node_modules/@appsignal/nodejs-ext` dir, because when an optional package fails to install it doesn't persist in the node_modules dir, it gets removed on failure. Probably why it is stored in `/tmp` now.

We can't prefix it with a package version because `npx @appsignal/cli diagnose` doesn't know which package version it was trying to be install. (Also, where does the [cli package](https://www.npmjs.com/package/@appsignal/cli) come from in the repo?)

We could _not_ have the ext package fail with status code 1, and not have it be removed, store the report in there and be able to read it later? As long as the diagnose cli can find that report file later by searching for it in the `node_modules/@appsignal/nodejs-ext` dir?

jeffkreeftmeijer:
I’m not sure about breaking the status code, as that might have effects on the compilation of the app that we’re not foreseeing. Can we use a digest of the current project’s path to create a unique filename to use as the name for the tempfile?

tombruijn:
@jeffkreeftmeijer we could generate a digest based on the app path like so?

```js
#!/usr/bin/node

const { createHash } = require('crypto');
let dir = __dirname

// Whatever I found in the docs to generate a hash
const hash = createHash('sha256');
hash.update(dir);
const hashi = hash.digest('hex');
console.log({ hashi }); // cdb5d15546d89efe58d3772272d3e39bfcc8d1d2435bbaad91d8ccd470413361

// Or basic stripping of dangerous characters
dir = dir.replace(/[\/\\:]/g, "").toLowerCase()
console.log({ dir }) // userstombruijnappsignaltests
```

PR #409, opened 2021-07-08
Fix extension installation reporting

### LINKED ISSUE #371: Diagnose report successfull install on download, not on installation
I noticed in [this conversation](https://app.intercom.com/a/apps/yzor8gyw/inbox/inbox/mentions/conversations/16410700029272) that the musl build doesn't work, see #29.

Based on the diagnose report I had the idea it was successfully installed because it reported "success" for "status". Looks like it's set to "success" if it managed to download the agent and extension package, not when it actually finished installing. We need to change it to only report it as "installed" when it's done installing.

https://github.com/appsignal/appsignal-nodejs/blob/92c1b952c9acc8a053176ee0872ac7ed953b2517/packages/nodejs-ext/scripts/extension.js#L143

tombruijn:
When the extension installed we never stored if it installed correctly
or not. Originally it would always report it as a "success" status. In a
previous commit cb08b7a16c4f81239724a1f71ac99371015b568e, PR #373, I
changed the report status to "unknown", because after the `extension.js`
script was run the `node-gyp rebuild` command was run through the
`package.json` `install` script. After the node-gyp command was run it
would not return to the extension.js script and we couldn't check the
exit status of that command.

Instead of running the node-gyp command after the `extension.js` script,
run the node-gyp command from the `extension.js` script so we can listen
to the exit status and report the installation as a success or "error"
when it encounters an error.

This should fix #371, the issue describing the scenario above, the
install report not accurately describing the installation status.

I am not sure why the originally implementation was chosen. The original
implementation was present since the initial commit of the project.
There may be a problem with this new implementation that I am unaware of
now, because I'm no Node.js expert. But we call out to other commands
like `tar` from the `extension.js` script as well, so I figure it will
be fine.

## CI build update

I've added tests for the extension build failure scenario, and the
success scenario. For this I needed to have Jest run certain specs only
some of the time, based on the `_TEST_APPSIGNAL_EXTENSION_FAILURE`
environment variable (same as in the Ruby gem CI setup). This is done
with the `--filter` option (which links to a file with a filter
function), which is configured through the `package.json` test scripts.

Since these new tests are only for the `nodejs-ext` package I found it
unnecessary to spin up a new job for it, and instead add it as
`extra_commands` to the package's test commands.

Before the failure state is tested, the extension installation is
cleaned up, so it doesn't think it has installed correctly.

Because we install the extension in the "Build" task, and it's not
present in the project dir, but instead the `/tmp` dir on the system,
I've also copied the report location from the "Build" task to the job
that tests this report. Otherwise it would fail on a missing file, or
we'd need to install the extension again. Hopefully we can find a better
location for the install report location in issue #372.

### DIFF
--- a/packages/nodejs-ext/scripts/extension.js
+++ b/packages/nodejs-ext/scripts/extension.js
@@ -21,9 +21,15 @@ const {
 } = require("./report")
 
 const EXT_PATH = path.join(__dirname, "/../ext/")
+const testExtensionFailure =
+  process.env._TEST_APPSIGNAL_EXTENSION_FAILURE === "true"
 
 function download(url, outputPath) {
   return new Promise((resolve, reject) => {
+    if (testExtensionFailure) {
+      throw new Error("AppSignal internal test failure")
+    }
+
     const file = fs.createWriteStream(outputPath)
 
     https.get(url, response => {
@@ -104,9 +110,21 @@ function getMetadataForTarget({ architecture, target }) {
   return TRIPLES[triple.join("")]
 }
 
+function install() {
+  return new Promise((resolve, reject) => {
+    childProcess.exec("node-gyp rebuild", error => {
+      if (error) {
+        return reject(error)
+      } else {
+        return resolve()
+      }
+    })
+  })
+}
+
 // Script logic begins here
 ;(function () {
-  if (hasLocalBuild()) {
+  if (hasLocalBuild() && !testExtensionFailure) {
     // check for a local build (dev only)
     console.warn(`Using local build for agent. Skipping download.`)
     return process.exit(0)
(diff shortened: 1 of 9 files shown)


packages/nodejs-ext/scripts/extension.js
```diff
@@ -21,9 +21,15 @@ const {
 } = require("./report")
 
 const EXT_PATH = path.join(__dirname, "/../ext/")
+const testExtensionFailure =
+  process.env._TEST_APPSIGNAL_EXTENSION_FAILURE == "true"
```
tombruijn, line 25:
```suggestion
  process.env._TEST_APPSIGNAL_EXTENSION_FAILURE === "true"
```

packages/nodejs-ext/test/filter.js
```diff
@@ -0,0 +1,25 @@
+"use strict"
+
+const testExtensionFailure =
+  process.env._TEST_APPSIGNAL_EXTENSION_FAILURE == "true"
```
tombruijn, line 4:
```suggestion
  process.env._TEST_APPSIGNAL_EXTENSION_FAILURE === "true"
```

jeffkreeftmeijer (approved):

Issue #413, opened 2021-07-14
Diagnose report reads whole appsignal.log file


tombruijn:
It should only read the last 2MB like we do in the Ruby gem: https://github.com/appsignal/appsignal-ruby/blob/599c9f28086ea86c4f6d29993a90b86f5612bd99/lib/appsignal/cli/diagnose/paths.rb#L71-L74

PR #414, opened 2021-07-14
Fix diagnose report logging 10 appsignal.log lines

tombruijn:
When the appsignal.log file is read and split into an array of lines,
the last line of the log, which is always empty, is also considered a
line to be printed.

This is because when a String is split on a line break, it will return
that last empty line as a empty String.

```
$ node
> "foo\nbar\nbaz\n".split("\n")
[ 'foo', 'bar', 'baz', '' ]
```

Since it's empty it shows no useful information and
breaks the formatting, so trim the read data to not include that last
empty line.

Depends on https://github.com/appsignal/diagnose_tests/pull/3

### DIFF
--- a/packages/nodejs/src/cli/diagnose.ts
+++ b/packages/nodejs/src/cli/diagnose.ts
@@ -145,8 +145,10 @@ export class Diagnose {
     console.log(
       `    Path: ${this.format_value(data["paths"]["appsignal.log"]["path"])}`
     )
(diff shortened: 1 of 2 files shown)


tombruijn:
While reading the code I also found the issue described in https://github.com/appsignal/appsignal-nodejs/issues/413 where we read too much of the file.

jeffkreeftmeijer (approved):
Awesome!

Issue #389, opened 2021-05-04
Integration tests for integrations in a build matrix


jeffkreeftmeijer:
For each integration, there should be at least one integration test that spins up an example application and does a request. We can then test the results of that request by inspecting the JSON span representation from the extension. We’ll set up a build matrix to test the integrations across supported versions of our integrations.

tombruijn:
To fix the issue I was talking about in the call I'm thinking of:

The integration example test apps currently don't run `npm install` in their app directory. Somehow (without error or warning) the apps rely on the versions of the packages installed in the root of the repo. Which seems to work, but is weird, and we wouldn't need the example app's `package.json` for that. It will probably break if the example app's `package.json` lists a dependency that isn't part of the root `package.json`.

Moving all the development dependencies to the packages themselves, rather than have them in the root `package.json` will help organize things a bit. For my brain that would make more sense, but I don't know if it would for the entire project. We'd need to see what would happen with the different versions of packages we install in the CI build.

It would also help with different test setups where we may need to use a different combination of packages (or more packages) that we don't want to always have installed in the root of the project.

tombruijn:
Fixes for the problems I described in my previous comment are made in PR https://github.com/appsignal/appsignal-nodejs/pull/431

tombruijn:
I'm actually going to close this as the last PR for this is created: https://github.com/appsignal/appsignal-nodejs/pull/433

PR #421, opened 2021-08-02
Apollo integration test

jeffkreeftmeijer:
Add an integration test for Apollo that calls an endpoint in a simple example application. 

Part of https://github.com/appsignal/appsignal-nodejs/issues/389.

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -94,37 +94,21 @@ blocks:
       commands:
       - mono test --package=@appsignal/nodejs-ext
       - mono run --package @appsignal/nodejs-ext -- npm run test:failure
-    - name: "@appsignal/apollo-server - apollo-server@latest"
-      commands:
-      - npm install apollo-server-plugin-base@latest --save-dev
-      - mono test --package=@appsignal/apollo-server
     - name: "@appsignal/apollo-server - apollo-server@latest - integrations"
       commands:
       - npm install apollo-server-plugin-base@latest --save-dev
       - BUNDLE_GEMFILE=packages/apollo-server/test/Gemfile bundle install
       - BUNDLE_GEMFILE=packages/apollo-server/test/Gemfile bundle exec rspec packages/apollo-server/test
-    - name: "@appsignal/apollo-server - apollo-server@0.13.0"
-      commands:
-      - npm install apollo-server-plugin-base@0.13.0 --save-dev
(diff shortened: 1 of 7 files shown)


jeffkreeftmeijer:
This patch works locally, but runs into an issue checking the build matrix on CI. The build matrix is regenerated and produces a different result on my machine and on CI, even while using the exact same code with the exact same checksum.
I even added a Gemfile to make sure the Rake version matches, which didn't help, unsuprisingly.

[Removing the validation](https://appsignal.semaphoreci.com/workflows/39ea8a27-7a82-44b5-b01b-e0f5e0d9e2c3?pipeline_id=a9106a8b-3cf1-4df9-8cc8-bf501044a0d5) fixes the build. Or at least, gets past the first issue.

I need a second pair of eyes on this, so I’m leaving this pull request for now to focus on the other integration tests.

tombruijn:
@jeffkreeftmeijer The validation fails because generating the semaphore.yml file will remove the `mono test --package=@appsignal/apollo-server` jobs. It can't find any Node.js/Jest test files, so it doesn't include jobs for those specs in the semaphore.yml file. Without those jobs it still includes the "extra_tests", which are the RSpec tests you wrote. I don't know how the other jobs were included unless it finds some Node.js test files on your machine locally.

Unless I missed something, regenerate the semaphore.yml file, commit it, and push it.

jeffkreeftmeijer:
That makes sense, but I don’t think I have any extra tests lingering on my machine (my local checkout is in a clean state, as far as I can see). Could you regenerate the matrix and push it, if you have time? I’ll probably have time to take another look at this tomorrow.

tombruijn:
@jeffkreeftmeijer Yes.

It [checks these two things](https://github.com/appsignal/appsignal-nodejs/blob/0dae939e8bc369c8f2b2d8d6a44bc71a4dd82e5f/Rakefile#L142-L143) so maybe you created a `__tests__` directory initially? Git won't check in an empty directory, but locally it will exist.

tombruijn:
I improved the test files check in this PR: https://github.com/appsignal/appsignal-nodejs/pull/422

The build failure in this PR is caused by another issue.

tombruijn:
Blocked on https://github.com/appsignal/appsignal-nodejs/pull/431

tombruijn:
Closing in favor of #433

PR #423, opened 2021-08-06
Fix middleware path detection for Express

tombruijn:
Based on #419

## Fix middleware path detection for Express

The action name set for routes defined in a middleware did not account
for the path they were mounted on. Instead they would only report the
paths as defined in the routes on the middleware.

If mounted like this:

```
const adminRoutes = require("./admin")
app.use("/admin", adminRoutes)
```

The path `/admin` would not be included in the action name.

Use the `baseUrl` to prepend the mounted path on the action name.

Fixes appsignal/support#125

## Test non-root path in Express tests

Do not only test the root path, but test another path that can exist in
the app. This is how I found out that the AppSignal middleware that was
required was not actually used in the test app.

Without the middleware it would set the action name to:
`GET [unknown route]`

## Add query params to test URLS

This will help test what part of the request URL we fetch and set as the
action name. These query params should not be included in the action
name on the span.

### DIFF
--- a/packages/express/src/middleware/index.ts
+++ b/packages/express/src/middleware/index.ts
@@ -38,7 +38,9 @@ export function expressMiddleware(appsignal: NodeClient): RequestHandler {
         // if there is no error passed to `next()`, the span name will
         // be updated to match the current path
         if (req.route?.path) {
-          span.setName(`${method} ${req.route.path}`)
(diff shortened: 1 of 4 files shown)


tombruijn:
Tests are randomly failing for some reason 🤷‍♂️

jeffkreeftmeijer (approved):

PR #425, opened 2021-08-09
Refactor integration tests to reuse helpers

tombruijn:
Add an AppRunner class with all the logic to start and stop the example
app. Since it knows a bit about the app, we can also ask it about the
output and the log contents.

With this AppRunner all the behavior is put in one place and not all the
integration specs need to reimplement this.

Remove the usage of `Dir.chdir` which changes the entire Ruby app's
current working directory and pass along the `chdir` option to the
`spawn` command that starts the app so that only the example app is run
in that directory.

Based on #424 

---

Note: The build has failed on randomly failing diagnose integration tests. I did not want to retry it a lot to finally get a green build. It is something we need to take a better look at, they should be more stable.

### DIFF
--- a/packages/express/test/spec/integration_spec.rb
+++ b/packages/express/test/spec/integration_spec.rb
@@ -1,34 +1,15 @@
-require 'net/http'
-require 'tempfile'
-require 'timeout'
+require "net/http"
 
-RSpec.describe "Express.js" do
-  before(:all) do
-    tmpdir = Dir.mktmpdir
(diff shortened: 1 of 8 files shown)


jeffkreeftmeijer (approved):
Looks good!

I didn’t initially move the tests to use a shared helper because that crosses the package boundary, making each package in the monorepo dependent on said monorepo. I’m not sure this is a problem, but it will make it more difficult to split these projects up later. Just a note, no action required.

PR #435, opened 2021-08-19
Fix timestamp calculation

thijsc:
We had an incorrect calculation in place for custom timestamps. Add a test for that and fix it.


tombruijn (approved):

tombruijn:
CI build fails because of a random test failure. We can ignore that.

luismiramirez (approved):

PR #436, opened 2021-08-25
Add root span helper in tracer

luismiramirez:
## Add root span helper in tracer

There's new storage for root spans based on uid based on the implementation done for child spans. Now there will always be a root span available, and all the context and spans created under it will be considered children's.

It's accessible as the current span using `rootSpan()` through the tracer.

## Add error tracking to RootSpan

The error handling was done using any span instance as the caller allowing to attach errors to child spans. This will cause the agent to ignore those errors and, therefore, not getting into AppSignal's application.

Now errors can be added as in our other integrations directly from the tracer, attaching them to the root span.

### Steps

- [x] Add access to the root span from the tracer
- [x] Add new functions and properties including `addError()` to Tracer type interface (https://github.com/appsignal/appsignal-javascript/pull/495)
- [x] Implement `addError()` Tracer function.
- [x] Allow `addError()` to be called only from `RootSpan`s
- [x] Deprecate old `addError()` from `BaseSpan`


thijsc (commented):
Looks sensible so far!

packages/apollo-server/src/index.ts
```diff
@@ -1,7 +1,6 @@
-import { NodeClient } from "@appsignal/types"
```
tombruijn, line 1: I think we should keep the types. They're also useful for our users who use TypeScript and use types from our library. This way they don't need to specify `any` as argument types.

We can update the interfaces in types package in the javascript repo if needed.

packages/nodejs/.changesets/add-global-error-tracking-helper--improve-root-span-management-.md
```diff
@@ -0,0 +1,5 @@
+---
+bump: "minor"
+---
+
+Add global error tracking helper. Improve root span management.
```
tombruijn, line 5:
```suggestion
Add [`tracer.addError` helper function](https://docs.appsignal.com/nodejs/tracing/exception-handling.html) to set errors on the root span. If errors were set on child spans they were ignored.
```
It's a message for the user, so we should explain how the change impacts them. We should also update the linked page to explain how `tracer.addError` works.

tombruijn (changes requested):
I still need to test this PR in a test app, but left some feedback so far.

PR #440, opened 2021-09-08
Add mirror download support

luismiramirez:
The agent was always downloaded from a single URL. Adding mirror support
we ensure download availability as we do with Ruby library.

### DIFF
--- a/packages/nodejs-ext/.changesets/add-mirrors-to-download-the-agent.md
+++ b/packages/nodejs-ext/.changesets/add-mirrors-to-download-the-agent.md
@@ -1,0 +1,5 @@
+---
+bump: "patch"
+---
+
+Add mirrors to download the agent from different online sources in case some sources are not available to the app network.
--- a/packages/nodejs-ext/scripts/extension.js
+++ b/packages/nodejs-ext/scripts/extension.js
@@ -6,7 +6,7 @@ const fs = require("fs")
 const crypto = require("crypto")
 const childProcess = require("child_process")
 
-const { TRIPLES } = require("./extension/constants")
+const { AGENT_VERSION, MIRRORS, TRIPLES } = require("./extension/constants")
 
 const {
   hasLocalBuild,
@@ -25,12 +25,17 @@ const EXT_PATH = path.join(__dirname, "/../ext/")
 const testExtensionFailure =
   process.env._TEST_APPSIGNAL_EXTENSION_FAILURE === "true"
 
-function download(url, outputPath) {
-  return new Promise((resolve, reject) => {
-    if (testExtensionFailure) {
-      throw new Error("AppSignal internal test failure")
-    }
+class DownloadError extends Error {
+  constructor(message, downloadUrl) {
+    super(message)
+    this.name = this.constructor.name
+    this.downloadUrl = downloadUrl
+  }
+}
 
+function downloadFromMirror(mirror, filename, outputPath) {
+  return new Promise((resolve, reject) => {
+    const url = path.join(mirror, AGENT_VERSION, filename)
     const file = fs.createWriteStream(outputPath)
 
     https.get(url, response => {
@@ -38,12 +43,40 @@ function download(url, outputPath) {
 
       if (statusCode >= 400) {
         return reject(
-          new Error(`Request to CDN failed with code HTTP ${statusCode}`)
+          new DownloadError(`Request to CDN failed with code HTTP ${statusCode}`, url)
         )
+      } else {
+        response.pipe(file).on("finish", () => resolve(url))
       }
-
-      response.pipe(file).on("finish", () => resolve(outputPath))
     })
+  })
+}
+
+function download(mirrors, filename, outputPath) {
+  return new Promise((resolve, reject) => {
+    if (testExtensionFailure) {
+      throw new DownloadError("AppSignal internal test failure", undefined)
+    }
+
+    if (mirrors.length === 0) {
+      reject(new DownloadError("Could not download agent from any mirror", undefined))
+      return
+    }
+
+    downloadFromMirror(mirrors.shift(), filename, outputPath)
+      .then(url => {
+        const downloadData = {
+          downloadUrl: url,
+          outputPath: outputPath
+        }
+        resolve(downloadData)
+      })
+      .catch(error => {
+        console.error("Error downloading from mirror:", error)
+        download(mirrors, filename, outputPath)
+          .then(downloadData => resolve(downloadData))
+          .catch(error => reject(error))
+      })
   })
 }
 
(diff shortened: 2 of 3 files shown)


packages/nodejs-ext/scripts/extension.js
```diff
@@ -25,28 +25,57 @@ const EXT_PATH = path.join(__dirname, "/../ext/")
...
+    download_from_mirror(mirrors.shift(), filename, outputPath)
+      .then(url => {
+        return {
+          downloadUrl: url,
+          outputPath: outputPath
+        }
+      })
+      .catch(error => {
+        console.log(error)
+        download(mirrors, filename, outputPath)
```
tombruijn, line 74: something tells me this should have a `return` in front of it, otherwise it will return the promise of on line 56, rather than the final promise for the mirror that worked

packages/nodejs-ext/scripts/extension.js
```diff
@@ -25,28 +25,57 @@ const EXT_PATH = path.join(__dirname, "/../ext/")
...
+
+    download_from_mirror(mirrors.shift(), filename, outputPath)
+      .then(url => {
+        return {
+          downloadUrl: url,
+          outputPath: outputPath
+        }
+      })
+      .catch(error => {
+        console.log(error)
```
tombruijn, line 73:
```suggestion
        console.error(error)
```

packages/nodejs-ext/scripts/extension.js
```diff
@@ -25,28 +25,57 @@ const EXT_PATH = path.join(__dirname, "/../ext/")
...
+        response.pipe(file).on("finish",() => resolve(url))
       }
-
-      response.pipe(file).on("finish", () => resolve(outputPath))
     })
   })
 }
 
+function download(mirrors, filename, outputPath) {
+  return new Promise((resolve, reject) => {
```
tombruijn, line 56: Does this promise resolve properly? We don't use the `resolve`

packages/nodejs-ext/scripts/extension.js
```diff
@@ -25,28 +25,57 @@ const EXT_PATH = path.join(__dirname, "/../ext/")
...
+    if (testExtensionFailure) {
+      throw new DownloadError("AppSignal internal test failure", undefined)
+    }
+
+    if (mirrors.length === 0) {
+      return reject(new DownloadError("Could not download agent from any mirror", undefined))
+    }
+
+    download_from_mirror(mirrors.shift(), filename, outputPath)
```
tombruijn, line 65:
```suggestion
    return download_from_mirror(mirrors.shift(), filename, outputPath)
```

I think?

packages/nodejs-ext/scripts/extension.js
```diff
@@ -25,28 +25,57 @@ const EXT_PATH = path.join(__dirname, "/../ext/")
...
 }
 
+function download(mirrors, filename, outputPath) {
+  return new Promise((resolve, reject) => {
+    if (testExtensionFailure) {
+      throw new DownloadError("AppSignal internal test failure", undefined)
+    }
+
+    if (mirrors.length === 0) {
+      return reject(new DownloadError("Could not download agent from any mirror", undefined))
```
tombruijn, line 62: why use `reject` on one failure scenario and `throw` on the other on line 58?

packages/nodejs-ext/scripts/extension.js
```diff
@@ -25,28 +25,57 @@ const EXT_PATH = path.join(__dirname, "/../ext/")
...
+    }
+
+    download_from_mirror(mirrors.shift(), filename, outputPath)
+      .then(url => {
+        return {
+          downloadUrl: url,
+          outputPath: outputPath
+        }
```
tombruijn, line 70:
```suggestion
        resolve({
          downloadUrl: url,
          outputPath: outputPath
        })
```
I think?

tombruijn (commented):
Late afternoon review with some questions. I'll have a fresh look tomorrow

packages/nodejs-ext/scripts/extension.js
luismiramirez: Yes, apparently when you resolve from inside the result of a promise, you have to use `return` instead of `resolve`

packages/nodejs-ext/scripts/extension.js
luismiramirez: This one is used by the script and is treated on the `catch` on line 212. The `throw` was already there for test purposes.

packages/nodejs-ext/scripts/extension.js
luismiramirez: It doesn't work with resolve as said in https://github.com/appsignal/appsignal-nodejs/pull/440#discussion_r705027072, as it is inside a `then` block, to resolve the outer one you have to use `return`

packages/nodejs-ext/scripts/extension.js
luismiramirez: Updated and now using `return resolve`

packages/nodejs-ext/scripts/extension.js
```diff
@@ -25,28 +25,60 @@ const EXT_PATH = path.join(__dirname, "/../ext/")
...
+      return reject(new DownloadError("Could not download agent from any mirror", undefined))
+    }
+
+    download_from_mirror(mirrors.shift(), filename, outputPath)
+      .then(url => {
+        const downloadData = {
+          downloadUrl: url,
+          outputPath: outputPath
+        }
+        return resolve(downloadData)
```
tombruijn, line 71:
```suggestion
        resolve({
          downloadUrl: url,
          outputPath: outputPath
        })
```
This also works for me. No return needed and can immediately pass in the object to the `resolve`

packages/nodejs-ext/.changesets/add-mirrors-to-download-the-agent.md
```diff
@@ -0,0 +1,5 @@
+---
+bump: "patch"
+---
+
+Add mirrors to download the agent
```
tombruijn, line 5:
```suggestion
Add mirrors to download the agent from different online sources in case some sources are not available to the app network.
```

tombruijn (changes requested):

thijsc (approved):

packages/nodejs-ext/scripts/extension.js
```diff
@@ -25,28 +25,60 @@ const EXT_PATH = path.join(__dirname, "/../ext/")
...
     https.get(url, response => {
       const { statusCode } = response
 
       if (statusCode >= 400) {
         return reject(
-          new Error(`Request to CDN failed with code HTTP ${statusCode}`)
+          new DownloadError(`Request to CDN failed with code HTTP ${statusCode}`, url)
         )
+      } else {
+        response.pipe(file).on("finish",() => resolve(url))
```
tombruijn, line 49:
```suggestion
        response.pipe(file).on("finish", () => resolve(url))
```

packages/nodejs-ext/scripts/extension.js
```diff
@@ -25,28 +25,60 @@ const EXT_PATH = path.join(__dirname, "/../ext/")
...
 }
 
+function download(mirrors, filename, outputPath) {
+  return new Promise((resolve, reject) => {
+    if (testExtensionFailure) {
+      throw new DownloadError("AppSignal internal test failure", undefined)
+    }
+
+    if (mirrors.length === 0) {
+      reject(new DownloadError("Could not download agent from any mirror", undefined))
```
tombruijn, line 62:
```suggestion
      reject(new DownloadError("Could not download agent from any mirror", undefined))
      return
```

Turns out we do need a return here. Not to return the promise result, but to have the function exit early. (Calling `reject` doesn't stop the function.) Otherwise, if any error occurs for all mirrors in the `downloadFromMirror` function, it will fall into an infinite loop, retrying an empty mirrors Array over and over and over and over again.

We also could use `throw` here, like we do on line 58, but I probably prefer to use reject. Line 58 should also be updated (in another PR).

tombruijn (approved):
I think these are the last two points of feedback I have. I tried breaking it as much as I can. I'm spending extra attention on this, because if the script breaks we're going to have a hard time debugging it later.

@luismiramirez great work!

PR #442, opened 2021-09-10
Remove outputPath from downloadData in installer

tombruijn:
The outputPath variable doesn't change or is affected by the `download`
and `downloadFromMirror` functions.

We can remove it from the `downloadData` resolve value and use the
`outputPath` variable directly.

[skip changeset]

### DIFF
--- a/packages/nodejs-ext/scripts/extension.js
+++ b/packages/nodejs-ext/scripts/extension.js
@@ -69,17 +69,11 @@ function download(mirrors, filename, outputPath) {
     }
 
     downloadFromMirror(mirrors.shift(), filename, outputPath)
-      .then(url => {
-        const downloadData = {
-          downloadUrl: url,
-          outputPath: outputPath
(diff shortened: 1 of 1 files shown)


packages/nodejs-ext/scripts/extension.js
```diff
@@ -71,8 +71,7 @@ function download(mirrors, filename, outputPath) {
     downloadFromMirror(mirrors.shift(), filename, outputPath)
       .then(url => {
         const downloadData = {
```
luismiramirez, line 73: I think we don't need `downloadData` to be an object anymore. We can just return the URL

luismiramirez (changes requested):

packages/nodejs-ext/scripts/extension.js
tombruijn: Good point! I removed the object.

PR #446, opened 2021-09-16
Add rootSpan and addError helpers

luismiramirez:
## Add rootSpan and addError helpers

Root span wasn't accessible when working with contexts, only the current
child span was. Now there's a different storage for root spans and
they're always accessible from the tracer object.

A new `addError` function has been added to the tracer object and this
is the one that must be used when sending errors to AppSignal. It'll
look for the root span no matter the context or the current span, and
will attach the errors to it so they don't get ignored by the agent.

## Rename addError helpers to setError

In order to comply with AppSignal naming rules, we rename the addError
helper to setError

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "packages/*"
   ],
   "devDependencies": {
-    "@appsignal/types": "^2.1.1",
+    "@appsignal/types": "^2.1.2",
     "@types/jest": "^26.0.19",
     "husky": "^4.3.6",
     "jest": "^26.6.3",
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "packages/*"
   ],
   "devDependencies": {
-    "@appsignal/types": "^2.0.2",
+    "@appsignal/types": "^2.1.1",
     "@types/jest": "^26.0.19",
     "husky": "^4.3.6",
     "jest": "^26.6.3",
--- a/packages/nodejs/.changesets/add-root-span-and-add-error-helpers.md
+++ b/packages/nodejs/.changesets/add-root-span-and-add-error-helpers.md
@@ -1,0 +1,10 @@
+---
+bump: "minor"
+---
+
+Add rootSpan and addError helpers.
+
+Errors added to child spans are ignored by the agent. Now the rootSpan is
+always accessible from the tracer object as well as addError. The addError
+function allows to track errors on demand and they will be always attached
+to the main current span, so they don't get ignored by the agent.
--- a/packages/nodejs/.changesets/add-root-span-and-add-error-helpers.md
+++ b/packages/nodejs/.changesets/add-root-span-and-add-error-helpers.md
@@ -2,9 +2,9 @@
 bump: "minor"
 ---
 
-Add rootSpan and addError helpers.
+Add rootSpan and setError helpers.
 
 Errors added to child spans are ignored by the agent. Now the rootSpan is
-always accessible from the tracer object as well as addError. The addError
+always accessible from the tracer object as well as setError. The setError
 function allows to track errors on demand and they will be always attached
 to the main current span, so they don't get ignored by the agent.
--- a/packages/nodejs/src/noops/tracer.ts
+++ b/packages/nodejs/src/noops/tracer.ts
@@ -21,6 +21,14 @@ export class NoopTracer implements Tracer {
     return new NoopSpan()
   }
 
+  public rootSpan(): NodeSpan {
+    return new NoopSpan()
+  }
+
+  public addError(error: Error): NodeSpan {
+    return new NoopSpan()
+  }
+
   public withSpan<T>(span: NodeSpan, fn: (s: NodeSpan) => T): T {
     return fn(span)
   }
--- a/packages/nodejs/src/noops/tracer.ts
+++ b/packages/nodejs/src/noops/tracer.ts
@@ -25,7 +25,7 @@ export class NoopTracer implements Tracer {
     return new NoopSpan()
   }
 
-  public addError(error: Error): NodeSpan {
+  public setError(error: Error): NodeSpan {
     return new NoopSpan()
   }
 
--- a/packages/nodejs/src/span.ts
+++ b/packages/nodejs/src/span.ts
@@ -144,18 +144,11 @@ export class BaseSpan implements NodeSpan {
   }
 
   /**
-   * Adds a given `Error` object to the current `Span`.
-   * @deprecated Since version 2.1.0. Will be deleted in version 3.0.0
-   * use tracer.addError() instead.
+   * Set a given `Error` object to the current `Span`.
+   * Use tracer.setError() instead.
    */
-  public addError(error: Error): this {
-    if (!error) return this
-
-    const stackdata = Data.generate(
-      error.stack ? error.stack.split("\n") : ["No stacktrace available."]
-    )
-
-    span.addSpanError(this._ref, error.name, error.message, stackdata)
+  public setError(error: Error): this {
+    console.warn("setError() can only be called from a RootSpan object")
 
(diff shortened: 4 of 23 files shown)


packages/nodejs/src/tracer.ts
```diff
@@ -64,6 +70,30 @@ export class BaseTracer implements Tracer {
...
+
+  /*
+   * Adds the given error to the root Span.
+   *
+   * If there is no root Span available to add the error, `undefined` is returned.
+   */
+  public addError(error: Error): NodeSpan | undefined {
+    const activeRootSpan = this.rootSpan()
+
+    if (activeRootSpan instanceof NoopSpan) return
```
tombruijn, line 90: Would calling any function on the noopspan not also prevent any action from being taken? I think we can remove this.

tombruijn (changes requested):

tombruijn:
I tested this branch and compared it to the main branch and notice something strange with the root/current span. (See also debug notes at the end of this comment.)

To debug I add this line to an app running the main branch to track the currentSpan trace and span id:

```js
console.log("!!! currentSpan in request", tracer.currentSpan(), tracer.currentSpan().traceId, tracer.currentSpan().spanId)
```

The output shows a new span per request:

```
Request on /
!!! currentSpan in request RootSpan { _ref: [External: 5282f50] } VbYocIPzwoRt9oJ6 Ry4elKa3
Request on /
!!! currentSpan in request RootSpan { _ref: [External: 527a570] } TiqJJSwCHe3h2G8F gr7I1yvW
Request on /
!!! currentSpan in request RootSpan { _ref: [External: 5363cc0] } 6xEZNldc7cMz3ig4 iR9DB02u
Request on /
!!! currentSpan in request RootSpan { _ref: [External: 51d1ef0] } RYuROuyNFsnYIitZ SiNiae5C
```

All good.

When I add the same line to an app running this PR's branch, to track the current span, this is the output for multiple requests:

```
Request on /
!!! currentSpan in request RootSpan { _ref: [External: 5354130] } qyqhsgUMTzw4sXyD sMpSChGf
Request on /
!!! currentSpan in request ChildSpan { _ref: [External: 5542e20] }
Request on /
!!! currentSpan in request ChildSpan { _ref: [External: 5355ab0] }
Request on /
!!! currentSpan in request ChildSpan { _ref: [External: 56284e0] }
```

The current span is a childspan on subsequent requests, not a rootspan, and the traceId and spanId are not printed.

If I wait around 10 seconds between requests it does print a new rootspan and the trace and span ids are printed.

If I add the following line to the app to track the rootSpan, using this PR's helper functions:

```js
console.log("!!! rootSpan in request", tracer.rootSpan(), tracer.rootSpan().traceId, tracer.currentSpan().spanId)
```

I get the output below. Notice how the root span ref is the same for those requests. I think the trace and span id are empty because that rootspan is already closed after the first requests. But somehow the rootspan seems to persist across multiple requests, for a short time. I don't think we unset it when the request is completed.

```
Request on /
!!! rootSpan in request RootSpan { _ref: [External: 54ffc20] } 0gPqxeiDsKUkqghy TiuEW0Ni
Request on /
!!! rootSpan in request RootSpan { _ref: [External: 54ffc20] }
Request on /
!!! rootSpan in request RootSpan { _ref: [External: 54ffc20] }
Request on /
!!! rootSpan in request RootSpan { _ref: [External: 54ffc20] }
Request on /
!!! rootSpan in request RootSpan { _ref: [External: 54ffc20] }
```

---

Debug notes:

I'm using the `nodejs/express-postgres` example app, testing on the [root route `/`](https://github.com/appsignal/test-setups/blob/main/nodejs/express-postgres/app/app.js#L27). I add the lines I showed above under the [`console.log("Request on /")` line on line 28.](https://github.com/appsignal/test-setups/blob/392da188db4851e63e25c500484282d8f7580a53/nodejs/express-postgres/app/app.js#L28)

luismiramirez:
Wow, that's a nice catch!

Thanks for all the debugging information, right now, I'm not able to reproduce it, but indeed we might be unsetting it unproperly. I will come back to you with updates when I get to reproduce it and then fix it.

packages/nodejs/src/span.ts
```diff
@@ -241,4 +243,19 @@ export class RootSpan extends BaseSpan {
       this._ref = span.createRootSpan(namespace)
     }
   }
+
+  /*
+   * Adds a given `Error` object to the current `RootSpan`.
+   */
+  public addError(error: Error): this {
```
tombruijn, line 250:
```suggestion
  public setError(error: Error): this {
```

We don't want to make `addError` public.

packages/nodejs/src/noops/tracer.ts
```diff
@@ -21,6 +21,14 @@ export class NoopTracer implements Tracer {
     return new NoopSpan()
   }
 
+  public rootSpan(): NodeSpan {
+    return new NoopSpan()
+  }
+
+  public addError(error: Error): NodeSpan {
```
tombruijn, line 28:
```suggestion
  public setError(error: Error): NodeSpan {
```

luismiramirez:
@tombruijn PR Updated!

I got to reproduce your scenario and managed a way to fix it. `executionAsyncId` might point to a root or a child span, so we have to clean up the storage for both types of spans when finishing a transaction. Also added the `package.json` as requested.

I am currently working on renaming `addError`.

tombruijn:
@luismiramirez looks good! I tested the new revision and I don't see the issue anymore. I added some code to see what the scopemanager's internals tracked and I notice that it seems to reuse the `executionAsyncId` for many of the same requests when I quickly refresh, but if I wait 10 seconds or so it uses another `executionAsyncId`. That explains why the 10 second delay resulted in different behavior. I also see the scopes root map doesn't grow anymore.

packages/nodejs/src/span.ts
luismiramirez: This function must remain public, it's used from the Tracer and from the ScopeManager. Sometimes we want to call `RootSpan.setError()` from the outside when the Tracer is not available or needed.

tombruijn: My comment was unclear. We don't want to make the logic of an `addError` public yet, adding multiple errors, because we don't support that yet. So we stay with `setError` for now.

jeffkreeftmeijer (approved):

packages/nodejs/.changesets/add-root-span-and-add-error-helpers.md
```diff
@@ -0,0 +1,10 @@
+---
+bump: "minor"
+---
+
+Add rootSpan and addError helpers.
+
+Errors added to child spans are ignored by the agent. Now the rootSpan is
+always accessible from the tracer object as well as addError. The addError
+function allows to track errors on demand and they will be always attached
+to the main current span, so they don't get ignored by the agent.
```
tombruijn, line 10:
```suggestion
Add rootSpan and setError helpers.

Errors added to child spans are ignored by the agent. Now the rootSpan is
always accessible from the tracer object as well as setError. The setError
function allows to track errors on demand and they will be always attached
to the main current span, so they don't get ignored by the agent.
```

tombruijn (approved):

thijsc (approved):

package.json
```diff
@@ -12,7 +12,7 @@
     "packages/*"
   ],
   "devDependencies": {
-    "@appsignal/types": "^2.0.2",
+    "@appsignal/types": "^2.1.1",
```
tombruijn, line 15:
```suggestion
    "@appsignal/types": "^2.1.2",
```

2.1.2 is the last version right? Can you update and also commit the updated lock file?

tombruijn (approved):
One last thing, but other than that is looks good! 🎉

PR #451, opened 2021-10-01
Remove interface usage from types package for Node

luismiramirez:
## Remove interface usage from types package for Node

All Node.js-specific interfaces from the types package are now defined
inside the nodejs core package. There's still a dependency from types
package for common types as `Func`, `HashMap`, and `HasMapValue`.

With this change, we keep taking advantage of interfaces, but now they're
defined in a place where they're used.

## Next steps

Switch `@appsignal/types` dependency on instrumentation packages for `@appsignal/nodejs`. Still finding a way to make them work locally while in dev and from npm when installing.

[skip changeset]


thijsc (approved):

package-lock.json
```diff
@@ -10899,7 +10899,10 @@
       "requires": {
         "@appsignal/core": "^1.1.4",
         "@appsignal/nodejs-ext": "=2.0.0",
+<<<<<<< HEAD
         "@appsignal/types": "^2.1.5",
+=======
+>>>>>>> ba05474 (amend me)
```
tombruijn, line 10905: Can you fix the merge conflict?

packages/koa/src/patches/index.ts
```diff
@@ -1,10 +1,11 @@
 /**
+ *
```
tombruijn, line 2:
```suggestion
```

Unnecessary change?

packages/nodejs/package-lock.json
```diff
@@ -0,0 +1,2557 @@
+{
```
tombruijn, line 1: This file shouldn't be committed I think, or created event. The workspace setup merges all lock files into the root lock file.

tombruijn (changes requested):

packages/nextjs/package.json
```diff
@@ -5,7 +5,7 @@
   "types": "dist/index",
   "license": "MIT",
   "dependencies": {
-    "@appsignal/types": "^2.1.5",
+    "@appsignal/nodejs": "^2.0.0",
```
tombruijn, line 8: This will need to be updated now that 2.1.0 has been released. In **all** `package.json` files.

package-lock.json
luismiramirez: 🤦🏼 Missed this one

Issue #448, opened 2021-09-27
Diagnose that can't read the install report will fail


tombruijn:
We've seen a user be unable to load the installation and submit the diagnose report: https://app.intercom.com/a/apps/yzor8gyw/inbox/inbox/conversation/16410700063450

The Node.js integration doesn't handle not being able to load the install report. It should report an error state like in the Ruby gem: https://github.com/appsignal/appsignal-ruby/blob/0f984d851debffc2b186dc47ceeca1662b419352/lib/appsignal/cli/diagnose.rb#L336-L348

but instead it replaces it with an empty object: https://github.com/appsignal/appsignal-nodejs/blob/ad40d9d8ec0dcedb38973339d75a18373eb7d3ef/packages/nodejs/src/diagnose.ts#L91

Which will cause an error when nested keys are fetched: https://github.com/appsignal/appsignal-nodejs/blob/ad40d9d8ec0dcedb38973339d75a18373eb7d3ef/packages/nodejs/src/cli/diagnose.ts#L41

shairyar:
https://app.intercom.com/a/apps/yzor8gyw/inbox/inbox/4356044/conversations/16410700064306

shairyar:
Adding it here in case it is helpful while working on the diagnose report

```
During the node modules installation following error occured while building @appsignal/nodejs-ext extension:

[4/4] Building fresh packages...
warning Error running install script for optional dependency: "/tmp/build_f6b6088e/node_modules/@appsignal/nodejs-ext: Command failed.
Exit code: 1
Command: node scripts/extension.js
Arguments:
Directory: /tmp/build_f6b6088e/node_modules/@appsignal/nodejs-ext
Output:
Error: Checksum verification failed
at Hash.<anonymous> (/tmp/build_f6b6088e/node_modules/@appsignal/nodejs-ext/scripts/extension.js:68:25)
at Hash.emit (events.js:387:35)
at finish (internal/streams/writable.js:657:10)
at processTicksAndRejections (internal/process/task_queues.js:82:21)"
info This module is OPTIONAL, you can safely ignore this error
Done in 36.32s.
```

PR #454, opened 2021-10-07
Diagnose print parsing error on missing report

tombruijn:
When the diagnose command is run and the installation report is missing
the diagnose command crashes with an error.

```
UnhandledPromiseRejectionWarning: TypeError: Cannot read property
'download_url' of undefined
```

This is because when the report can't be read the
`getInstallationReport` function would only return an empty object,
rather than the expected object with a parsing error. (Expected in the
sense that's how our other integrations handle this type of error and
that's what the diagnose tool on the server side expects.)

With this change the report should no longer crash and the sent report
includes the error about the missing installation report.

I didn't know how to resolve a TypeScript issue with the optional `raw`
key on the parsing error object, so I created a ParsingError type to
indicate that it's a optional key. I couldn't get it to work otherwise.

Part of #448 
Based on #453

### DIFF
--- a/packages/nodejs/.changesets/report-parse-errors-in-diagnose-report.md
+++ b/packages/nodejs/.changesets/report-parse-errors-in-diagnose-report.md
@@ -1,0 +1,5 @@
+---
+bump: "patch"
+---
+
+The diagnose report will report parsing errors on reading or parsing the installation report. Previously, a missing installation report file would crash the diagnose tool.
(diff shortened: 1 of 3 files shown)


tombruijn:
Blocked on https://github.com/appsignal/appsignal-nodejs/pull/453#issuecomment-937682734 for the situation about the tests for this.

PR #474, opened 2021-10-29
Fix diagnose report's Push API key transmission

tombruijn:
## The problem

The diagnose report was transmitted without the Push API key as the
`api_key` query parameter. This caused the reports to not be properly
linked to their parent organizations on AppSignal.com, making it more
difficult to find them back for users.

It also caused the issues described in this server side issue:
https://github.com/appsignal/appsignal-server/issues/7710

## The solution

The request to submit the report now includes the `api_key` query
parameter.

I've moved the report transmission to the DiagnoseTool, because the
AppSignal config is available there, and I didn't want to initialize it
again in the Diagnose CLI class or move it to the Diagnose CLI class, or
fetch it from the Diagnose Tool's generated report. There's no big
design idea behind it, this was just easiest.

I've implemented error handling when the server returned an error code,
because otherwise it would print an ugly backtrace with where the JSON
parsing failed.

This behavior's tests should be implemented in the diagnose tests
submodule when we test the transmitted report as well.

### DIFF
--- a/packages/nodejs/src/diagnose.ts
+++ b/packages/nodejs/src/diagnose.ts
@@ -194,6 +194,52 @@ export class DiagnoseTool {
 
     return config
   }
+
+  public sendReport(data: object) {
+    const json = JSON.stringify(data)
+
+    const config = this.#config.data
+    const params = new URLSearchParams({ api_key: config["apiKey"] || "" })
+
+    const opts = {
+      port: 443,
+      method: "POST",
+      host: "appsignal.com",
+      path: `/diag?${params.toString()}`,
+      headers: {
+        "Content-Type": "application/json",
+        "Content-Length": json.length
+      },
+      cert: fs.readFileSync(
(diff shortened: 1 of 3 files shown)


packages/nodejs/src/diagnose.ts
```diff
@@ -194,6 +194,52 @@ export class DiagnoseTool {
...
+  public sendReport(data: object) {
+    const json = JSON.stringify(data)
+
+    const config = this.#config.data
+    const params = new URLSearchParams({ api_key: config["apiKey"] || "" })
+
+    const opts = {
+      port: 443,
+      method: "POST",
+      host: "staging.lol",
```
unflxw, line 207: Shouldn't this be `appsignal.com`, like it was on the previous implementation of this method?

tombruijn: Yep!

luismiramirez (approved):

unflxw (approved):

PR #475, opened 2021-11-01
Allow configuration of diagnose endpoint

tombruijn:
## Allow configuration of diagnose endpoint

For the diagnose tests we need to send the diagnose JSON report to a
custom endpoint. This change introduces the APPSIGNAL_DIAGNOSE_ENDPOINT
environment variable for this purpose.

The APPSIGNAL_DIAGNOSE_ENDPOINT environment variable is not a variable
we'll document publicly, because it's only meant for testing purposes.

The Node.js app needs to either use the https or http module, for the
production or local server. I didn't find a way to call `http:` protocol
endpoints with the https module.

## Wrap diagnose report in diagnose key

The diagnose CLIs in the Ruby and Elixir integrations wrap the diagnose
report in a "diagnose" key. This is because of how the Rails endpoint
was originally written.

Update the Node.js integration to match, because that's the least amount
of work.

## Remove unused app key from diagnose report

The app key is non-standard, no other integrations transmit this key as
part of the diagnose report. It's also empty, so we do not do anything
with it. Remove it.

## Update diagnose tests submodule

Includes tests for the transmitted diagnose JSON report.

Update the diagnose tests to match the new diagnose format without an
`app` key in the JSON report and transmit it wrapped in a `diagnose`
key.

Relies on https://github.com/appsignal/diagnose_tests/pull/23

[skip changeset]

### DIFF
--- a/packages/nodejs/src/diagnose.ts
+++ b/packages/nodejs/src/diagnose.ts
@@ -51,7 +51,6 @@ export class DiagnoseTool {
       library: this.getLibraryData(),
       installation: this.getInstallationReport(),
       host: this.getHostData(),
-      app: {},
       agent: this.#extension.diagnose(),
       config: {
         options: this.getConfigData(),
--- a/packages/nodejs/src/diagnose.ts
+++ b/packages/nodejs/src/diagnose.ts
@@ -197,7 +197,7 @@ export class DiagnoseTool {
   }
 
   public sendReport(data: object) {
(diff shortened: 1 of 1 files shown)


unflxw (approved):

tombruijn:
@unflxw thanks for the review! FYI: I added two small commits to iron out some differences between integrations. Probably nothing that would change your review.

luismiramirez (approved):

PR #478, opened 2021-11-03
Add all config opts to diagnose

luismiramirez:
## Add explicit default values for config options

Only a few config options were set up when initializing AppSignal. With
this change, all config user-overwritable options have now an explicit
default value.

## Add all user-configurable config opts to diagnose

With the default options in place, now we can iterate through the
configuration object data to print a more complete diagnose report.

The first option was to iterate through the constants defined in the config
map, but that would give us empty values in the diagnose report.

I decided to iterate through the default values and ignore those that
the user cannot set, as that's what we do in the other integrations.

### Caveats

I'll update the docs to add these two undocumented options:

- `files_world_accessible`
- `transaction_debug_mode`

https://github.com/appsignal/appsignal-docs/pull/492/


unflxw (approved):
Make sure the linter's happy before merging! 🎉

tombruijn:
> [skip changeset]

This change will print all the config options to the diagnose output and include them in the report right? I think that warrants a changeset.

luismiramirez:
> This change will print all the config options to the diagnose output and include them in the report right? I think that warrants a changeset. A change the users, and us, benefit from to ship

Agree, I'll add a patch changeset 👌🏼

packages/nodejs/src/config.ts
```diff
@@ -48,6 +43,34 @@ export class Configuration {
...
+      filesWorldAccessible: true,
+      filterDataKeys: [],
+      filterParameters: [],
+      filterSessionData: [],
+      ignoreActions: [],
+      ignoreErrors: [],
+      ignoreNamespaces: [],
+      log: "file",
+      logPath: "/tmp",
+      transactionDebugMode: false
```
tombruijn, line 70: I don't think this is everything. Some options I see missing (but I may be missing some):

- send_params: true
- skip_session_data: false
- request_headers: `[]` (we supply a list of headers in the other integrations.

It could also be there are not completely implemented in this integration? I can find some references to them in the `configmap.ts` file, but I can't find them being used anywhere. If they're not properly implemented, let's create issues for them to fix that and we can merge this PR as is.

tombruijn (changes requested):

packages/nodejs/src/config.ts
luismiramirez: None of them are present in the docs. I'll open issues for them cause as you say, they're probably not properly implemented.

tombruijn (approved):

PR #482, opened 2021-11-04
Implement config validation

### LINKED ISSUE #479: Config validator always returns true
## The problem

Due to this code:
https://github.com/appsignal/appsignal-nodejs/blob/243951a47631ee8b625c9155e0d7d0887bea94b0/packages/nodejs/src/config.ts#L47-L49

the configuration is always considered valid. This can give false positives on diagnose and make it harder to track possible bugs.

## The proposal

Consider the configuration valid only if the push API key is set up as we do in other integrations.

luismiramirez:
Config was always considered valid as `Config.isValid()` just had
a `return true` inside. Now we check the presence of an apiKey in the
configuration data to validate the config.

Closes #479

### DIFF
--- a/packages/nodejs/src/config.ts
+++ b/packages/nodejs/src/config.ts
@@ -36,11 +36,9 @@ export class Configuration {
 
   /**
    * Returns `true` if the current configuration is valid.
-   *
-   * @todo
    */
   public get isValid(): boolean {
-    return true
+    return (this.data.apiKey || "").trim() !== ""
   }
 
   /**
--- a/package-lock.json
+++ b/package-lock.json
@@ -10630,10 +10630,10 @@
     },
     "packages/apollo-server": {
       "name": "@appsignal/apollo-server",
-      "version": "1.0.13",
+      "version": "1.0.16",
       "license": "MIT",
       "dependencies": {
-        "@appsignal/nodejs": "=2.2.1",
+        "@appsignal/nodejs": "=2.2.4",
         "apollo-server-plugin-base": "^0.10.3",
         "tslib": "^2.0.3"
       },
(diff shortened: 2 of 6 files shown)


tombruijn (changes requested):

unflxw (approved):
Note that there is a different concept of validation for the push API key already in the codebase: the diagnose tool [does a request against the Appsignal servers to verify the API key](https://github.com/appsignal/appsignal-nodejs/blob/c750216b4c0ce4e40f97513787516532a35b35de/packages/nodejs/src/diagnose.ts#L111-L131) and adds that as `.validation.push_api_key` in the diagnose report.

Just mentioning it in case it's _that_ sort of validation we want here.

tombruijn:
> Just mentioning it in case it's that sort of validation we want here.

No, this is just checking if it's set at all or not. The validation check in the diagnose will help if users see no data being reported.

luismiramirez:
> Note that there is a different concept of validation for the push API key already in the codebase: the diagnose tool [does a request against the Appsignal servers to verify the API key](https://github.com/appsignal/appsignal-nodejs/blob/c750216b4c0ce4e40f97513787516532a35b35de/packages/nodejs/src/diagnose.ts#L111-L131) and adds that as `.validation.push_api_key` in the diagnose report.
> 
> Just mentioning it in case it's _that_ sort of validation we want here.

Yes, that endpoint validates that the provided API key authenticates properly. This change validates the provided config is in shape to start the integration before doing other checks.

packages/nodejs/src/config.ts
```diff
@@ -36,11 +36,9 @@ export class Configuration {
 
   /**
    * Returns `true` if the current configuration is valid.
-   *
-   * @todo
    */
   public get isValid(): boolean {
-    return true
+    return this.data.apiKey ? true : false
```
tombruijn, line 41:
```suggestion
    return (this.data.apiKey || "").trim() !== ""
```

This was more the implementation I'm looking for. Checking if the Push API key is not an empty string, or string with only spaces.

tombruijn (changes requested):

tombruijn (approved):

PR #484, opened 2021-11-04
Add sources to the diagnose report

### LINKED ISSUE #473: Diagnose report doesn't include sources section
The diagnose report doesn't include the sources section like we do in Elixir and Ruby. See the empty `sources` object in the report below.

It's hardcoded here to an empty object: https://github.com/appsignal/appsignal-nodejs/blob/ea685296957fdbd61dd13e168a5adbb64797d243/packages/nodejs/src/diagnose.ts#L57

Example diagnose report (on staging): https://staging.lol/admin/diagnose_reports/diag48db106d49eb0857a8a1

```json
{
  "config": {
    "options": {
      "debug": false,
      "log": "file",
      "log_path": "/tmp",
      "ca_file_path": "/Users/tombruijn/appsignal/nodejs/packages/nodejs/cert/cacert.pem",
      "endpoint": "https://push.appsignal.com",
      "env": "development",
      "push_api_key": "[FILTERED]",
      "active": true,
      "undefined": "/tmp/appsignal.log"
    },
    "sources": {
    }
  },
}
```

unflxw:
Closes #473.

### Implement sources in diagnose report

This commit implements the `.config.sources` attribute of the diagnose report.

### Add sources to the report output

The diagnose report output now contains information about which source provides a given configuration value (when that source is not `default`), and which other sources, if any, also provide the given configuration value.

This brings the diagnose report output's behaviour regarding sources in line with that of the Ruby and Elixir integration.

### Remove log_file_path from diagnose config

This commit removes the `log_file_path` entry from the diagnose config report and the corresponding output. The corresponding private environment variable `_APPSIGNAL_LOG_FILE_PATH` is still created, and the config has a `logFilePath` getter that can be accessed by other parts of the code, but it is no longer part of the "actual config", the one that is sent on the diagnose report.


packages/nodejs/src/cli/diagnose.ts
```diff
@@ -341,6 +337,80 @@ export class Diagnose {
...
+      .forEach(key => {
+        let keySources = this.configurationKeySources(key, sources)
+
+        switch (Object.keys(keySources).length) {
+          case 0:
+            if (key == 'log_file_path') {
+              // The `log_file_path` configuration key is generated after
+              // the config is read and does not belong to any sources.
+              // Since it is derived from `log_path`, let's give it the
+              // same source as `log_path`'s final value.
```
tombruijn, line 358: We should not set this key on the config object. See also my comment here: https://github.com/appsignal/integration-guide/issues/78#issuecomment-960716072

That way we don't have to filter it out again here. I don't want to print a different value here, that sounds confusing.

tombruijn (changes requested):

packages/nodejs/src/cli/diagnose.ts
unflxw: > We should not set this key on the config object.

Agreed 👍 I'll look into making this change. It's the only value that works in this way, so by removing `log_file_path` from the config, we could remove `case 0:` here entirely, or throw an error if it were to happen.

> I don't want to print a different value here, that sounds confusing.

It's not printing a different value, just pretending the value comes from the same source as wherever `log_path` comes from. So, if `log_path` was overriden on the environment variables, the output would be:

```
  log_path: /custom/location
    Sources:
      default: /tmp
      env: /custom/location
  log_file_path: /custom/location/appsignal.log (Loaded from: env)
```

Without this, it would not say `Loaded from: env`, implying it was set from `default`, which I thought would be more confusing.

luismiramirez (approved):

packages/nodejs/src/config.ts
```diff
@@ -14,17 +15,20 @@ import { ENV_TO_KEY_MAPPING, PRIVATE_ENV_MAPPING } from "./config/configmap"
...
+  sources: HashMap<Partial<AppsignalOptions>>
 
   constructor(options: Partial<AppsignalOptions>) {
-    writePrivateConstants()
-
-    this.data = {
-      ...this._defaultValues(),
-      ...this._loadFromEnvironment(),
-      ...options
+    this.sources = {
```
tombruijn, line 21: Can you add some tests to test if these values are properly set? In the diagnose tests we will test if the values are set at all, with some variations, but we need to unit test in the integration itself if the sources are set as expected when they are loaded from different sources.

tombruijn (changes requested):

packages/nodejs/src/config.ts
unflxw: Done! Ended up reorganising the config tests quite a bit, removing some redundant tests and adding checks about the sources to existing tests. There's no tests specifically about the sources, but there's tests about how values are overriden, and the resulting state of the sources is checked in those tests.

tombruijn: I think we can maybe test the rest as part of https://github.com/appsignal/appsignal-nodejs/issues/489

tombruijn (approved):

PR #485, opened 2021-11-04
Standardize key validation request

tombruijn:
Add the other query params to the auth request.

Update the error message to be standard for all integrations. Not all
integrations necessarily have the response status code on error, so
make sure all integrations start with "Failed to validate".

I removed the empty JSON body from the request, because we don't parse
the body in this request and our test server interpreted it as the
parameter `{}` with the value of an empty String.

Add test for diagnose Push API key output and report.
Update the diagnose tests with tests for the Push API key validation in
the diagnose CLI.

Depends on https://github.com/appsignal/diagnose_tests/pull/37
Blocked on https://github.com/appsignal/appsignal-nodejs/pull/484. Should be rebased after.

### DIFF
--- a/packages/nodejs/.changesets/standardize-diagnose-validation-error-message.md
+++ b/packages/nodejs/.changesets/standardize-diagnose-validation-error-message.md
@@ -1,0 +1,5 @@
+---
+bump: "patch"
+---
+
+Standardize diagnose validation failure message. Explain the diagnose request failed and why.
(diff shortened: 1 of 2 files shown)


tombruijn:
Build failures are because of changes only added in PR #484.

luismiramirez (approved):

unflxw (approved):

PR #496, opened 2021-11-10
Add tests for config module

### LINKED ISSUE #489: Tests needed for config options
We only test a selection of config options are set to the private environment variables `_APPSIGNAL_*`. We should test if all these values are at least set properly once.
All these keys need to be tested: https://github.com/appsignal/appsignal-nodejs/blob/3e78323af95527b34ae594ea5ca82f8e12aec295/packages/nodejs/src/config/configmap.ts#L34-L57

tombruijn:
## Add tests for config module

Reorganize the tests to test single behavior. The "Private environment
variables" describe block also tested things for the `logFilePath`
option. Now it only tests if the private env vars are written if we
specify all the options.

I removed the usage of `process.env` to configure AppSignal, and rely on
the config options given to the Configuration module. We will test the
if the config is read from the environment variables in other tests.

I removed the default config initialization at the top-level
`beforeEach`, because it sets env vars that aren't overwritten if unset.
I could also call `resetEnv`, but I think in this case it's better to
only initialize it when needed in the configuration that is needed.
Otherwise we end up three levels deep calling `new Configuration(...)`
and `resetEnv()`.

## Add tests for private env vars

Test if the values are set to anything that resembles what they should
be, rather than if they have any value at all.

## Split the different config source tests up

The different config sources (default, initial and env) were all grouped
in the "Default options" context initially. I split them out and made
them less dependent on shared context. It's bit more lines, but it
should be easier to read and less depending on other contexts.

---

[skip changeset] because it's only tests.
Closes #489

### DIFF
--- a/packages/nodejs/src/__tests__/config.test.ts
+++ b/packages/nodejs/src/__tests__/config.test.ts
@@ -16,6 +16,10 @@ describe("Configuration", () => {
     })
   }
 
+  function env(key: string): string | undefined {
+    return process.env[key]
+  }
+
   beforeAll(() => {
     initialEnv = Object.assign({}, process.env)
   })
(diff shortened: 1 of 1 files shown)


packages/nodejs/src/__tests__/config.test.ts
```diff
@@ -27,124 +54,211 @@ describe("Configuration", () => {
...
+          enableMinutelyProbes: false,
+          enableStatsd: true,
+          filesWorldAccessible: true,
+          filterDataKeys: ["password", "confirm_password"],
+          hostname: "MyHostName",
+          httpProxy: "http://localhost",
+          ignoreActions: ["MyAction", "MyOtherAction"],
+          ignoreErrors: ["MyError", "MyOtherError"],
+          ignoreNamespaces: ["MyNamespace", "MyOtherNamespace"],
+          logPath: os.tmpdir(),
```
unflxw, line 209: Nitpick: in Linux (that is, in our CI), this is `/tmp`, which _is_ the default value, so it's not testing non-default values.

packages/nodejs/src/__tests__/config.test.ts
```diff
@@ -27,124 +54,211 @@ describe("Configuration", () => {
...
+    describe("with logPath option", () => {
+      beforeEach(() => {
+        new Configuration({ logPath: "/other_path" })
       })
 
-      it("with environment values", () => {
-        process.env["APPSIGNAL_DEBUG"] = "true"
-        process.env["APPSIGNAL_ENABLE_STATSD"] = "true"
+      it("uses the overwritten path", () => {
+        expect(env("_APPSIGNAL_LOG_FILE_PATH")).toEqual(
```
unflxw, line 128: Nitpick: these expectations could be done against the config object's `.logFilePath` method, instead of fetching the value from the environment (there's tests checking that it is being fetched from the environment later in the file)

unflxw (approved):
Nice cleanup! 👍

jeffkreeftmeijer (approved):

PR #497, opened 2021-11-11
Print config opts camelcase

### LINKED ISSUE #491: Diagnose report should print config option in camelCase
Currently we print the config options as underscored versions of the config options in the diagnose report. This is great for making the output the same for all integrations, but it's not great user experience for users that configure `ignoreErrors` and see `ignore_errors` in the output.

The output printed by the diagnose report should match the config option names as the user configures.

We will still send the config options as their underscored versions in the diagnose report JSON.

## Example

```js
// Currently printed as
ignore_errors: ["MyError"]

// What it should be
ignoreErrors: ["MyError"]
```

## To do

- [ ] Print config option as their camelCase names.
- [ ] Update diagnose tests for Node.js to expect the camelCase names.

luismiramirez:
## Use camelcase for config opts in report 

Diagnose report now displays config options written in camelcase, and
the diagnose JSON still sends them written in snake case.

The transformation functions `getConfigData()` and `getSources` are
now only used to build the JSON.

`Diagnose` object holds sources and options in camel case directly from
the `Configuration` object.

## Update diagnose tests to fit with camel case opts

As diagnose object config options are now written in camel case, tests
need an update for it.

Closes #491

### DIFF
--- a/packages/nodejs/.changesets/diagnose-report-config-opts-are-now-printed-in-camel-case-format.md
+++ b/packages/nodejs/.changesets/diagnose-report-config-opts-are-now-printed-in-camel-case-format.md
@@ -1,0 +1,6 @@
+---
+bump: "patch"
+---
+
+Diagnose report config opts are now printed in camel case format,
+matching how the configuration options are provided to the `Appsignal` object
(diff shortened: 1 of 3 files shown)


tombruijn (approved):

unflxw (approved):
Good stuff! 💯 I thought we would have to convert from the snake case in the report JSON to the camel case and back. This is much more simple.

PR #499, opened 2021-11-12
Add AppSignal client to global object

luismiramirez:
The client entity is added to the global object under the name of
`__APPSIGNAL__` after initialization.

This helps to not to pass the client instance everywhere to perform
operations.

Closes #499

### DIFF
--- a/packages/nodejs/.changesets/appsignal-client-is-available-in-global-object.md
+++ b/packages/nodejs/.changesets/appsignal-client-is-available-in-global-object.md
@@ -1,0 +1,6 @@
+---
+bump: "patch"
+---
+
+The AppSignal client is stored on the global object as to make
+access easier for AppSignal developers.
--- a/packages/nodejs/src/__tests__/client.test.ts
+++ b/packages/nodejs/src/__tests__/client.test.ts
@@ -28,6 +28,10 @@ describe("BaseClient", () => {
     expect(client.isActive).toBeFalsy()
   })
 
+  it("stores the client on global object", () => {
+    expect(global.__APPSIGNAL__).toEqual(client)
+  })
+
   it("does not start the client if config is not valid", () => {
     process.env["APPSIGNAL_PUSH_API_KEY"] = undefined
     client = new BaseClient({ name, enableMinutelyProbes: false })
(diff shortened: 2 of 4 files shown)


packages/nodejs/.changesets/appsignal-client-is-available-in-global-object.md
```diff
@@ -0,0 +1,5 @@
+---
+bump: "patch"
+---
+
+AppSignal's client is now available in global object as: `__APPSIGNAL__`
```
unflxw, line 5: Do we want this to be part of the documented public interface? If we don't want people to rely on this, then we probably shouldn't mention it in the changelog.

unflxw (approved):
Looks good to me!

Last time we discussed this, we considered whether we should support other engines such as Deno. It seems like [the global object in Deno is not `global`, but `window`](https://doc.deno.land/builtin/stable#window). In any case, I think the global object is the least of our concerns, regarding potential Deno support, since [Deno's FFI](https://deno.land/manual@v1.15.3/runtime/ffi_api#foreign-function-interface-api) is different to the [Node-API](https://nodejs.org/api/n-api.html) one. I don't think we should worry about this for now.

packages/nodejs/.changesets/appsignal-client-is-available-in-global-object.md
luismiramirez: I'm not sure, I don't see it as a bad thing

tombruijn: I don't think we want to advertise where we store it. Users shouldn't access it directly like `global.__APPSIGNAL__`. If we want to allow access it should be a helper function like `Appsignal.config()` to fetch it from the config object.

We should make a helper function for that, but users don't need to know about it at the moment.

```suggestion
The AppSignal client is stored on the global object as to make access easier for AppSignal developers.
```

luismiramirez: OK, as the issue states, will add the accessor on a future PR

jeffkreeftmeijer (approved):

PR #504, opened 2021-11-17
Change global augmentation for newer TS

luismiramirez:
Newer TypeScript versions don't accept the way we augment the global
object making the build fail sometimes. Doing it this way, we ensure it
always works.

### DIFF
--- a/packages/nodejs/.changesets/fix-compatibility-for-global-object-augmentation.md
+++ b/packages/nodejs/.changesets/fix-compatibility-for-global-object-augmentation.md
@@ -1,0 +1,5 @@
+---
+bump: "patch"
+---
+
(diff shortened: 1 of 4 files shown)


packages/nodejs/.changesets/fix-compatibility-for-global-object-augmentation.md
```diff
@@ -0,0 +1,5 @@
+---
+bump: "patch"
+---
+
+Fix compatibility for global object augmentation
```
tombruijn, line 5:
```suggestion
Fix TypeScript compatibility for global object augmentation.
```

tombruijn (approved):

unflxw (approved):

PR #506, opened 2021-11-19
Add request headers allowlist

### LINKED ISSUE #487: The requestHeaders config option is not implemented
## The problem

The `requestHeaders` config option is an _allowlist_ of the headers the users want to see in AppSignal for their requests.

The option is not implemented. It is present in the [config maps](https://github.com/appsignal/appsignal-nodejs/blob/main/packages/nodejs/src/config/configmap.ts), but it doesn't have default values nor is checked in the diagnose tests.

## The proposal

- [ ] Add a default value for the new option and make it user-configurable
- [ ] Send the headers with the key `environment` in the spans sample data after filtering them using the config option
- [ ] Update the docs to let the users know about the new option and how it works

luismiramirez:
## Add requestHeaders config option and defaults

`requestHeaders` is a configurable option now. Default values are based
on the other integrations using Node.js naming conventions for them.

## Add requestHeaders allow list feature

`requestHeaders` allow list is now implemented. Now users are able to
define a list of headers they want to be shown in AppSignal. All
integrations except from Express make use of the core HTTP
instrumentation.

As we don't have helpers yet to access the Client stored in global
object, the code is repeated in HTTP instrumentation, and in Express
integration. This will be removed in favor of global helpers in the
future.

Closes #487

### DIFF
--- a/packages/express/.changesets/requestheaders-config-option-is-now-available.md
+++ b/packages/express/.changesets/requestheaders-config-option-is-now-available.md
@@ -1,0 +1,6 @@
+---
+bump: "patch"
+---
+
+Collect request headers from express requests by default. See also our new `requestHeaders` config
+option, added in the @appsignal/nodejs package.
--- a/packages/express/src/middleware/index.ts
+++ b/packages/express/src/middleware/index.ts
@@ -1,5 +1,5 @@
 import { Client } from "@appsignal/nodejs"
-
+import { HashMap } from "@appsignal/types"
 import {
   Request,
   Response,
@@ -7,6 +7,7 @@ import {
   ErrorRequestHandler,
   RequestHandler
 } from "express"
+import { IncomingHttpHeaders } from "http"
 
 /**
  * Returns an Express middleware that can augment the current span
@@ -33,7 +34,8 @@ export function expressMiddleware(appsignal: Client): RequestHandler {
       res.end = function (this: Response) {
         res.end = originalEnd
 
-        const { method = "GET", params = {}, query = {} } = req
+        const { method = "GET", params = {}, query = {}, headers } = req
+        const filteredHeaders = filterHeaders(headers, appsignal)
 
         // if there is no error passed to `next()`, the span name will
         // be updated to match the current path
@@ -48,6 +50,7 @@ export function expressMiddleware(appsignal: Client): RequestHandler {
         // @TODO: keep an eye on this
         // @ts-ignore
         span.setSampleData("params", { ...params, ...query })
+        span.setSampleData("environment", filteredHeaders)
 
         return res.end.apply(this, arguments as any)
       }
@@ -79,4 +82,20 @@ export function expressErrorHandler(appsignal: Client): ErrorRequestHandler {
 
     return next(err)
   }
+}
+
+function filterHeaders(
+  headers: IncomingHttpHeaders,
+  appsignal: Client
+): HashMap<any> {
+  let filtered: HashMap<any> = {}
+  const headersAllowList = appsignal.config.data.requestHeaders || []
+
+  headersAllowList.forEach(key => {
+    if (headers[key] != undefined) {
+      filtered[key] = headers[key]
+    }
+  })
+
+  return filtered
 }
--- a/packages/nodejs/.changesets/requestheaders-config-option-is-now-available.md
+++ b/packages/nodejs/.changesets/requestheaders-config-option-is-now-available.md
@@ -1,0 +1,8 @@
+---
+bump: "patch"
+---
+
+The `requestHeaders` config option is now available. An allow list that gives the ability to define
+which request headers you want to be shown in sample detail views. The default is a list of common
+headers that do not include [personal identifiable information](https://docs.appsignal.com/appsignal/gdpr.html#allowed-request-headers-only).
+Read more about [request headers](https://docs.appsignal.com/application/header-filtering.html) on our documentation website.
--- a/packages/nodejs/src/cli/diagnose.ts
+++ b/packages/nodejs/src/cli/diagnose.ts
@@ -416,5 +416,9 @@ export class Diagnose {
 }
 
 function format_value(value: any) {
-  return util.inspect(value)
+  if (typeof value == "object") {
+    return JSON.stringify(value)
+  } else {
+    return util.inspect(value)
+  }
 }
--- a/packages/nodejs/src/config.ts
+++ b/packages/nodejs/src/config.ts
@@ -128,13 +128,7 @@ export class Configuration {
         "cache-control",
         "connection",
         "content-length",
-        "path-info",
-        "range",
-        "request-method",
-        "request-uri",
-        "server-name",
-        "server-port",
-        "server-protocol"
+        "range"
       ],
       transactionDebugMode: false
     }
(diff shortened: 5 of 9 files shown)


packages/nodejs/.changesets/requestheaders-config-option-is-now-available.md
```diff
@@ -0,0 +1,7 @@
+---
+bump: "patch"
+---
+
+`requestHeaders` config option is now available. An allow list that
+gives the ability to define which request headers you want to be
+shown in AppSignal performance traces.
```
tombruijn, line 7:
```suggestion
The `requestHeaders` config option is now available. An allow list that gives the ability to define which request headers you want to be shown in sample detail views. The default is a list of common headers that do not include [personal identifiable information](https://docs.appsignal.com/appsignal/gdpr.html#allowed-request-headers-only). Read more about [request headers](https://docs.appsignal.com/application/header-filtering.html) on our documentation website.
```

packages/express/.changesets/requestheaders-config-option-is-now-available.md
```diff
@@ -0,0 +1,7 @@
+---
+bump: "patch"
+---
+
+`requestHeaders` config option is now available. An allow list that
+gives the ability to define which request headers you want to be shown
+in AppSignal performance traces.
```
tombruijn, line 7:
```suggestion
Collect request headers from express requests by default. See also our new `requestHeaders` config option, added in the @appsignal/nodejs package.
```

packages/express/src/middleware/index.ts
```diff
@@ -80,3 +83,16 @@ export function expressErrorHandler(appsignal: Client): ErrorRequestHandler {
     return next(err)
   }
 }
+
+function filterHeaders(headers: IncomingHttpHeaders): HashMap<any> {
+  let filtered: HashMap<any> = {}
+  const headersAllowList = global.__APPSIGNAL__.config.data.requestHeaders || []
```
tombruijn, line 89: Let's use the client's config object to fetch the request headers for now. Until we have an accessor for the client/config on on the Appsignal module. That way we hopefully don't need to add the global type definition in this package as well.

We should have the client available here, because it's given to the `expressMiddleware` function.

packages/nodejs/src/cli/diagnose.ts
```diff
@@ -416,5 +416,9 @@ export class Diagnose {
 }
 
 function format_value(value: any) {
-  return util.inspect(value)
+  if (typeof value == "object") {
```
tombruijn, line 419:
```suggestion
  if (typeof value === "object") {
```

packages/nodejs/src/instrumentation/http/lifecycle/incoming.ts
```diff
@@ -83,6 +86,19 @@ function incomingRequest(
   }
 }
 
+function filterHeaders(headers: IncomingHttpHeaders): HashMap<any> {
+  let filtered: HashMap<any> = {}
+  const headersAllowList = global.__APPSIGNAL__.config.data.requestHeaders || []
+
+  headersAllowList.forEach(key => {
+    if (headers[key] != undefined) {
```
tombruijn, line 94:
```suggestion
    if (headers[key] !== undefined) {
```

tombruijn (changes requested):

packages/nodejs/src/cli/diagnose.ts
luismiramirez: Thanks to the TS compiler we don't need to write equality operators as in vanilla JS

packages/nodejs/src/instrumentation/http/lifecycle/incoming.ts
luismiramirez: Thanks to the TS compiler we don't need to write equality operators as in vanilla JS

packages/express/src/middleware/index.ts
luismiramirez: Yup, noted that in the commit.

unflxw (approved):
A random idea that may not be plausible/desirable: to avoid having to implement `filterHeaders` twice, in Express and in the integration, maybe the filtering could happen in `span.setSampleData` when the key is `"environment"`?

In a way, that would match the direction we're going, regarding moving the filtering behaviour to when the sample data is set in the agent -- except here it would be the integration side doing it _before_ passing it to the agent, at least for now, since the agent does not support allowlists yet.

tombruijn:
> A random idea that may not be plausible/desirable: to avoid having to implement filterHeaders twice, in Express and in the integration, maybe the filtering could happen in span.setSampleData when the key is "environment"?

That could cause a problem for any non-header data we store in the environment. We don't do that in the Node.js integration at the moment, but we might later on.

packages/express/src/middleware/index.ts
tombruijn: Yes, but let's not access it directly on the global object here until we have an accessor. We have the client object here already, we don't need to fetch it from the global object.

jeffkreeftmeijer (approved):

packages/express/global.d.ts
```diff
@@ -0,0 +1,7 @@
+export {}
+
+import { Client } from "@appsignal/nodejs"
+
+declare global {
+  var __APPSIGNAL__: Client
+}
```
tombruijn, line 7: I think this can be removed now that we don't access it on global anymore?

packages/nodejs/src/cli/diagnose.ts
tombruijn: It's not good practice to keep doing this? Does TS give you a warning if it could ever be ambiguous?

tombruijn:
> A random idea that may not be plausible/desirable: to avoid having to implement filterHeaders twice, in Express and in the integration, maybe the filtering could happen in span.setSampleData when the key is "environment"?

Reading this back I do think it's nice if we can reuse this function. Export it from the Node.js package for example, and call it in the express package. We can do this in another PR later.

luismiramirez:
> It's not good practice to keep doing this? Does TS give you a warning if it could ever be ambiguous?

No, it only throws a compile error if you're checking different types when using `==`

packages/nodejs/src/config.ts
```diff
@@ -120,6 +120,22 @@ export class Configuration {
...
+        "cache-control",
+        "connection",
+        "content-length",
+        "path-info",
+        "range",
+        "request-method",
+        "request-uri",
+        "server-name",
+        "server-port",
+        "server-protocol"
```
tombruijn, line 137:
```suggestion
        "accept",
        "accept-charset",
        "accept-encoding",
        "accept-language",
        "cache-control",
        "connection",
        "content-length",
        "range"
```

I can't actually find the headers I removed in these two sources:

- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers
- https://en.wikipedia.org/wiki/List_of_HTTP_header_fields

Might they be legacy fields from the Ruby gem's Rack "environment" values? Or are some of these names custom headers passed along by a proxy server?

tombruijn: I noticed this because the list of headers included in the sample is quite short: 
image

luismiramirez: They're from the Elixir integration. Removing them.

luismiramirez: We'll probably need to remove them from the Elixir integration as well. Because I think you're right, they look like legacy values from the Ruby gem.

tombruijn: Ok! Creating an issue.

tombruijn (approved):

PR #512, opened 2021-11-23
Fix writable path check in diagnose

tombruijn:
The `fs.accessSync` in the `isWriteableFile` function only checked if
the path was readable with `fs.constants.R_OK`. Not if it was writable
with `fs.constants.W_OK`.

> - `F_OK`: Flag indicating that the file is visible to the calling
>    process. This is useful for determining if a file exists, but says
>    nothing about rwx permissions. Default if no mode is specified.
> - `R_OK`: Flag indicating that the file can be read by the calling
>    process.
> - `W_OK`: Flag indicating that the file can be written by the calling
>    process.
> - `X_OK`: Flag indicating that the file can be executed by the
>    calling process. This has no effect on Windows (will behave like
>    fs.constants.F_OK).

Source: https://nodejs.org/api/fs.html#file-access-constants

I've also renamed the function to `isWriteable` because it doesn't only
check files, but also directories.

### DIFF
--- a/packages/nodejs/.changesets/fix-writable-check-in-diagnose-report.md
+++ b/packages/nodejs/.changesets/fix-writable-check-in-diagnose-report.md
@@ -1,0 +1,5 @@
+---
+bump: "patch"
+---
+
+Fix writable check for paths in the diagnose report. Previously it only checked if a path was readable, not writable.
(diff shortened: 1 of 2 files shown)


unflxw (approved):

luismiramirez (approved):
🤦🏼 I swear my brain was making me see a `W` instead of a `R`

Issue #517, opened 2021-11-25
The caFilePath option isn't used by the diagnose transmitter or Push API key validator


tombruijn:
The `caFilePath` config option is currently only configured to be used by the agent, but the diagnose report transmitter and Push API key validator do not use this config option's value.

## To do

- [ ] Update the [diagnose report transmitter](https://github.com/appsignal/appsignal-nodejs/blob/1a2bef0a710ff9b40129fc90aa25bd2f4230f400/packages/nodejs/src/diagnose.ts#L263) to use the config option value. It's currently hardcoded.
- [ ] Update the [Push API key validator](https://github.com/appsignal/appsignal-nodejs/blob/1a2bef0a710ff9b40129fc90aa25bd2f4230f400/packages/nodejs/src/diagnose.ts#L124) to use the config option value. It's currently not configured at all.
- [ ] Check the difference between the two transmitters and see if they can't be merged into one reusable transmitter. Related to https://github.com/appsignal/appsignal-nodejs/issues/518

PR #531, opened 2021-12-10
Allow transmitters to use cafile

luismiramirez:
## Create a single transmitter for Diagnose

Diagnose script performs two external HTTP calls, one for push api key
validation, and another one to send the diagnose results to AppSignal if
the user wants.

These calls were written separatedly inside the diagnose script, now
they are done through a new class called `Transmitter`.

## Allow transmitter calls to use caFilePath

`caFilePath` config option was only used by the agent. With this change,
the certificate file will also be used by the diagnose transmitter to
validate the api key and to send the diagnose results to AppSignal.

## Protect transmitter JSON parsing 

Sometimes, the server response received by the transmitter is not a
parseable JSON. Added a try/catch to protect the script.

Closes: #517

### DIFF
--- a/packages/nodejs/src/diagnose.ts
+++ b/packages/nodejs/src/diagnose.ts
@@ -12,6 +12,7 @@ import { AGENT_VERSION, VERSION } from "./version"
 import { JS_TO_RUBY_MAPPING } from "./config/configmap"
 import { AppsignalOptions } from "."
 import { HashMap } from "@appsignal/types"
+import { Transmitter } from "./transmitter"
 
 interface FileMetadata {
   content?: string[]
@@ -112,31 +113,27 @@ export class DiagnoseTool {
   }
 
   private async validatePushApiKey() {
-    return new Promise((resolve, reject) => {
-      const config = this.#config.data
-      const params = new URLSearchParams({
-        api_key: config["pushApiKey"] || "",
-        name: config["name"] || "",
-        environment: config["environment"] || "",
-        hostname: config["hostname"] || ""
-      })
-      const url = new URL(`/1/auth?${params.toString()}`, config["endpoint"])
-      const options = { method: "POST" }
+    const config = this.#config.data
+    const url = new URL(`/1/auth`, config["endpoint"])
+    const transmitter = new Transmitter(url.toString())
+    let statusCode = null
 
-      const requestModule = url.protocol == "http:" ? http : https
-      const request = requestModule.request(url, options, function (response) {
-        const status = response.statusCode
-        if (status === 200) {
-          resolve("valid")
-        } else if (status === 401) {
-          reject("invalid")
-        } else {
-          reject(`Failed to validate: status ${status}`)
-        }
+    await transmitter
+      .transmit()
+      .then(responseData => {
+        statusCode = responseData["status"]
       })
-      request.write("") // Send empty body
-      request.end()
-    })
+      .catch(responseData => {
+        statusCode = responseData["status"]
+      })
+
+    if (statusCode == 200) {
+      return Promise.resolve("valid")
+    } else if (statusCode == 401) {
+      return Promise.reject("invalid")
+    } else {
+      return Promise.reject(`Failed to validate: status ${statusCode}`)
+    }
   }
 
   private getPathsData() {
@@ -232,63 +229,32 @@ export class DiagnoseTool {
     )
   }
 
-  public sendReport(data: HashMap<any>) {
+  public async sendReport(data: HashMap<any>) {
     data.config.options = this.getConfigData()
     data.config.sources = this.getSources()
     const json = JSON.stringify({ diagnose: data })
 
-    const config = this.#config.data
-    const params = new URLSearchParams({
-      api_key: config["pushApiKey"] || "",
-      name: config["name"] || "",
-      environment: config["environment"] || "",
-      hostname: config["hostname"] || ""
-    })
-
     const diagnoseEndpoint =
       process.env.APPSIGNAL_DIAGNOSE_ENDPOINT || "https://appsignal.com/diag"
-    const url = new URL(diagnoseEndpoint)
 
-    const opts = {
-      method: "POST",
-      protocol: url.protocol,
-      host: url.hostname,
-      port: url.port,
-      path: `${url.pathname}?${params.toString()}`,
-      headers: {
-        "Content-Type": "application/json",
-        "Content-Length": json.length
-      },
-      cert: fs.readFileSync(
-        path.resolve(__dirname, "../cert/cacert.pem"),
-        "utf-8"
-      )
-    }
+    const transmitter = new Transmitter(diagnoseEndpoint, json)
 
-    const requestModule = url.protocol == "http:" ? http : https
-    const request = requestModule.request(opts, (response: any) => {
-      const responseStatus = response.statusCode
-      response.setEncoding("utf8")
-
-      response.on("data", (responseData: any) => {
-        if (responseStatus === 200) {
-          const { token } = JSON.parse(responseData.toString())
-          console.log(`  Your support token:`, token)
-          console.log(
-            `  View this report:   https://appsignal.com/diagnose/${token}`
-          )
-        } else {
-          console.error(
-            "  Error: Something went wrong while submitting the report to AppSignal."
-          )
-          console.error(`  Response code: ${responseStatus}`)
-          console.error(`  Response body:\n${responseData}`)
-        }
+    await transmitter
+      .transmit()
+      .then(responseData => {
+        const { token } = responseData["body"]
+        console.log(`  Your support token:`, token)
+        console.log(
+          `  View this report:   https://appsignal.com/diagnose/${token}`
+        )
       })
-    })
-
-    request.write(json)
-    request.end()
+      .catch(responseData => {
+        console.error(
+          "  Error: Something went wrong while submitting the report to AppSignal."
+        )
+        console.error(`  Response code: ${responseData["status"]}`)
+        console.error(`  Response body:\n${responseData["body"]}`)
+      })
   }
 }
 
--- a/packages/nodejs/src/diagnose.ts
+++ b/packages/nodejs/src/diagnose.ts
@@ -253,7 +253,9 @@ export class DiagnoseTool {
           "  Error: Something went wrong while submitting the report to AppSignal."
         )
         console.error(`  Response code: ${responseData["status"]}`)
-        console.error(`  Response body:\n${responseData["body"]}`)
+        console.error(
+          `  Response body:\n${JSON.stringify(responseData["body"])}`
+        )
       })
   }
 }
--- a/packages/nodejs/src/transmitter.ts
+++ b/packages/nodejs/src/transmitter.ts
@@ -75,10 +75,22 @@ export class Transmitter {
         "Content-Type": "application/json",
         "Content-Length": this.#data.length
       },
-      cert: fs.readFileSync(
-        path.resolve(__dirname, "../cert/cacert.pem"),
-        "utf-8"
+      ca: this.certificateFile()
+    }
+  }
+
+  private certificateFile(): string {
+    const configData = this.#config.data
+    const caFilePathFromConfig = configData["caFilePath"] || ""
+
+    try {
+      fs.accessSync(caFilePathFromConfig, fs.constants.R_OK)
+      return fs.readFileSync(caFilePathFromConfig, "utf-8").toString()
+    } catch {
+      console.warn(
+        `Provided caFilePath: '${caFilePathFromConfig}' is not readable.`
       )
+      return ""
     }
   }
 }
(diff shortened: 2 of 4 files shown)


packages/nodejs/src/transmitter.ts
```diff
@@ -0,0 +1,78 @@
...
+      method: "POST",
+      protocol: requestUrl.protocol,
+      host: requestUrl.hostname,
+      port: requestUrl.port,
+      path: requestUrl.toString(),
+      headers: {
+        "Content-Type": "application/json",
+        "Content-Length": this.#data.length
+      },
+      cert: fs.readFileSync(configData["caFilePath"] || "", "utf-8")
```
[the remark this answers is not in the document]
unflxw: Additionally, I'm not entirely sure about this, but going by [the docs for the `tls.createSecureContext` function](https://nodejs.org/api/tls.html#tlscreatesecurecontextoptions), I _think_ the contents of the file, which default to a dump of the Mozilla CAs, should be passed as the `ca` option, not `cert`:

> `ca`: Optionally override the trusted CA certificates. Default is to trust the well-known CAs curated by Mozilla. Mozilla's CAs are completely replaced when CAs are explicitly specified using this option. [...]

> `cert`: Cert chains in PEM format. One cert chain should be provided per private key. Each cert chain should consist of the PEM formatted certificate for a provided private `key`, followed by the PEM formatted intermediate certificates (if any), in order, and not including the root CA (the root CA must be pre-known to the peer, see `ca`). [...]

I believe that, since we're not providing any private keys as `key`, the contents of the CA file that we're passing to `cert` are being ignored. I also believe that those contents would not be valid to be passed to `cert` anyway, as they contain root CAs and do not contain intermediate certificates. See `head packages/nodejs/cert/cacert.pem`:

```
## Bundle of CA Root Certificates
##
## Certificate data from Mozilla as of: Wed Jan  1 04:12:10 2020 GMT
##
## This is a bundle of X.509 certificates of public Certificate Authorities
## (CA). These were automatically extracted from Mozilla's root certificates
## file (certdata.txt).
```

Again, I'm not very confident about any of this. TLS is hard.

packages/nodejs/src/transmitter.ts
```diff
@@ -0,0 +1,78 @@
...
+      const opts = this.requestOptions(url)
+      const requestModule = url.protocol == "http:" ? http : https
+
+      const request = requestModule.request(opts, (response: any) => {
+        const responseStatus = response.statusCode
+        response.setEncoding("utf8")
+        let responseBody = {}
+
+        response
+          .on("data", (responseData: any) => {
```
unflxw, line 41: Is the response data received in the `"data"` event complete? I can't find any authoritative source on this, because bafflingly, the Node.js docs don't seem to document the `"data"` event for [the `http.IncomingRequest` object](https://nodejs.org/api/http.html#class-httpincomingmessage) at all, or at least I can't find where it is documented. But looking at examples sprinkled on the page, it seems like the argument passed to the `"data"` event is a partial chunk of the response, and you're responsible for collecting those chunks into an array or buffer or a string and piecing them together. Quoting [one of the examples sprinkled on the docs](https://nodejs.org/api/http.html#httpgeturl-options-callback):

```js
http.get('http://localhost:8000/', (res) => {
  // [... some stuff we don't care about ...]
  let rawData = '';
  res.on('data', (chunk) => { rawData += chunk; });
  // [...]
}
```

Of course, it doesn't help understand what's going on that many of the examples [[1]](https://nodejs.dev/learn/make-an-http-post-request-using-nodejs) [[2]](https://nodejs.dev/learn/making-http-requests-with-nodejs) [[3]](https://nodejs.org/api/http.html#event-connect) just take the data _chunk_ and push it to standard output, without mentioning that it's a _chunk_ of data, not the full thing, and doing anything with it by itself won't be useful for most usecases. 

Anyway, I think you need to collect the chunks every time the `"data"` event is called, then attempt to `JSON.parse` it on the `"end"` event. Such fun.

unflxw (changes requested):
The core of what this PR is trying to do looks good to me, but I _think_ the implementation in this PR is broken. This is a natural consequence of the Node.js documentation being bafflingly vague about details that really matter and leaving you to infer the actual behaviour through examples.

packages/nodejs/src/transmitter.ts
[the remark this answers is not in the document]
luismiramirez: > Won't this always fail if configData["caFilePath"] is undefined? It seems like fs.readFileSync("", "utf-8") will always throw an ENOENT error.

`configData["caFilePath"]` should never be undefined as it has a default. It'll only be undefined if the users decide it. If that happens, making the script fail is OK to me.

Wow, we were already using `cert` before these changes, now I know why all my tests passed with this. Gonna try using the `ca` option.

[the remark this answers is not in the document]
unflxw: A good way to test would be to replace `cacert.pem` with an empty file -- if the CAs are being replaced correctly with those on the file, the request should fail.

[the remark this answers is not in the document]
luismiramirez: I've tested it and everything passes 🤔

packages/nodejs/src/transmitter.ts
luismiramirez: Sorted out in the latest changes. Now the response is treated as a stream.

packages/nodejs/src/transmitter.ts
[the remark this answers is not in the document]
luismiramirez: Empty files, empty strings, `undefined`, and `null` values are bypassed using the default certificates.

Now we use `ca` instead of `cert`, allowing it to fail if the file has the wrong contents.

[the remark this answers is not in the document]
tombruijn: > configData["caFilePath"] should never be undefined

It can be set by the user (which is why we listen to it here), and also set to an empty value. Which means we should check if the config option is set, the file exists and is a readable file, as documented here: https://github.com/appsignal/integration-guide/pull/104

If I now do this I get an error:

```
$ export APPSIGNAL_CA_FILE_PATH="/foo"
$ bin/diagnose
[...]
  Send diagnostics report to AppSignal? (Y/n): y
  Transmitting diagnostics report

  Error: Something went wrong while submitting the report to AppSignal.
  Response code: undefined
  Response body:
undefined
```

With an empty value:

```
$ export APPSIGNAL_CA_FILE_PATH=""
$ bin/diagnose
[...]
  Send diagnostics report to AppSignal? (Y/n): y
  Transmitting diagnostics report

  Your support token: token
  View this report:   https://appsignal.com/diagnose/token
```

In the Ruby gem, in both scenarios, the config option value is ignored because it's invalid.

packages/nodejs/src/diagnose.ts
```diff
@@ -112,30 +113,28 @@ export class DiagnoseTool {
...
-        } else {
-          reject(`Failed to validate: status ${status}`)
-        }
+      .catch(responseData => {
+        statusCode = responseData["status"]
       })
-      request.write("") // Send empty body
-      request.end()
+
+    return new Promise((resolve, reject) => {
```
tombruijn, line 130: It feels a bit weird to create and immediately return a promise here. You can also directly return `Promise.resolve("valid")` or `Promise.reject("invalid")` here without the overhead of the promise function.

You probably didn't add the resolve and reject to `transmitter.transmit()` that now `await`s, because both the `then` and `catch` fetch the statuscode?

packages/nodejs/src/transmitter.ts
```diff
@@ -0,0 +1,81 @@
...
+            }
+
+            if (responseStatus == 200) {
+              resolve({ status: responseStatus, body: parsedResponse })
+            } else {
+              reject({ status: responseStatus, body: parsedResponse })
+            }
+          })
+      })
+
```
tombruijn, line 59: If we're improving things, can we also listen to the `request.on("error", ..)` event? For when there's a problem with making the request, and not just the response. This can also be in a separate PR.

tombruijn (changes requested):

packages/nodejs/src/transmitter.ts
luismiramirez: Let's take care of it in a separate PR then

packages/nodejs/src/diagnose.ts
luismiramirez: > It feels a bit weird to create and immediately return a promise here. You can also directly return Promise.resolve("valid") or Promise.reject("invalid") here without the overhead of the promise function.

I'll use those helpers as return values, thank you

> You probably didn't add the resolve and reject to transmitter.transmit() that now awaits, because both the then and catch fetch the statuscode?

Yep, that's correct

packages/nodejs/src/transmitter.ts
tombruijn: I created an issue for it: https://github.com/appsignal/appsignal-nodejs/issues/535

luismiramirez: Good, let's do it after this one is merged.

packages/nodejs/src/transmitter.ts
```diff
@@ -0,0 +1,96 @@
...
+    const caFilePathFromConfig = configData["caFilePath"] || ""
+    let caFilePath
+
+    if (caFilePathFromConfig && fs.existsSync(caFilePathFromConfig)) {
+      return fs.readFileSync(caFilePathFromConfig, "utf-8").toString()
+    } else {
+      console.warn(
+        `Provided caFilePath: '${caFilePathFromConfig}' is not readable.`
+      )
+      return ""
```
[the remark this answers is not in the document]
tombruijn: Yes, that is fine. It should only set it if the file exists and is readable. I've clarified this in the reference here: https://github.com/appsignal/integration-guide/pull/104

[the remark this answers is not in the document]
luismiramirez: Yes, this is done on purpose based on the latest integration-guide updates as @tombruijn pointed.

packages/nodejs/src/transmitter.ts
```diff
@@ -0,0 +1,96 @@
...
+        "Content-Length": this.#data.length
+      },
+      ca: this.certificateFile()
+    }
+  }
+
+  private certificateFile(): string {
+    const configData = this.#config.data
+    const caFilePathFromConfig = configData["caFilePath"] || ""
+    let caFilePath
```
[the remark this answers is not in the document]
luismiramirez: I updated it to `let statusCode = null` on the latest changes. This must be here because the variable value is assigned from an async function using `await`. Our TS config is strict, so `strictNullChecks` are activated by default. We have to initialize these variables as "something" (`null` in this case) to make the compiler happy.

packages/nodejs/src/transmitter.ts
```diff
@@ -0,0 +1,96 @@
...
+
+    if (caFilePathFromConfig && fs.existsSync(caFilePathFromConfig)) {
+      return fs.readFileSync(caFilePathFromConfig, "utf-8").toString()
```
tombruijn, line 88: This still throw an error if the file can't be read I think? Can we check the readability permissions beforehand, or catch any error that occurs when reading the file.

packages/nodejs/src/transmitter.ts
```diff
@@ -0,0 +1,96 @@
...
+  #url: string
+  #data: string
+
+  constructor(url: string, data = "", active = true) {
+    this.#config = new Configuration({ active })
+    this.#url = url
+    this.#data = data
+  }
+
+  public async transmit() {
```
tombruijn, line 21: If we return a promise, we don't need to make it async. We also don't call `await` anywhere in the function so I think we can remove it. That also means you can remove `await` anywhere we call `transmitter.transmit()`

This is a similar function that doesn't need async: https://github.com/appsignal/appsignal-nodejs/blob/402a364305777bda77fc71b9114d0429595b2cd5/packages/nodejs-ext/scripts/extension.js#L44-L45

tombruijn (changes requested):

packages/nodejs/src/transmitter.ts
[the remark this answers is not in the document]
tombruijn: But we don't use the variable anywhere right? So we can remove this entire line.

packages/nodejs/src/transmitter.ts
luismiramirez: It has to be async because of the `validatePushApiKey()` function. It returns promises based on the result of `transmit()` and if we don't call it with `await`, it always returns `undefined`.

I removed the `await` call from `sendReport()`

tombruijn: I'm still confused about the whole async-await in combination with promises, but I think what I said before is wrong, in this context. If we don't await on `transmitter.transmit()` in `sendReport` it continues the program execution, but because it's the last thing it does, it waits on the promise before exiting. I think calling await on everywhere that we call `transmitter.transmit()` was correct before. 

Every async function returns a promise, even if we don't explicitly wrap the return value in one apparently.
I read that one would usually not need both async-await and promises with `then`-calls but I don't see how just yet, we also need to "catch" failure states.

tombruijn (approved):
Looks good! Please remove the unused `let caFilePath` as well.

PR #534, opened 2021-12-13
Add helper for globally stored client

luismiramirez:
A new static helper method is available from `BaseClient` (exported as
`Appsignal`). It allows access to the globally stored client when
AppSignal is initialized.

This will give us the ability to develop missing features that are based
on config options.

### DIFF
--- a/packages/nodejs/.changesets/add-helper-to-access-the-globally-stored-appsignal-client.md
+++ b/packages/nodejs/.changesets/add-helper-to-access-the-globally-stored-appsignal-client.md
@@ -1,0 +1,6 @@
+---
+bump: "patch"
+type: "add"
+---
+
+Add helper to access the globally stored AppSignal client.
--- a/packages/nodejs/src/client.ts
+++ b/packages/nodejs/src/client.ts
@@ -29,6 +29,17 @@ export class BaseClient implements Client {
   #metrics: Metrics = new BaseMetrics()
 
   /**
+   * Global accessors to Client and Config
+   */
+  static get client(): Client {
+    return global.__APPSIGNAL__
+  }
+
+  static get config(): Configuration {
+    return global.__APPSIGNAL__?.config
+  }
+
+  /**
    * Creates a new instance of the `Appsignal` object
    */
   constructor(options: Partial<AppsignalOptions> = {}) {
--- a/packages/nodejs/src/__tests__/client.test.ts
+++ b/packages/nodejs/src/__tests__/client.test.ts
@@ -32,6 +32,14 @@ describe("BaseClient", () => {
     expect(global.__APPSIGNAL__).toEqual(client)
   })
 
+  it("returns the client from global object", () => {
+    expect(BaseClient.client).toEqual(client)
+  })
+
+  it("returns the client config from global object", () => {
+    expect(BaseClient.config).toEqual(client.config)
+  })
+
   it("does not start the client if config is not valid", () => {
     process.env["APPSIGNAL_PUSH_API_KEY"] = undefined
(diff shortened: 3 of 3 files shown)


packages/nodejs/.changesets/add-helper-to-access-the-globally-stored-appsignal-client.md
```diff
@@ -0,0 +1,6 @@
+---
+bump: "patch"
+type: "add"
+---
+
+Add helper to access the globally stored AppSignal client
```
tombruijn, line 6:
```suggestion
Add helper to access the globally stored AppSignal client.
```

packages/nodejs/src/client.ts
```diff
@@ -49,6 +49,13 @@ export class BaseClient implements Client {
     this.storeInGlobal()
   }
 
+  /**
+   * Returns the global stored client
+   */
+  static client(): Client {
+    return global.__APPSIGNAL__
```
tombruijn, line 56: How can I best access this? With `BaseClient.client()` or `Appsignal.client()`?
Would it make sense to make this a getter?

luismiramirez: This is currently accessed internally (within the Node.js package) with `BaseClient.client()`, and with `Appsignal.client()` from the outside.

Yes, making it a setter would be more convenient. This is just the accessor for the implementation of the missing config opt. Customers will not need to use this function once we have the new API helpers in the future.

One option is to create an accessor just for the config like `Appsignal.config` | `BaseClient.config`

jeffkreeftmeijer (approved):

unflxw (approved):

packages/nodejs/src/client.ts
luismiramirez: Yes, I meant "getter", sorry. 

I'll create a static getter for the client.

> To reduce one additional step of trying to access the config on the client you mean?

Yes, this may be convenient, cause internally, we'll want to access the client's config 99% of the time

tombruijn: > Yes, this may be convenient, cause internally, we'll want to access the client's config 99% of the time

Ok, so we'll have two getters, one for the client, one for the config. Sounds good.

packages/nodejs/src/client.ts
```diff
@@ -28,6 +28,12 @@ export class BaseClient implements Client {
   #tracer: Tracer = new BaseTracer()
   #metrics: Metrics = new BaseMetrics()
 
+  /**
+   * Global accessors to Client and Config
+   */
+  static client: Client = global.__APPSIGNAL__
```
tombruijn, line 34: If I call `Appsignal.client` in a test app it returns `undefined`. Could be that it doesn't update when we overwrite the value?

```
const { Appsignal } = require("@appsignal/nodejs");
  "active": true,
});

console.log("!!! global", Appsignal.client);
// => undefined
```

unflxw: I think it has to be a getter, otherwise it'll grab the value of `global.__APPSIGNAL__` when the class is defined.

luismiramirez: Yep, it gets the value when the class is defined but it is overwritten every time the config is initialized on `storeInGlobal()` and it worked on my tests. I'll do a couple more, if it doesn't work we should go with the static functions.

luismiramirez: The results are not consistent. I would go with static functions instead to avoid future bugs.

luismiramirez:
Feedback applied, merging this one to continue with the implementation of `sendParams`

PR #537, opened 2021-12-27
Implement sendParams config option

### LINKED ISSUE #514: Config option sendParams is not implemented
The config option is not used anywhere in the Node.js integration. 
Look at the other integrations and implement it in the same way.

When this config option is set, it should not add "params" sample data to any span.

luismiramirez:
## Add sendParams config option

The `sendParams` config option is now available and its default value is
`true`. It'll be used to globally determine if the users want to send
request params in the spans or not.

## Implement sendParams config option

The `sendParams` config option allows the users to decide if they want
to send request params to AppSignal. Its default value is `true`.

`Span#setSampleData()` now checks if the `sendParams` option is enabled
when sending `params` to decide if they're going to be sent to AppSignal
or not.

Closes #514

### DIFF
--- a/packages/nodejs/src/config/configmap.ts
+++ b/packages/nodejs/src/config/configmap.ts
@@ -88,6 +88,7 @@ export const JS_TO_RUBY_MAPPING: { [key: string]: string } = {
   revision: "revision",
   runningInContainer: "running_in_container",
   sendEnvironmentMetadata: "send_environment_metadata",
+  sendParams: "send_params",
   transactionDebugMode: "transaction_debug_mode",
   workingDirPath: "working_dir_path",
   workingDirectoryPath: "working_directory_path"
--- a/packages/nodejs/src/span.ts
+++ b/packages/nodejs/src/span.ts
@@ -5,6 +5,7 @@ import { Span, SpanOptions, SpanContext } from "./interfaces"
 import { span } from "./extension_wrapper"
 import { Data } from "./internal/data"
 import { getAgentTimestamps } from "./utils"
+import { BaseClient } from "./client"
 
 /**
  * The `Span` object represents a length of time in the flow of execution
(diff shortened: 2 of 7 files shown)


packages/nodejs/src/config/configmap.ts
```diff
@@ -56,6 +56,7 @@ export const PRIVATE_ENV_MAPPING: { [key: string]: string } = {
   _APPSIGNAL_PUSH_API_KEY: "pushApiKey",
   _APPSIGNAL_RUNNING_IN_CONTAINER: "runningInContainer",
   _APPSIGNAL_SEND_ENVIRONMENT_METADATA: "sendEnvironmentMetadata",
+  _APPSIGNAL_SEND_PARAMS: "sendParams",
```
unflxw, line 59: Does the agent need to know about this? If not, we don't need to define the environment variable (and we don't need to test for its presence, either)

unflxw (approved):

packages/nodejs/src/config/configmap.ts
luismiramirez: Not at this moment. I'll remove the private env var assignation from this PR, but as we talked, we might consider moving this feature to the agent.

jeffkreeftmeijer (approved):

packages/nodejs/src/span.ts
```diff
@@ -125,6 +126,7 @@ export class BaseSpan implements Span {
         >
   ): this {
     if (!key || !data) return this
+    if (key == "params" && !BaseClient.config.data.sendParams) return this
```
tombruijn, line 129: Could this ever be `undefined`? If a user hasn't initialized the config yet for example.

luismiramirez: No user should call this function without a configuration initialized. Spans are born from the Tracer object, which needs an initialized config. It'll fail before getting to this condition.

tombruijn: Look like it would break before it arrives here, with the integrations calling `tracer()` on an empty/undefined AppSignal client trying to start a span. When we rely on the globally stored client more we could return a "noop" client object if needed. Let's revisit it later if needed.

tombruijn (approved):

PR #538, opened 2021-12-27
Check if the process is running in a container from the extension

luismiramirez:
## Add RunningInContainer function to extension

This function checks if the agent is running in a container. The Ruby
and the Elixir integrations already use this function in the diagnose
report.

This commit makes the function available to be used in the integration.

## Use extension to check if the process is running in container

The runningInContainer check is now run through the extension.

Removed the old code that made the checks using JS.

### DIFF
--- a/packages/nodejs/.changesets/extension-now-checks-running-in-container.md
+++ b/packages/nodejs/.changesets/extension-now-checks-running-in-container.md
@@ -1,0 +1,7 @@
+---
+bump: "patch"
+type: "change"
+---
+
+The extension is now responsible of determining if the process is running in a container. This check
+was previously made by the Node.js integration code.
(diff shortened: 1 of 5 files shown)


packages/nodejs/.changesets/extension-now-checks-running-in-container.md
```diff
@@ -0,0 +1,7 @@
+---
+bump: "patch"
+type: "change"
+---
+
+The extension is now responsible of determining if the process is running in a container. This check
+was previously made by the Node.js integration code.
```
unflxw, line 7: Not sure if this needs any changeset entries? My understanding is that this change is an implementation detail, and functionally it should behave as it did before this change. I don't know if these details about our internals are interesting information for the users. (@tombruijn thoughts?)

(If this fixes something, then we could talk in the changelog about what it fixes, i.e. "The integration now correctly detects it's running in a container when so-and-so happens")

unflxw (approved):

tombruijn (approved):

Issue #535, opened 2021-12-15
Transmitter should listen to error event


tombruijn:
As discussed in https://github.com/appsignal/appsignal-nodejs/pull/531#discussion_r768507341, we should update the transmitter to listen to the "error" event, when the request could not even be made.

PR #539, opened 2021-12-29
Handle request errors on diagnose HTTP transmitter

luismiramirez:
Request errors are now handled on diagnose HTTP external calls (Push API
key validation and report sending).

The promise returned by `Transmitter#transmit` will only reject if
there's some kind of error during the request. Non-200 response codes
are not considered an error anymore.

Output example:

```
Error submitting the report: getaddrinfo ENOTFOUND appsignal.com
```

Closes: https://github.com/appsignal/appsignal-nodejs/issues/535

### DIFF
--- a/packages/nodejs/.changesets/handle-request-errors-on-diagnose-http-transmitter-.md
+++ b/packages/nodejs/.changesets/handle-request-errors-on-diagnose-http-transmitter-.md
@@ -1,0 +1,7 @@
+---
+bump: "patch"
+type: "add"
+---
+
+Handle request errors in the internal HTTP transmitter used by diagnose tool
+and Push API key validator.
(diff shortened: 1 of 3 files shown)


unflxw (commented):
It might be best to return a differently-shaped object when there's an error (maybe just `{"error": error}`?) so that it can be handled separately where the transmitter is used. That way, the error message can be outputted differently, instead of treating it like a response.

For example, the output when failing to send the diagnose report would say "Response code: request_error" and "Response body: Error: getaddrinfo ENOTFOUND appsignal.com". I think it would be clearer if we only showed something like "Error: getaddrinfo ENOTFOUND appsignal.com".

As an aside to this, my opinion is that `Promise.reject` should be used only for unexpected failures (i.e. "internet is down"), and things like "non-200 status code" or "API key is invalid" should be given as information in a `Promise.resolve`. I believe this would simplify implementing the above.

unflxw (approved):

luismiramirez:
@unflxw Thank you! I gave the commit a different approach using your feedback.

tombruijn (approved):

jeffkreeftmeijer (approved):

PR #541, opened 2022-01-05
Add and implement sendSessionData config option

luismiramirez:
## Add sendSessionData config option

The `sendSessionData` config option is now available and its default
value is `true`. It'll be used to globally determine if the users want
to send session data in the spans or not.

## Implement sendSessionData config option

The `sendSessionData` config option allows the user to decide if they
want to send session data to AppSignal. Its default value is `true`.

`Span#setSampleData()` now checks if the `sendSessionData` option is
enabled when sending `session_data` to decide if they're going to be
sent to AppSignal or not.

Closes: #516

### DIFF
--- a/packages/nodejs/.changesets/skipsessiondata-config-opt-is-now-available.md
+++ b/packages/nodejs/.changesets/skipsessiondata-config-opt-is-now-available.md
@@ -1,0 +1,7 @@
+---
+bump: "patch"
+type: "add"
+---
+
+The `sendSessionData` config option is now available. When set to `false`, it prevents the
+integration from sending session data to AppSignal.
--- a/packages/nodejs/src/__tests__/config.test.ts
+++ b/packages/nodejs/src/__tests__/config.test.ts
@@ -40,6 +40,7 @@ describe("Configuration", () => {
     ],
     sendEnvironmentMetadata: true,
     sendParams: true,
+    sendSessionData: true,
     transactionDebugMode: false
   }
 
--- a/packages/nodejs/src/__tests__/span.test.ts
+++ b/packages/nodejs/src/__tests__/span.test.ts
@@ -192,7 +192,7 @@ describe(".setSampleData()", () => {
     jest.clearAllMocks()
   })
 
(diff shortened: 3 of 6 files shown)


jeffkreeftmeijer (approved):

tombruijn:
Should we implement it as `sendSessionData` instead, if we agree in https://github.com/appsignal/integration-guide/pull/114? That way we don't introduce it in one release and deprecate it in another.

packages/nodejs/src/version.ts
```diff
@@ -1,3 +1,3 @@
 // Do not touch this file, auto-generated by scripts/create-versionfile
 export const VERSION = "2.2.8"
-export const AGENT_VERSION = "0db01c2"
+export const AGENT_VERSION = "15ee07b"
```
tombruijn, line 3: Where does this change come from? It's not related to this PR I think.

luismiramirez:
>Should we implement it as sendSessionData instead, if we agree in appsignal/integration-guide#114? That way we don't introduce it in one release and deprecate it in another.

Yes, moving it to DRAFT status so it doesn't get merged.

packages/nodejs/.changesets/skipsessiondata-config-opt-is-now-available.md
```diff
@@ -0,0 +1,7 @@
+---
+bump: "patch"
+type: "add"
+---
+
+The `sendSessionData` config option is now available. When set to `false`, it prevents the
+integration from sending request params to AppSignal.
```
unflxw, line 7: Is the content of the session data "request params"? Maybe it's a bit redundant, but I'd say "session data" instead.

unflxw (approved):

tombruijn (approved):
Looks good. But please undo the change in the `version.ts` file.

luismiramirez:
> But please undo the change in the version.ts file.

I thought it was already removed, don't know why this keeps happening.

Fixing it...

jeffkreeftmeijer (approved):

Issue #540, opened 2022-01-04
Heap statistics should be displayed for every host


shairyar:
We track the metrics for heap statistics but don't separate those from different apps on different hosts. The last host that pushes metrics for a certain minute overwrites all the other collected metrics and this makes it not possible to know from which app the metrics in the graph are coming from.

This is the file where we collect the metrics, but no tags are being added: https://github.com/appsignal/appsignal-nodejs/blob/main/packages/nodejs/src/probes/v8/index.ts

## To do

- [x] Add a tag with the hostname to the linked metrics - https://github.com/appsignal/appsignal-nodejs/pull/543
- [x] Update the magic dashboard config - https://github.com/appsignal/public_config/pull/36
- [ ] Update the magic dashboard for existing dashboards.

shairyar:
https://app.intercom.com/a/apps/yzor8gyw/inbox/inbox/4356044/conversations/16410700082948

tombruijn:
@unflxw I think the magic dashboard was updated in https://github.com/appsignal/public_config/pull/36? I'll close the issue then.
I think @thijsc was working on (auto) updating the existing magic dashboards in a different issue.

PR #543, opened 2022-01-11
Add hostname tag to V8 probe gauges

unflxw:
Add a tag containing the hostname to the V8 probe gauges. Before this change, hosts would override each others' values. Now each of the hosts' values is given for its own hostname.

This PR completes the first task in #540.


tombruijn:
@unflxw the prettier thing failed in the CI. You can check that the Git hooks are fired on your local machine if you want to automate that.

tombruijn (approved):

jeffkreeftmeijer (approved):

luismiramirez:
@unflxw As commented before, let's grab the hostname from the client stored in the global object so we don't have to pass it around between function calls

luismiramirez (approved):

Issue #471, opened 2021-10-29
Merge extension package into nodejs package


luismiramirez:
## Background

The extension installation and install report from the `@appsignal/nodejs-ext` package are outside of the core package `@appsignal/nodejs`. The `nodejs-ext` extension package is an [optional dependency](https://github.com/appsignal/appsignal-nodejs/blob/ea685296957fdbd61dd13e168a5adbb64797d243/packages/nodejs/package.json#L18-L20) of the core package.

When the extension installation fails in the `nodejs-ext` package, npm will delete the failed package directory. Which is why we have to check if the module was loaded or not in the [`ExtensionWrapper`](https://github.com/appsignal/appsignal-nodejs/blob/ea685296957fdbd61dd13e168a5adbb64797d243/packages/nodejs/src/extension_wrapper.ts#L5-L18)

While working on issue #357 to merge the two listed files, I noticed that we can't rely on `nodejs-ext` package to be present to get the version from as it may not be loaded if something fails during installation. So, to know the current agent version at all times in the `nodejs` package, we copy it to another file in the `nodejs` package files using this [version file script](https://github.com/appsignal/appsignal-nodejs/blob/ea685296957fdbd61dd13e168a5adbb64797d243/packages/nodejs/scripts/create-versionfile.js).
 
## The problem

- Keeping the extension-related operations in a separated package from the core one doesn't make any sense as `@appsignal/nodejs` will not work without the extension, and no other package is directly dependent on the extension.
- An unsuccessful installation should not panic in any case making the dependency installation fail for the customer's application or preventing a package from being installed. Our other integrations don't work this way either.
- The diagnose tool has to fetch the extension install report from the host's global `/tmp` dir, which may not include the install report if an intermediate build machine was used and the app was copied over to the production host.

## The proposal

- [x] Move the extension-related operations into `@appsignal/nodejs` package
- [x] Prevent the extension installation from panicking when something goes wrong
- [x] Get version values from the extension constants
- [x] Remove `create-versionfile.js` script from `@appsignal/nodejs`
- [x] Do manual testing on extension installation failures, to check
    - That the `npm install` process __does not__ exit with a failure exit code.
    - That the `appsignal/nodejs` package __is not__ removed from the file system on failure.
- [ ] Do manual testing on Windows to also verify the above tests

luismiramirez:
Could you add your feedback about this, please?

@jeffkreeftmeijer, @unflxw, @thijsc

No feedback will be taken as an approve

[...]

PR #545, opened 2022-01-11
Merge extension package into core one

luismiramirez:
The extension package is no longer an independent entity, it is now
merged into @appsignal/nodejs.

The installation script for the extension no longer aborts `npm install`
if something goes wrong during the process. This allows the users to
keep running their apps.

There's no need to use the script that creates the version file as
the extension constants are always available now.
So the script has been removed.

Fixes: #471


packages/nodejs/.changesets/merge-extension-package-into-nodejs.md
```diff
@@ -0,0 +1,7 @@
+---
+bump: "minor"
+type: "change"
+---
+
+The extension package (@appsignal/nodejs-ext) is no longer a separated dependency. Now, it is
+merged into @appsignal/nodejs.
```
tombruijn, line 7:
```suggestion
The extension package (@appsignal/nodejs-ext) is no longer a separated dependency. The extension is part of the @appsignal/nodejs package now.
```

packages/nodejs/package.json
```diff
@@ -26,17 +25,26 @@
     "redis": "*"
   },
   "scripts": {
-    "prebuild": "npm run versionfile",
     "build": "tsc -p tsconfig.json",
-    "build:watch": "npm run versionfile && tsc -p tsconfig.json -w --preserveWatchOutput",
-    "clean": "rimraf dist coverage",
+    "build:ext": "node-gyp rebuild",
```
tombruijn, line 29:
```suggestion
```
I would just remove this. It's like only calling "make" for the Ruby extension installation. That doesn't work if you haven't done all the other steps beforehand already.

package.json
```diff
@@ -28,7 +28,7 @@
     "bootstrap:ci": "echo Please run 'mono bootstrap --ci' instead.",
     "build": "echo Please run 'mono build' instead.",
     "build:watch": "echo Please run 'mono run npm run build:watch --parallel' instead.",
-    "build:ext": "echo Please run 'mono run npm run build:ext --package @appsignal/nodejs-ext' instead.",
+    "build:ext": "echo Please run 'mono run npm run build:ext --package @appsignal/nodejs' instead.",
```
tombruijn, line 31:
```suggestion
```
See my other comment about this script.

packages/nodejs/package.json
```diff
@@ -26,17 +25,26 @@
...
+    "test:watch": "jest --filter=./test/filter.js --watch"
   },
+  "os": [
+    "linux",
+    "darwin",
+    "freebsd"
+  ],
+  "cpu": [
+    "x64",
+    "x86"
```
tombruijn, line 46: Do we need to include ARM here? We support that for Node.js too right?

packages/nodejs/scripts/extension/extension.js
```diff
@@ -6,13 +6,13 @@ const fs = require("fs")
 const crypto = require("crypto")
 const childProcess = require("child_process")
 
-const { AGENT_VERSION, MIRRORS, TRIPLES } = require("./extension/constants")
+const { AGENT_VERSION, MIRRORS, TRIPLES } = require("./support/constants")
```
tombruijn, line 9: Can you also send in a PR for the agent repo to update this path?

packages/nodejs/src/diagnose.ts
```diff
@@ -265,7 +265,7 @@ export class DiagnoseTool {
   }
 }
 
-// This implementation should match the `packages/nodejs-ext/scripts/report.js`
+// This implementation should match the `/scripts/extension/report.js`
```
tombruijn, line 268:
```suggestion
// This implementation should match the `scripts/extension/report.js`
```
The leading `/` implies root of the file system.

tombruijn (changes requested):

packages/nodejs/scripts/extension/extension.js
luismiramirez: PR: https://github.com/appsignal/appsignal-agent/pull/738

tombruijn (approved):

jeffkreeftmeijer (approved):

Issue #546, opened 2022-01-14
Some environment variables are ignored


unflxw:
Due to [the way that the `active` and `enableMinutelyProbes` configuration options are processed in the `BaseClient` constructor](https://github.com/appsignal/appsignal-nodejs/blob/434599b1496cc1570713465610f794b5118f1733/packages/nodejs/src/client.ts#L47-L49), their values will only be taken into account when passed through the constructor's options object. This means that the `APPSIGNAL_ACTIVE` and `APPSIGNAL_ENABLE_MINUTELY_PROBES` environment variables are, in practice, ignored.

🚨 Blocked on: https://github.com/appsignal/appsignal-nodejs/pull/547

PR #547, opened 2022-01-14
Fix `enableMinutelyProbes` behaviour

### LINKED ISSUE #544: Setting `enableMinutelyProbes` to false does not disable the minutely probes
The `enableMinutelyProbes` configuration setting should disable all minutely probes functionality when it is set to false. However, this is not actually the case. [As it is currently implemented](https://github.com/appsignal/appsignal-nodejs/blob/a3f8db67cdca6e68c1df8e54791671bce0c6cad6/packages/nodejs/src/bootstrap.ts#L48-L50), the only effect of setting `enableMinutelyProbes` to false is that the default Node.js minutely probe is not registered, but the minutely probes system will still function, and custom probes registered by the user will still run.

unflxw:
Fixes #544.

### Implement unregister method for probes

Before this change, it was possible to disable a probe by registering an empty probe with the same name, but this is neither
obvious nor documented. This commit adds a `probes.unregister()` method that removes a probe in a more explicit manner.

[This will help in improving the docs relating to the minutely probes, as outlined in appsignal/appsignal-docs#532.]

### Fix enableMinutelyProbes behaviour

Before this change, setting `enableMinutelyProbes` to `false` would not disable the minutely probes system, it merely does not register the default minutely probe.

The flag's behaviour has been changed so that the minutely probes system is disabled in its entirety when it is set to `false`. The `Probes` object within `Metrics` is replaced by a `NoopProbes` object, which does nothing when a probe is registered.

[This also fixes the issue with the `APPSIGNAL_ENABLE_MINUTELY_PROBES` environment variable being effectively ignored, as described in #546]

### Fix probes not being overwritten

In the documentation for other integrations, we describe probes as being overwritten if a different probe with the same name is registered. While this was not documented as such for the Node.js integration in our user-facing documentation, our internal documentation for the `register` method describes its behaviour as such.

However, since the `EventEmitter` listeners were not being removed, when a new probe function was registered under the same name as an existing probe function, both probe functions would be called every minute. This commit fixes this, so that the behaviour matches the other integrations and the internal documentation.

### DIFF
--- a/packages/nodejs/.changesets/fix-enableminutelyprobes-behaviour.md
+++ b/packages/nodejs/.changesets/fix-enableminutelyprobes-behaviour.md
@@ -1,0 +1,21 @@
+---
+bump: "patch"
+type: "fix"
+---
+
+Setting `enableMinutelyProbes` to `false` now disables the minutely probes
+system. Custom probes will not be called when the minutely probes are
+disabled. In addition, the `APPSIGNAL_ENABLE_MINUTELY_PROBES` environment
+variable can now be used to enable or disable the minutely probes.
+
+Before this change, setting `enableMinutelyProbes` to `false` would not
+register the default Node.js heap statistics minutely probe, but custom
+probes would still be called. To opt in for the previous behaviour,
+disabling only the Node.js heap statistics minutely probe without disabling
+custom probes, you can use the `probes.unregister()` method to unregister
+the default probe:
+
+```js
+const probes = appsignal.metrics().probes();
+probes.unregister("v8_stats");
+```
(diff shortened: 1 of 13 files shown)


tombruijn (approved):

unflxw:
Realised there was _another_ issue with this (probes with the same name not overwriting each other) while writing the PR message body about the `EventEmitter` in [the docs PR](https://github.com/appsignal/appsignal-docs/pull/534). I added the fix for it to this PR.

Ping @tombruijn in case you want to re-review.

luismiramirez (approved):
👍🏼  after making the liner happy

jeffkreeftmeijer (approved):

tombruijn (approved):

tombruijn:
I've created an issue for the Ruby gem to have a similar `unregister` method, now that the Elixir and Node.js integration both have it.

PR #548, opened 2022-01-17
Use APPSIGNAL_ACTIVE environment value

### LINKED ISSUE #546: Some environment variables are ignored

unflxw:
Before this change, the value of the `APPSIGNAL_ACTIVE` environment variable would never be read, as the `active` config option was read directly from the `BaseClient` constructor. After this change, the option is read from the `Config` object, after it has been initialised in the constructor.

Fixes #546.

### DIFF
--- a/packages/nodejs/.changesets/use-appsignal_active-environment-variable.md
+++ b/packages/nodejs/.changesets/use-appsignal_active-environment-variable.md
@@ -1,0 +1,6 @@
+---
+bump: "patch"
+type: "fix"
+---
+
+Use the APPSIGNAL_ACTIVE environment variable to determine whether AppSignal is active.
--- a/packages/nodejs/src/__tests__/config.test.ts
+++ b/packages/nodejs/src/__tests__/config.test.ts
@@ -13,6 +13,7 @@ describe("Configuration", () => {
   let initialEnv: { [key: string]: any }
 
   const expectedDefaultConfig = {
+    active: false,
     caFilePath: path.join(__dirname, "../../cert/cacert.pem"),
     debug: false,
     dnsServers: [],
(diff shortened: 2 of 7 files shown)


tombruijn (changes requested):
I see the same problem exists in the diagnose tool here: https://github.com/appsignal/appsignal-nodejs/blob/c77ff63e416c6cae41e699cc245559614035a9fb/packages/nodejs/src/diagnose.ts#L40-L42

It defaults to "active: true" on line 40, and only uses the diagnose's "true" value as the active option's value for the extension loading.

Can you update that part to match the new fixed behavior?

unflxw:
@tombruijn I have addressed it for the configuration object. The diagnose will now always show `active` to be `false` unless `APPSIGNAL_ACTIVE` is set to true. I have not changed it for the extension, as it is necessary for the extension to start in order to obtain diagnostic information from it -- the extension is always given `{active: true}` as an argument.

tombruijn:
> I have not changed it for the extension, as it is necessary for the extension to start in order to obtain diagnostic information from it -- the extension is always given {active: true} as an argument.

Yes those two things should not be related to one another and should be separated. I've created an issue to fix that: https://github.com/appsignal/appsignal-nodejs/issues/551

tombruijn (approved):

luismiramirez (approved):

jeffkreeftmeijer (approved):

PR #553, opened 2022-01-18
Update install report path

### LINKED ISSUE #550: Store install report in extension directory
The Node.js package now writes the install report in the system tmp directory. This was necessary because the `nodejs-ext` package would not persist on the file system in case of an installation failure. In PR #545 we've merged the `nodejs-ext` package in the main `nodejs` package, and it will no longer fail on install. Even if the installation fails the install files persist on the file system.

We need to change the install report location from `/tmp/appsignal-<generated key>-install.report` to the location in the `ext` dir in the `nodejs` package. The new report file can be named `install.report` like in the other integrations.

## To do

- [x] Move the install report to the `nodejs/ext` directory.
- [x] Rename the install report file to `install.report`.

Be aware that the paths need to be updated for the installer script that writes the file and the diagnose tool that reads it.

tombruijn:
## Update report path to match new package structure

In PR #545 we've merged the nodejs-ext package into the nodejs package.
This change moved some files around, and moved the extension install
scripts one level deeper into the `extension` subdirectory. Update the
report path name generation to match the new depth of the file.

Previously: `packages/nodejs-ext/scripts/extension.js`
New: `packages/nodejs/scripts/extension/extension.js`

It's interesting that no tests failed because of this. I think it's not
caught in the diagnose_tests, because we stub the install report.

I've now added a simple test to test the presence of the report, and not
a parsing error that occurs when no file could be read or there was a
problem reading the file. I'm struggling with Jest's mock system to
write a test for this and I'd rather look at it when addressing issue
#550 that updates the path location, which makes it easier to stub I
hope.

## Update install report location to ext directory

In PR #545 we've merged the nodejs-ext package into the nodejs package.
It no longer fails the package install on extension installation
failure, which means the ext directory remains after a installation
failure.

We can now store the install report file in the package's `ext`
directory, like we do in all the other integrations. This means we no
longer need to generate a filename for the install report to store it in
the system tmp directory.

Closes #550

### DIFF
--- a/packages/nodejs/.changesets/update-install-report-path.md
+++ b/packages/nodejs/.changesets/update-install-report-path.md
@@ -1,0 +1,6 @@
+---
+bump: "patch"
+type: "change"
+---
+
+Update install report path location. The "nodejs" package now stores the install report in its own package path, with the rest of the extension files. This makes it easier to access by the diagnose report.
--- a/packages/nodejs/.changesets/fix-install-report-path-name-generation.md
+++ b/packages/nodejs/.changesets/fix-install-report-path-name-generation.md
@@ -1,0 +1,6 @@
+---
+bump: "patch"
+type: "fix"
+---
+
+Fix install report path name generation. This fixes the diagnose report not being able to find the path where the install report is stored.
--- a/packages/nodejs/.changesets/fix-install-report-path-name-generation.md
+++ b/packages/nodejs/.changesets/fix-install-report-path-name-generation.md
@@ -1,6 +1,0 @@
----
-bump: "patch"
-type: "fix"
----
-
-Fix install report path name generation. This fixes the diagnose report not being able to find the path where the install report is stored.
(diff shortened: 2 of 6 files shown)


jeffkreeftmeijer (approved):

packages/nodejs/.changesets/update-install-report-path.md
```diff
@@ -0,0 +1,6 @@
+---
+bump: "patch"
+type: "change"
+---
+
+Update install report path location. The "nodejs" package now stores the install report in its own package path, with the rest of the extension files. This makes it easier accessible by the diagnose report.
```
unflxw, line 6:
```suggestion
Update install report path location. The "nodejs" package now stores the install report in its own package path, with the rest of the extension files. This makes it easier to access by the diagnose report.
```

(or "more easily accessible")

unflxw (approved):
Very happy to see that this gets rid of all the path hashing stuff. 🎉

PR #554, opened 2022-01-18
Improve extension loading state and starting

### LINKED ISSUE #551: Extension should be accessible even if active is set to false
The Node.js extension is currently unaccessible when the config is `active: false`. It currently checks if the `active` config option is `true` or not, and will only call `start` when it is `true`.

The extension is always loaded, but it sets the `isLoaded` attribute only when `start` is called. Which it should not do. The extension functions can be called, even if it isn't "started".

## To do

- [ ] Call `Extension.start` when `active` is `true`, but from the AppSignal client. Not in the Extension class itself.
- [ ] Remove the `isLoaded` assignment from the `Extension.start` function.
- [ ] Remove the `this.isLoaded` check from the `Extension` class's `stop` and `diagnose` functions.
- [ ] Set the `Extension`'s `isLoaded` attribute when the [C-extension is loaded in Node.js](https://github.com/appsignal/appsignal-nodejs/blob/32179126405de44b186a1d18b2c8a99e16756be5/packages/nodejs/src/extension_wrapper.ts#L6).

tombruijn:
## The problem

The `Extension`'s '`isLoaded` attribute was only set when the
extension's `start` function was called. This made it difficult to
determine if the extension was really loaded in the app or not, without
starting the extension first.

This made it necessary to start the extension in scenarios where we need
to call extension functions without needing to starting it first, such
as in the diagnose command.

Calling `Extension`'s `start` doesn't change any state in the extension
that needs to be set before any other function can be called. I think it
was written this way as a way to communicate loading issues. A better
name for this attribute would have been `isRunning`, but this is not
something we have to track.

## Solution

Update how the `isLoaded` is set, by checking if the extension was
loaded into the app or not on load, not just when `start` is called.

I've chosen to communicate this with an export from the extension
wrapper, rather than if the start function returns an error or not.
Since the value doesn't change at runtime it's now a static attribute of
the class.

I've removed the `isLoaded` return values from the extension's `start`
and `stop` function, these are unused.

## About the error thrown by `start`

I've left the `Extension`'s `start` function intact to print an error
when called and the extension isn't loaded. In other integrations we
print this on load of the extension. I suspect this was done on start,
so that it wouldn't always be printed on unsupported Operating Systems.
We may want to revisit this, or find another way to not print this all
the time for those systems.

## Only start extension when its loaded and active

Do not automatically start the extension when it's initialized, but do
this from the client like in all the other integrations. This also makes
it easier to initialize the `Extension` class for other uses, like the
diagnose without having to worry about the active state or not.

Closes #551

### DIFF
--- a/packages/nodejs/.changesets/change-the-way-the-extension-is-started.md
+++ b/packages/nodejs/.changesets/change-the-way-the-extension-is-started.md
@@ -1,0 +1,6 @@
+---
+bump: "patch"
+type: "change"
+---
+
+Change the way the extension is started. Previously it was started on initialization of the extension class (if the config was active), but now it's started by the client class.
--- a/packages/nodejs/.changesets/fix-extension-fallback-functions.md
+++ b/packages/nodejs/.changesets/fix-extension-fallback-functions.md
@@ -1,0 +1,6 @@
+---
+bump: "patch"
+type: "fix"
+---
+
+Fix the extension function fallbacks on installation failure. When the extension fails to install and calls are made to the not loaded functions, it will no longer throw an error.
--- a/packages/nodejs/.changesets/improve-extension-loading.md
+++ b/packages/nodejs/.changesets/improve-extension-loading.md
@@ -1,0 +1,6 @@
+---
+bump: "patch"
+type: "change"
+---
+
+Improve the reporting of extension loading state in the diagnose report. Previously it would only report it as loaded when it was also running.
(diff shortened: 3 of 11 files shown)


luismiramirez (approved):

tombruijn:
Rebased on PR #558 because the failure scenario tests were not run.

luismiramirez (approved):

jeffkreeftmeijer (approved):

PR #558, opened 2022-01-19
Fix the CI build to include intentional failures

tombruijn:
In the package mergers of PR #545, we merged the `nodejs-ext` package in
the `nodejs` package.

The tests that belong to the `nodejs-ext` package that tested the
behavior when the extension failed to install were not run after merging
that PR.

I've updated the build matrix to run those tests for the `nodejs`
package now.

For Jest I also had to configure a new root directory. In the
`nodejs-ext` package this was not configure, so I assume it previously
searched in all subdirectories.

For the clean command I also had to add the `build` directory be cleaned
up, otherwise it would continue to load the loaded extension without
problem.

[skip changeset]

### DIFF
--- a/packages/nodejs/package.json
+++ b/packages/nodejs/package.json
@@ -27,13 +27,14 @@
   "scripts": {
     "build": "tsc -p tsconfig.json",
     "build:watch": "tsc -p tsconfig.json -w --preserveWatchOutput",
-    "clean": "rimraf dist coverage && rimraf ext/appsignal-agent ext/libappsignal.a ext/appsignal.* ext/*.tar.gz",
+    "clean": "rimraf dist coverage build && rimraf ext/appsignal-agent ext/libappsignal.a ext/appsignal.* ext/*.tar.gz ext/*.report",
     "install": "node scripts/extension/extension.js",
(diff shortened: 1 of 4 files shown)


packages/nodejs/package.json
```diff
@@ -27,13 +27,14 @@
   "scripts": {
     "build": "tsc -p tsconfig.json",
     "build:watch": "tsc -p tsconfig.json -w --preserveWatchOutput",
-    "clean": "rimraf dist coverage && rimraf ext/appsignal-agent ext/libappsignal.a ext/appsignal.* ext/*.tar.gz",
+    "clean": "rimraf dist coverage build && rimraf ext/appsignal-agent ext/libappsignal.a ext/appsignal.* ext/*.tar.gz",
```
unflxw, line 30: I'm wondering if `ext/install.report` should be added to this as well.

unflxw (approved):

packages/nodejs/package.json
tombruijn: yep!

luismiramirez (approved):

PR #566, opened 2022-01-25
Lock Redis supported version to 3.x.x

luismiramirez:
The supported Redis version is now locked to 3.x.x. This will prevent app startup failures for users
with Redis 4.x.x.

Support issue: https://github.com/appsignal/support/issues/184

### DIFF
--- a/packages/nodejs/src/instrumentation/redis/index.ts
+++ b/packages/nodejs/src/instrumentation/redis/index.ts
@@ -82,7 +82,7 @@ export const instrument = (
   mod: RedisModule,
   tracer: Tracer
 ): Plugin<RedisModule> => ({
-  version: ">= 3.0.0",
+  version: "~3",
   install(): RedisModule {
     shimmer.wrap(
       mod.RedisClient.prototype,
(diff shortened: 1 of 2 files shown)


packages/nodejs/src/instrumentation/redis/index.ts
```diff
@@ -82,7 +82,7 @@ export const instrument = (
   mod: RedisModule,
   tracer: Tracer
 ): Plugin<RedisModule> => ({
-  version: ">= 3.0.0",
+  version: ">=3.0.0 <4.0.0",
```
tombruijn, line 85: I'm looking at the other operators we can use to do this version restriction. Would the tilde operator `~3` also work? Or `^3.x`?

https://github.com/npm/node-semver#tilde-ranges-123-12-1

luismiramirez: Yes, `~3` should work.

I use this for testing: https://semver.npmjs.com/

unflxw (approved):

tombruijn (approved):

PR #587, opened 2022-02-01
Update Span type used in ScopeManager

tombruijn:
We changed the types used in PR #451. We no longer rely on the
`@appsignal/types` package for this, but define it in the package
itself.

The ScopeManager appears to not have been updated to match and is the
only place in which the `NodeSpan` from the types package is still
used. If we update this usage we can remove all Node.js types from the
types package.

### DIFF
--- a/packages/nodejs/src/scope.ts
+++ b/packages/nodejs/src/scope.ts
@@ -8,7 +8,8 @@
  * Copyright 2018, Google LLC
  */
 
-import { NodeSpan, Func } from "@appsignal/types"
+import { Func } from "@appsignal/types"
+import { Span } from "./interfaces/span"
 import * as asyncHooks from "async_hooks"
 import { EventEmitter } from "events"
 import shimmer from "shimmer"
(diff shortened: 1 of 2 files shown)


luismiramirez (approved):

jeffkreeftmeijer (approved):

packages/nodejs/src/scope.ts
```diff
@@ -8,7 +8,8 @@
  * Copyright 2018, Google LLC
  */
 
-import { NodeSpan, Func } from "@appsignal/types"
+import { Func } from "@appsignal/types"
```
unflxw, line 11: > If we update this usage we can remove all Node.js types from the types package.

I see here that we still import `Func` from it. What's the plan for the types package? Would we want to get rid of this usage as well in a future PR?

unflxw (approved):

packages/nodejs/src/scope.ts
luismiramirez: This `Func` type is shared between the JS and the Node.js integrations. When we removed the usage of Node interfaces from the types package, we didn't remove the shared ones.

I think they're good as they are right now to avoid duplication between the two monorepos.

tombruijn: Agreed. I think it's fine to keep these for now.

Issue #532, opened 2021-12-13
Use transmitter class in extension installer


tombruijn:
The node.js integration's extension installer now uses the https library directly. We've made an abstraction for this in PR #531, we should use that instead. This way it also supports the cacert file and other config options.

~🚨 This is blocked on merging the extension package into the main package: https://github.com/appsignal/appsignal-nodejs/issues/471~
~Because the extension package can't load the transmitter from the main package without creating a circular dependency chain. And we don't want to duplicate it.~

## To do

- [ ] Use the transmitter class in the `extension.js` script.
    - https://github.com/appsignal/appsignal-nodejs/blob/434599b1496cc1570713465610f794b5118f1733/packages/nodejs/scripts/extension/extension.js#L51-L65

tombruijn:
Moving directly to "to do", as we've already discussed this with the team in the call.

unflxw:
PR #600 also related to this task.

PR #589, opened 2022-02-02
Test the Transmitter class

unflxw:
Adds tests for the existing functionality of the `Transmitter` class. It checks that the query parameters are added to the URL, using values from the config, that the promise is resolved with a status code and body, or rejected with an error, that the body is parsed as JSON or replaced with an empty object, and that it uses the CA file defined in the config in HTTPS requests.

Fixes an issue with the Transmitter where the full URL was used in place of the path -- that is, `GET http://foo.com/bar` was sent over the wire, instead of `GET /bar`.

Removes the unused `active` parameter from the Transmitter class, and no longer passes it as `true` to the Config. Using the Transmitter should not initialise the extension.

Changes the Transmitter implementation so that, when a CA file could not be read, the `ca` property of the HTTPS options object is unset, instead of being set to an empty string.

This is a first step towards #532. 

[skip changeset]

### DIFF
--- a/packages/nodejs/src/__tests__/transmitter.test.ts
+++ b/packages/nodejs/src/__tests__/transmitter.test.ts
@@ -1,0 +1,170 @@
+import { Transmitter } from "../transmitter"
+import nock from "nock"
+import https from "https"
+import fs from "fs"
+import { ClientRequest, IncomingMessage } from "http"
+
+describe("Transmitter", () => {
+  let originalEnv: any = undefined
+
+  beforeEach(() => {
+    if (!nock.isActive()) {
+      nock.activate()
+    }
+
+    nock.disableNetConnect()
+
+    originalEnv = { ...process.env }
+  })
+
+  afterEach(() => {
+    jest.clearAllMocks()
+
+    process.env = originalEnv
+  })
+
+  afterAll(() => {
+    nock.restore()
+  })
+
+  describe(".transmit", () => {
+    function sampleTransmitter(): Transmitter {
+      return new Transmitter("https://example.com/foo", "request body")
+    }
+
+    function mockSampleRequest(responseBody: object | string, query = {}) {
+      nock("https://example.com")
+        .post("/foo", "request body")
+        .query({
+          api_key: "",
+          name: "",
+          environment: "test",
+          hostname: "",
+          ...query
+        })
+        .reply(200, responseBody)
+    }
+
+    async function expectResponse(
+      response: Promise<any>,
+      expectedBody: object
+    ) {
+      await expect(response).resolves.toEqual({
+        status: 200,
+        body: expectedBody
+      })
+    }
+
+    function mockHTTPSRequest() {
+      return jest.spyOn(https, "request").mockImplementation(() => {
+        const clientRequestMock = {
+          on: () => {},
+          write: () => {},
+          end: () => {}
+        }
+
+        return (clientRequestMock as unknown) as ClientRequest
+      })
+    }
+
+    it("resolves to a response on success", async () => {
+      const transmitter = sampleTransmitter()
+
+      mockSampleRequest({ json: "response" })
+
+      await expectResponse(transmitter.transmit(), { json: "response" })
+    })
+
+    it("resolves to an empty object if the response is not JSON", async () => {
+      const transmitter = sampleTransmitter()
+
+      mockSampleRequest("not JSON")
+
+      await expectResponse(transmitter.transmit(), {})
+    })
+
+    it("rejects on error", async () => {
+      const transmitter = sampleTransmitter()
+
+      // do not mock the request -- beforeEach uses nock to disable network
+      // requests, so any request to a non-mocked host will throw an error
+
+      await expect(transmitter.transmit()).rejects.toMatchObject({
+        error: expect.any(Error)
+      })
+    })
+
+    it("uses config values in the query string", async () => {
+      process.env.APPSIGNAL_PUSH_API_KEY = "some_api_key"
+      process.env.APPSIGNAL_APP_NAME = "myApp"
+      ;(process.env as any).NODE_ENV = "development"
+      process.env.APPSIGNAL_HOSTNAME = "myHostname"
+
+      const transmitter = sampleTransmitter()
+
+      mockSampleRequest(
+        {},
+        {
+          api_key: "some_api_key",
+          name: "myApp",
+          environment: "development",
+          hostname: "myHostname"
+        }
+      )
+
+      await expectResponse(transmitter.transmit(), {})
+    })
+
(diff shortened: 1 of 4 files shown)


packages/nodejs/src/__tests__/transmitter.test.ts
```diff
@@ -0,0 +1,168 @@
...
+      })
+
+      const consoleWarnSpy = jest.spyOn(console, "warn")
+
+      const httpsRequestMock = mockHTTPSRequest()
+
+      sampleTransmitter().transmit()
+
+      expect(httpsRequestMock).toHaveBeenCalledWith(
+        expect.objectContaining({ ca: "" }),
```
tombruijn, line 159: Have we actually tested it works with an empty string? It maybe better if we not set the `ca` property when no file could be read?

packages/nodejs/src/__tests__/transmitter.test.ts
```diff
@@ -0,0 +1,168 @@
...
+
+      expect(fsReadFileSyncMock).toHaveBeenCalledWith("/foo/bar", "utf-8")
+
+      expect(httpsRequestMock).toHaveBeenCalledWith(
+        expect.objectContaining({ ca: "ca file contents" }),
+        expect.anything()
+      )
+    })
+
+    it("uses the system CAs if the CA file is not readable", () => {
```
tombruijn, line 140:
```suggestion
    it("doesn't configure the CA if the CA file is not readable", () => {
```

At first I read this like we would set the CA with the system CAs if the file is not readable. The description is what ultimately happens, but that's out of scope of our package.

packages/nodejs/src/__tests__/transmitter.test.ts
```diff
@@ -0,0 +1,168 @@
...
+
+      await expect(transmitter.transmit()).rejects.toMatchObject({
+        error: expect.any(Error)
+      })
+    })
+
+    it("uses config values in the query string", async () => {
+      process.env.APPSIGNAL_PUSH_API_KEY = "some_api_key"
+      process.env.APPSIGNAL_APP_NAME = "myApp"
+      ;(process.env as any).NODE_ENV = "development"
```
tombruijn, line 99: Is the semicolon needed here?

packages/nodejs/src/__tests__/transmitter.test.ts
```diff
@@ -0,0 +1,168 @@
...
+
+      await expectSampleResponse(transmitter.transmit())
+    })
+
+    it("resolves to an empty object if the response is not JSON", async () => {
+      const transmitter = sampleTransmitter()
+
+      mockSampleRequest({ responseBody: "not JSON" })
+
+      await expectSampleResponse(transmitter.transmit(), {})
```
tombruijn, line 82: Can you add a similar assertion to the test above this for a valid JSON body?

packages/nodejs/src/__tests__/transmitter.test.ts
unflxw: Yup -- without it, that line and the previous one are interpreted as one, that is:
```typescript
... = "myApp"(process.env as any).NODE_ENV = ...
```

I had put semicolons at the end of all these lines, for consistency, but prettier has interesting ideas of what pretty means, so it removed the unnecessary ones, and moved the necessary one there instead. 🤷

packages/nodejs/src/transmitter.ts
```diff
@@ -63,14 +62,12 @@ export class Transmitter {
...
   private requestOptions(requestUrl: URL): HashMap<any> {
-    const configData = this.#config.data
-
     return {
       method: "POST",
       protocol: requestUrl.protocol,
       host: requestUrl.hostname,
       port: requestUrl.port,
-      path: requestUrl.toString(),
+      path: requestUrl.pathname + requestUrl.search,
```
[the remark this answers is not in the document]
luismiramirez: Yes, that's the format needed to add proxy support. But I think this change is good as this is the correct usage of the `path` option when you're not using a proxy. Both ways work anyway.

I also think it'll be a good idea to change the HTTP client to Axios once we add proxy support to avoid future complications.

packages/nodejs/src/__tests__/transmitter.test.ts
unflxw: It is implicitly done -- If you don't pass a second argument to `expectSampleResponse()` with a custom response body to expect, it expects the sample response, which is `{json: "response"}`.

packages/nodejs/src/transmitter.ts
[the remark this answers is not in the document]
unflxw: It will need to do that when running it through a proxy, as that's what HTTP proxies expect -- but it's not supposed to do that when _not_ running through a proxy. It seems like many HTTP servers, in practice, deal with it fine: both our Sinatra mock server in the diagnose tests, and the actual server handling diagnose reports, are okay with it. It's still not the correct behaviour, I don't think, and `nock` wasn't having it.

packages/nodejs/src/__tests__/transmitter.test.ts
unflxw: Last time @luismiramirez and I were looking into the transmitter, we did check that an empty string value was interpreted the same way as if the property was not set.

I do agree not setting it at all would be preferable, and clearer -- I'll see if the implementation can be easily changed to do that.

packages/nodejs/src/__tests__/transmitter.test.ts
luismiramirez: I think removing the `responseBody` default from `mockSampleRequest()` will improve the readability of the tests.

tombruijn: I think so too. Now it looks like we don't assert the body on line 74.

unflxw: Agreed, went a bit too far with moving everything to default arguments. The purpose of it was to avoid having to talk about the response body at all in the "query values from config" test.

I have removed the default `{json: "response"}` object from both `mockSampleRequest` and `expectSampleResponse`, which is now just `expectResponse`, as there's nothing "sample" about it anymore. The signature of `mockSampleRequest` has been simplified as only one of its parameters is now optional.

Let me know if this is clearer, and if there's anything else that could improve it.

packages/nodejs/src/__tests__/transmitter.test.ts
unflxw: Done! Now the property is not set.

tombruijn: Awesome! 🚀

tombruijn (approved):

luismiramirez (approved):

PR #592, opened 2022-02-03
Add Span.toObject function, replacing Span.toJSON

tombruijn:
The `Span.toJSON` function required us to always parse it manually. Why
not do that in the function itself? That's what `Span.toObject` does.
Other name suggestions are welcome!

This way we don't cheat in tests and try and match on the JSON string
with `toMatch(/substring/)`. Now we need to specify which attribute
value we want to match.

All fields of the SpanData object are optional because if no span data
is returned by the extension, all fields are empty.

Based on https://github.com/appsignal/appsignal-nodejs/pull/591

### DIFF
--- a/packages/nodejs/src/interfaces/span.ts
+++ b/packages/nodejs/src/interfaces/span.ts
@@ -116,7 +116,24 @@ export interface Span {
   close(endTime?: number): this
 
   /**
-   * Returns a JSON string representing the internal Span in the agent.
+   * Returns a SpanData object representing the internal Span in the extension.
+   *
+   * @private
    */
-  toJSON(): string
+  toObject(): SpanData
+}
+
+/**
+ * The internal data structure of a `Span` inside the AppSignal Extension.
+ */
+export type SpanData = {
+  closed?: boolean
+  name?: string
+  namespace?: string
+  parent_span_id?: string
+  span_id?: string
+  start_time?: number
+  trace_id?: string
+  error?: { name: string; message: string; backtrace: Array<String> }
+  attributes?: { [key: string]: string }
 }
--- a/packages/nodejs/.changesets/remove-span-tojson-function.md
+++ b/packages/nodejs/.changesets/remove-span-tojson-function.md
@@ -1,0 +1,6 @@
+---
+bump: "patch"
+type: "remove"
+---
+
+Remove the private function `Span.toJSON`. This wasn't previously marked as private, but it was.
--- a/packages/nodejs/src/__tests__/scope.test.ts
+++ b/packages/nodejs/src/__tests__/scope.test.ts
@@ -98,12 +98,12 @@ describe("ScopeManager", () => {
         const span = scopeManager.active()
 
         expect(span).toBeDefined()
-        expect(span?.toJSON()).toMatch(/modified/)
+        expect(span?.toObject().name).toEqual("modified")
       }
 
       scopeManager.withContext(test, span => {
         span.setName("default")
-        expect(span.toJSON()).toMatch(/default/)
+        expect(span.toObject().name).toEqual("default")
       })
 
       scopeManager.withContext(test, span => {
@@ -115,5 +115,6 @@ describe("ScopeManager", () => {
     })
   })
 
+  // TODO: Add tests
   describe(".emitWithContext()", () => {})
 })
--- a/packages/nodejs/src/__tests__/span.test.ts
+++ b/packages/nodejs/src/__tests__/span.test.ts
@@ -2,28 +2,13 @@ import { ChildSpan, RootSpan } from "../span"
 import { span } from "../extension_wrapper"
 import { BaseClient } from "../client"
 import { Data } from "../internal/data"
-
-type SpanData = {
-  closed: boolean
-  name?: string
-  namespace?: string
-  parent_span_id?: string
-  span_id?: string
-  start_time?: number
-  trace_id?: string
-  attributes?: { [key: string]: string }
-}
+import { SpanData } from "../interfaces/span"
 
 describe("RootSpan", () => {
-  let span: RootSpan
-  let internal: SpanData
-
-  beforeEach(() => {
-    span = new RootSpan()
-    internal = JSON.parse(span.toJSON())
-  })
-
   it("creates a RootSpan", () => {
+    const span = new RootSpan()
+    const internal = span.toObject()
+
     expect(span).toBeInstanceOf(RootSpan)
     expect(internal.closed).toBeFalsy()
   })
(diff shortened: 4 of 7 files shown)


luismiramirez (approved):

packages/nodejs/src/interfaces/span.ts
```diff
@@ -116,7 +116,24 @@ export interface Span {
...
    */
-  toJSON(): string
+  toObject(): SpanData
+}
+
+/**
+ * The internal data structure of a `Span` inside the AppSignal Extension.
+ */
+export type SpanData = {
+  closed?: boolean
```
unflxw, line 130: The `closed` attribute should not be optional, I think, as it's present even when the span is closed.

Side note: one could also do something like

```typescript
type SpanData = ClosedSpanData | OpenSpanData
type ClosedSpanData = { closed: false }
type OpenSpanData = {
  closed: true,
  name: string,
  ...
}
```

This way, once the tests check that `closed != false` (for example, by throwing an error if it's closed) then the compiler will infer it's an OpenSpanData from that point onwards, and the other attributes can be accessed without the `.!` syntax.

I've been [playing with this idea in the TypeScript playground][playground] for a bit. None of this is necessary, I just think [discriminated unions](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#discriminated-unions) are neat.

[playground]: https://www.typescriptlang.org/play?#code/C4TwDgpgBAKgFgSwHYHMoF4oHlJPstAHygGEAbAewGcIATfVAKEdElMpvsVQygG9GUIVADGHOgC4owAE4BXCIwC+zVtBwQ83NJgHDR42lIBmAQzI0ANIOFJTAWwhSqsgsubG5SEcAQUkUCgQwAwoABQIUkhy9gBGEDIAlFKh-DZCCMZQERiYACwATIlp+voywXIyAXxi1JJQZhYQllB2jlIARAAiFHIoZKZUHSr6SlAQTSWlQuXAldW1nFKyCiPCKiqMIv4u0tq8QSHaYYWJzNtIVBRkEAB0lOHA2rdtEMUA9O97CFRQP1CmaTgaAJGQUGSMTLZACETwIt0WdGKemEFyuN3uFEez1eHy+cN+-2MyGg8REpjkNCgAHdoCI4BARABrOjuLY7a53B5hOGoF4ON5QT7fQm-QFqcYyMEyAEoUzISFZHnPRG0ZHpPZg6mtCDagCiUvBYQ68GgvLQVDgvTItFaFGAUHiBjqtA6ZxUaM5mOx8NxQq+sTkDoJf1+SAo2uJSFJjIpVIoWSeZrgWuYwtpolMAWpMlMYD+DuQAIaXh8fiQ0MYnm8vn84wAHpAfBotARlQQUtpkthcKkURkleaEYZ1dMhE8tTr9YaZMbTd8eJbrbbww6naq3Rq1jMKlUFyg2WiHRRe-tMBBG4zgC3QmFDrfTmdPRjuSfNKF+Y48SLQw0SY7Y0pEFL2bU8CHGS5KggQlgEYIA

unflxw (approved):
> This way we don't cheat in tests and try and match on the JSON string
with toMatch(/substring/). Now we need to specify which attribute
value we want to match.

👍💯

packages/nodejs/src/interfaces/span.ts
tombruijn: > The closed attribute should not be optional, I think, as it's present even when the span has been garbage-collected (or at least that's what toObject says)

It's not set if no data is returned from the extension at all. Would fit that into the union?

unflxw: I suppose you could, yes! In the playground I linked above, you can define `ClosedThing = { closed?: true }`, so that it's optional, and then it can also be an empty object.

unflxw: Huh, adding the empty object as an option makes this surprisingly tricky. I've tried fitting it into the union in several ways, by adding an additional `EmptyThing = {}`, or by making the `closed` property optional in the `ClosedThing` object, or some mix of the two. Thing is, it seems like if it's optional, it can't be a discriminant the compiler will still consider it possible, even if you check for `undefined`. And if there's an `EmptyThing` declaration in the union, then you can't easily check for the property -- it's a type error to try to access a property that doesn't exist in the object.

The only way I've been able to make it work is using [this magically typed version of Object.hasOwnProperty that I found on the internet](https://fettblog.eu/typescript-hasownproperty/). [Here's a playground showcasing it.][playground]

[playground]: https://www.typescriptlang.org/play?#code/C4TwDgpgBAKgFgSwHYHMoF4oHlJPstAHygGEAbAewGcIATfVKYgUQFsxQGUAob0SKGw4guGKAG8Avr37Ry1OqMzjuUNVADGlGrQBcUYACcArhG7S+4aDgh5EjZavVaFeqADMAhmRoAaJ2pInqwQ+lRGBOa87sZIGsAIFEhQKBDAXAAUsvoARBS4OUxQOS46hcQ5EOygOQCU+qIq6lBUAO4IwBpwUFlWtRIBzZqeNMX5tjm6g82GacaGyeKldPpePhC+UEEhuQAiFMYoZCM50kPOI9CV1SCT0+qzwPOLZ+fDoyXadHdvUI-PEmWbiMpleamkFhicQSSSgcBGWFaSAACoZxoZQAAeAAaUAgAA9gLZaFQJJJNgBNPGE4mk1Ho0AAaQgIAAfE4MhQAEYAK302M2YDRYH0FPqUG5PKgCFJuIAZFAAEoQDQUQy0TEUzaxADWSAoSNZAzU-wWEt5ADp4VRESjhRAMSAMkL8rUotxVUhwgZ7GhMKl0r6MnkCm6PUkqBQyBALZQUFlfRagf0APQpn0y6WkzwGKx4wxowzcBDuHrW230yCOhMETYAciBddq-SazgjUZjcZrqCTX1oqfTyDhDo20uAddJ+uAUBzsnzhYMFBnGg0ECopIABr3XBvBp7I9HYxR48BE9sIAOM6TM+EEGQyDPcwIHYXBiWegBCU8Ebc6Fv3NR9w7I8TzPYILygNMoC5YxpyHOARygEADmGZJPBXNdNwtc8NxnUlWgge9BghXgoII1CoFYTwdWgU9oC6FUdWlJBgCXHMoXiRIkG4DiYWSGUbCQbsUAaX1xW-RhM0ExpBlNZJyyRSsHVAYT60bfo5QVL9EyBd0oKAw8uwklBsPAy9v2vbMn2gF81RnFBPGQYtSwyD8BNwYTm2NdRTzRVotggfzmALNVg3gWjfRaOADjIWgtgoacuXovs6iiAzO2PYTTJCS8YOnCys3i-z3GQaAko0TxjFGChSzon0-NI9NPCQOLyIq5JWkMTwwDHZjH14rifU8acShGYAqEKOrJRVac7N8g1s2SWyiwG2ECUgeJpKDYzRIIcUtoIby1HfVz3NsTz-1+eb-KQQLBBCwwwoQjNGCoaLjFi+LEuS1xUuaME-jmM1jPdfdZtwJRqQ24ADtQDIA0yEMJmbcMvWArtxjsH9z3MxBLI8UroJVSrRnWmbYbQWwqHmNcx24IA

tombruijn: Wow cool! But that seems like a lot for something we only use in the test suite, right?

unflxw: Ah yes, definitely! 😅 Sorry, got carried away. Type systems are fun, in their own way. Disregard the above.

PR #594, opened 2022-02-03
Update confusing install messages for local build

tombruijn:
When the local agent and extension files were used for installation it
would also print the message that it downloaded the agent again. This is
confusing, because it didn't download anything.

Update the message to be specific for the local and remote builds.

[skip changeset]


luismiramirez (approved):
🙏🏼

unflxw (approved):

PR #603, opened 2022-02-07
Use transmitter in the extension download script

### LINKED ISSUE #532: Use transmitter class in extension installer

unflxw:
Closes #532.

[skip changeset]

### [Use transmitter to download the agent](https://github.com/appsignal/appsignal-nodejs/commit/7a7af759ad042a1376d5dd234dd13d7b3b6fdd47)

Refactor the `Transmitter` class and create a new `.download`
method in the transmitter, to be used in the download request in
the extension installation script.

Create a new `.request` method, which is used by both `.download`
and `transmit`, which takes care of setting up a well-formed
request and using the CA certificate specified in the environment,
leaving the handling of the response stream to the caller.

Move the existing logic in the extension installation script that
relates to the initialisation of the HTTPS request to the
Transmitter.

### [Build on install if working locally](https://github.com/appsignal/appsignal-nodejs/commit/42e831286731018d4215ab18cdc6a9fb784c44dd)

When working with the integration locally, `mono bootstrap` would
fail to run due to `npm install` failing to run the extension
installation script, as the latter now depends on the Transmitter,
which has not been built yet. This commit makes the install
command run the build when running locally.


packages/nodejs/scripts/extension/extension.js
```diff
@@ -47,36 +48,40 @@ function downloadFromMirror(mirror, filename, outputPath) {
...
-      .get(url, response => {
-        const { statusCode } = response
+    const transmitter = new Transmitter(url)
 
-        if (statusCode >= 400) {
-          reject(
-            new DownloadError(
-              `Request to CDN failed with code HTTP ${statusCode}`,
-              url
+    const promise = transmitter.download()
```
luismiramirez, line 53:
```suggestion
    const downloadPromise = transmitter.download()
```

What about renaming this one?

luismiramirez (approved):

packages/nodejs/src/transmitter.ts
```diff
@@ -4,79 +4,171 @@ import http from "http"
...
-            resolve({ status: responseStatus, body: parsedResponse })
-          })
+        onError(error) {
+          reject({ kind: "requestError", error })
+        }
       })
+    })
+  }
+
+  public async transmit() {
```
tombruijn, line 48: Later on I'd rename this to `post`. Doesn't have to be in this PR.

packages/nodejs/src/__tests__/transmitter.test.ts
```diff
@@ -114,6 +105,165 @@ describe("Transmitter", () => {
...
+      const { callback, onData, onError } = await transmitterRequest(
+        "GET",
+        "http://example.invalid/foo"
+      )
+
+      expect(callback).toHaveBeenCalledWith(
+        expect.objectContaining({ statusCode: 200 })
+      )
+
+      expect(onData).toHaveBeenCalledWith(Buffer.from("response body"))
```
tombruijn, line 210: Do we need to test this (the response body and status code), on what is basically a private function, if the transmit and download function tests also already tests this?

packages/nodejs/src/transmitter.ts
```diff
@@ -4,79 +4,171 @@ import http from "http"
...
+        },
+
+        onError(error) {
+          reject({ error: error })
+        }
       })
+    })
+  }
+
+  public request({
```
tombruijn, line 82:
```suggestion
  private request({
```
If possible I'd like to make this private. We probably shouldn't call it directly.

tombruijn (changes requested):

packages/nodejs/src/transmitter.ts
```diff
@@ -4,79 +4,171 @@ import http from "http"
...
-  public async transmit() {
-    return new Promise<HashMap<any>>((resolve, reject) => {
-      const config_data = this.#config.data
-      const params = new URLSearchParams({
-        api_key: config_data["pushApiKey"] || "",
-        name: config_data["name"] || "",
-        environment: config_data["environment"] || "",
-        hostname: config_data["hostname"] || ""
-      })
+  public download() {
```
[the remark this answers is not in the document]
unflxw: > If we every want to do a "normal GET request" I wonder if we can make this optional by adding a stream: true type of argument here, or move it to a getStream function.

I'm happy to implement some sort of `stream: true`, and that way we can have `.request` take care of the handling of the response to some extent. When looking at the original implementation of the extension download, I drew the line at the separation between the HTTP request and the filesystem -- I left the filesystem handling bits to the extension, and moved the HTTP request bits to the transmitter. In that sense, it's true that `.download` is not a good name, as it doesn't really download anything by itself. I'll apply this change and maybe rename it to `.downloadStream` or `.getStream` as you suggested.

>  Can we return the fully read response to the resolve and write it directly without streaming? It's probably not very efficient, right?

In practice, it wouldn't be a problem if we did. The agent is just a few megabytes. It does feel to me like piping the response stream to the file stream is the idiomatic Node.js thing to do, though.

> I kind of want to rename this to get, but I see we pass through the response body stream to the file streamer. 

I have aimed for `.request` to be the "reasonable enough" interface for making generic HTTP requests, and `.transmit` and `.download` are the specific wrappers for the unique things that our application currently wants to do. Based on that, I'm not so sure about renaming them to `.get` and `.post`, precisely because they don't do normal GET and POST requests, they do somewhat specific things instead:

- `.transmit` is "do a POST request with these four specific values from our config as query parameters, and resolve to the status code and the response value parsed as JSON, discarding it if it's not valid JSON"
- `.download` is a bit more reasonable, but it's still "do a GET request and resolve to the resulting return stream, or reject if the status code is > 400"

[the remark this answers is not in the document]
tombruijn: > I'm happy to implement some sort of stream: true, and that way we can have .request take care of the handling of the response to some extent.

I wouldn't do that now. Only when we come back to this and need to add more types of requests to it. For now I'd rename this function to `downloadStream` to make clear it's not a normal "GET" or download type of request.

> In practice, it wouldn't be a problem if we did. The agent is just a few megabytes. It does feel to me like piping the response stream to the file stream is the idiomatic Node.js thing to do, though.

Yeah let's keep it this way.

packages/nodejs/src/__tests__/transmitter.test.ts
unflxw: On this comment, and the one below about making it private: I've been thinking of this refactor as going in the opposite direction. I think that, if we implemented a good enough `.request` method, we could get rid of the `.transmit` and `.download` helper methods entirely, and just call `.request` with the needed parameters whenever we need to perform these requests.

I think it's getting there, but for now the caller still needs to provide a callback that figures out what to do with the response object, and it still needs to wrap the whole thing in a promise itself. If the `.request` method took care of that, too, then using it would be fairly straightforward, and the helpers would be unnecessary.

packages/nodejs/src/transmitter.ts
unflxw: It's not possible to write tests against it if it's private. I wouldn't want to duplicate the tests for the CA behavior between `.transmit` and `.downloadStream`.

packages/nodejs/src/transmitter.ts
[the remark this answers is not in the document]
unflxw: Renamed it to `.downloadStream`.

packages/nodejs/scripts/extension/extension.js
unflxw: Ended up getting rid of this `const` declaration entirely, to avoid giving this a name at all. 🎉

tombruijn (approved):

jeffkreeftmeijer (approved):

packages/nodejs/src/__tests__/transmitter.test.ts
tombruijn: That sounds good. I'm not sure if we need to update that right now. Let's focus on some other things first. Can you create an issue with your plans for a refactor?

PR #606, opened 2022-02-08
Add tests for http outgoing integration

tombruijn:
Add a TestRegistry (name suggestions welcome!) to store Spans that are
created during the test. This registry will be accessible after the test
to fetch all created spans. It also mocks out the `Span.close` function
so we can call `toObject` on them later, and the Extension doesn't
return an empty String for the Span.

It currently only tests `ChildSpan`s, but this can be updated the same
way for RootSpans.

## About this implementation

This implementation is much the same as the Ruby gem and does not rely
on mock objects (`MockTracer`, `MockSpan`) that may or may not mimic
their real world behavior.

To avoid all the hassle of this Span creation tracking, because this was
not easy to figure out in Node.js, I'm thinking adding a test mode to
the extension where we can just ask which spans it created during the
test. That way we also only have to implement this once for all
integrations.
https://github.com/appsignal/appsignal-agent/issues/744

[skip changeset]

### DIFF
--- a/packages/nodejs/src/instrumentation/http/lifecycle/__tests__/outgoing.test.ts
+++ b/packages/nodejs/src/instrumentation/http/lifecycle/__tests__/outgoing.test.ts
@@ -61,8 +61,8 @@ describe("HTTP outgoing requests", () => {
       attributes: { "appsignal:category": "request.http", method: "GET" },
       error: null,
       name: "GET http://localhost",
-      sample_data: {}
+      sample_data: {},
+      closed: true
     })
-    expect(SpanTestRegistry.closedSpans).toContain(lastSpan)
   })
 })
--- a/packages/nodejs/src/instrumentation/http/lifecycle/__tests__/outgoing.test.ts
+++ b/packages/nodejs/src/instrumentation/http/lifecycle/__tests__/outgoing.test.ts
@@ -1,0 +1,68 @@
+import { SpanTestRegistry } from "../../../../../test/registry"
+import nock from "nock"
+import { HashMap } from "@appsignal/types"
+import { BaseClient as Client } from "../../../../client"
+import { BaseTracer as Tracer } from "../../../../tracer"
+import { instrument } from "../../http"
+import { ChildSpan } from "../../../../span"
+
+describe("HTTP outgoing requests", () => {
+  const name = "TEST APP"
+  const pushApiKey = "TEST_API_KEY"
+
+  let http: any
+  let client: Client
+  let tracer: Tracer
+
+  // enableMinutelyProbes is set to false so we don't leak timers
+  const DEFAULT_OPTS = {
+    active: true,
+    name,
+    pushApiKey,
+    enableMinutelyProbes: false
+  }
+
+  beforeEach(() => {
+    SpanTestRegistry.clear()
+    client = new Client({ ...DEFAULT_OPTS })
+    tracer = new Tracer()
+    http = require("http")
+    instrument(http, tracer).install()
+  })
+
+  async function performRequest() {
+    return new Promise<HashMap<any>>((resolve, reject) => {
+      const options = {
+        host: "example.com",
+        path: "/foo",
+        method: "GET"
+      }
+      const request = http.request(options, (stream: any) => {
+        stream.on("data", () => {}).on("end", resolve)
+      })
+      request.on("error", (error: Error) => {
+        reject({ error: error })
+      })
+      request.end()
+    })
+  }
+
+  it("creates a span for the outgoing HTTP request", async () => {
+    nock("http://example.com").get("/foo").reply(200, "response body")
+    const span = tracer.createSpan() // Create any root span
+    expect(SpanTestRegistry.spans).toHaveLength(1)
+
+    await performRequest()
+
+    expect(SpanTestRegistry.spans).toHaveLength(2)
+    const lastSpan = SpanTestRegistry.lastSpan()
+    expect(lastSpan).toEqual(expect.any(ChildSpan))
+    expect(lastSpan.toObject()).toMatchObject({
+      attributes: { "appsignal:category": "request.http", method: "GET" },
+      error: null,
+      name: "GET http://localhost",
+      sample_data: {}
+    })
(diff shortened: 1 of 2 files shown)


packages/nodejs/test/registry.ts
```diff
@@ -0,0 +1,46 @@
+import { Span, SpanOptions, SpanContext, SpanData } from "../src/interfaces"
+import { RootSpan, ChildSpan } from "../src/span"
+
+export class TestRegistry {
```
unflxw, line 4: Consider giving this a more specific name about _what_ it registers (`ChildSpanRegistry`?)

packages/nodejs/test/registry.ts
```diff
@@ -0,0 +1,46 @@
...
+    const length = this.spans.length
+    return this.spans[length - 1]
+  }
+
+  static clear() {
+    this.spans = []
+  }
+}
+
+jest.mock("../src/span", () => {
```
unflxw, line 26: I think the fact that requiring this file automatically mocks the `ChildSpan` class for you is somewhat unexpected. Maybe this could be moved to a `.mock` function in the registry that we explicitly call in the test's `.beforeAll()`?

It would also be nice, in turn, to have an `.unmock` function that calls `jest.unmock("../src/span")`, which we can explicitly call from the test's `.afterAll()`.

packages/nodejs/test/registry.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export class TestRegistry {
+  static spans: Span[] = []
+  static closedSpans: Span[] = []
+
+  static addSpan(span: Span) {
+    this.spans.push(span)
+  }
+
+  static addClosedChildSpan(span: Span) {
```
unflxw, line 12: This method says `ChildSpan`, but the other ones just say `Span`. Is there a situation where `addSpan` could be called with a `RootSpan`? If not, then I'd pick one term or the other.

packages/nodejs/test/registry.ts
```diff
@@ -0,0 +1,53 @@
+import { Span, SpanOptions, SpanContext, SpanData } from "../src/interfaces"
+import { RootSpan, ChildSpan } from "../src/span"
```
unflxw, line 2: I think `SpanData` and `RootSpan` are unused.

packages/nodejs/src/instrumentation/http/lifecycle/__tests__/outgoing.test.ts
```diff
@@ -0,0 +1,89 @@
...
+import {
+  Span,
+  SpanOptions,
+  SpanContext,
+  SpanData
+} from "../../../../interfaces"
+import { BaseClient as Client } from "../../../../client"
+import { BaseTracer as Tracer } from "../../../../tracer"
+import { PLUGIN_NAME, instrument } from "../../http"
+import { RootSpan, ChildSpan } from "../../../../span"
```
unflxw, line 13: I think `RootSpan` and `SpanData` are unused here, too.

packages/nodejs/src/instrumentation/http/lifecycle/__tests__/outgoing.test.ts
```diff
@@ -0,0 +1,89 @@
...
+            if (responseStatus !== 200) {
+              reject(value)
+            } else {
+              resolve(value)
+            }
+          })
+      })
+      request.on("error", (error: Error) => {
+        reject({ error: error })
+      })
```
unflxw, line 66: If we don't care about the request, other than that we wait for it and it doesn't error, then:

```suggestion
        stream
          .on("data", () => {})
          .on("end", resolve)
      })
      
      request.on("error", reject)
```

packages/nodejs/test/registry.ts
unflxw: Also, I'm not sure what to make of the note in Jest's documentation that says:

> Modules that are mocked with `jest.mock` are mocked only for the file that calls `jest.mock`. Another file that imports the module will get the original implementation even if it runs after the test file that mocks the module.

Because, uh, this is working, right? So this is clearly not true. I wonder if by "file" they mean "test file"?

packages/nodejs/test/registry.ts
luismiramirez: What about `SpanTestRegistry`?

packages/nodejs/test/registry.ts
tombruijn: Yes. I'm renaming things to just support both types of spans.

packages/nodejs/test/registry.ts
unflxw: Tracking them all in one registry sounds good to me too.

packages/nodejs/test/registry.ts
tombruijn: > {before/after}All, not {before/after}Each -- so, I meant to mock when the describe block containing the tests starts, and unmock when it ends. Jest's naming is confusing.

You mean to unmock at the end of every test file? Jest does that automatically. No need to manually unmock.

> I found a workaround for that! Instead of keeping a list of closed spans and mocking out the closing, keep the state from before the span is closed and use it after it's closed. [Have a look!](https://github.com/appsignal/appsignal-nodejs/compare/http-outgoing-tests...http-outgoing-tests-keep-state-close)

Ah you found a way to make it work! I was trying to implement that, the same way it [works in the Ruby gem tests suite](https://github.com/appsignal/appsignal-ruby/blob/3f503ade83f22f4b0d86d76ea00e5f4dd3c56b6f/spec/support/testing.rb#L79), but I kept running into issues with calling super in the constructor like `const span = super(...)` and adding new fields on the class. I see in my latest version that's no longer an issue and we can apply your change. Can you fix up the commit for a PR based on this one? Or just push it on this branch.

unflxw: Pushed it to this branch.

luismiramirez (approved):

packages/nodejs/test/registry.ts
tombruijn: @unflxw looks good. If you're happy with it, please approve the PR so we can merge it.

unflxw (approved):

PR #611, opened 2022-02-10
Show extension load error on require

### LINKED ISSUE #555: Revisit printing extension load error on Extension start
At the moment we print the extension load error when the extension is started: https://github.com/appsignal/appsignal-nodejs/blob/58f1ae0b5ed62f9a92c736515a2efb2b5d2859da/packages/nodejs/src/extension.ts#L22-L31

In other integrations we do this on load: https://github.com/appsignal/appsignal-ruby/blob/ef2982fd7d1760f8062a20fa15a46131a0745cd2/lib/appsignal/extension.rb#L13-L19

We should check why this is done on start in the Node.js integration and if we can't do this on load as well.

If it's a problem with not supported Operating Systems printing this message every time on start, we can also think of a way to not always show this. Like an env var to disable to message or track if a user has ignored it somehow with a local config option.

unflxw:
Before this commit, any errors in loading the extension wrapper module wouldn't be shown if the `Extension` object wasn't started, and the `BaseClient` object would not attempt to start the extension if the extension wrapper wasn't loaded, meaning that errors in loading the extension wrapper were, in practice, never shown.

This commit changes those errors to be emitted when the extension wrapper module is required, instead of when the `Extension` object is started.

Fixes #555.

### DIFF
--- a/packages/nodejs/src/extension_wrapper.ts
+++ b/packages/nodejs/src/extension_wrapper.ts
@@ -6,12 +6,15 @@ try {
   mod = require("../build/Release/extension.node") as ExtensionWrapper
   mod.isLoaded = true
 } catch (e) {
+  console.error(
+    "AppSignal extension not loaded. This could mean that your current " +
+      "environment isn't supported, or that another error has occurred."
+  )
+
   mod = {
     isLoaded: false,
     extension: {
-      start() {
-        throw new Error("Extension module not loaded")
-      },
+      start() {},
       stop() {},
       diagnoseRaw() {},
       runningInContainer() {}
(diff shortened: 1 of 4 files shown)


packages/nodejs/src/extension_wrapper.ts
```diff
@@ -6,12 +6,15 @@ try {
   mod = require("../build/Release/extension.node") as ExtensionWrapper
   mod.isLoaded = true
 } catch (e) {
+  console.warn(
```
tombruijn, line 9:
```suggestion
  console.error(
```

It can probably be an error, because it's caused by an error. As long as it won't crash the program.

tombruijn (changes requested):

packages/nodejs/src/extension_wrapper.ts
unflxw: Done!

tombruijn (approved):

jeffkreeftmeijer (approved):

Issue #520, opened 2021-11-25
Add a centralized logger


tombruijn:
The Node.js integration only uses the `console.log/warn/debug/error` type functions to log things. It should log to "appsignal.log" instead if it has anything to log (that's not a deprecation warning).

See also the older issue about this: https://github.com/appsignal/appsignal-nodejs/issues/25

~🚨 Blocked on: https://github.com/appsignal/integration-guide/pull/102~

## To do

- [x] Create centralized logger.
    - [x] Look at other integrations for inspiration, such as prefixing the "stdout" logger with the `appsignal:` prefix.
    - [x] Implement a "file" based logger and an "stdout" logger.
    - [x] Probably needs to be an instance on the client object, accessible through the global client getter, or even has its own `logger` getter.
- [x] Update all log statements to use the centralized logger (with the exception of deprecation warnings and other messages that should be visible to the user).
- [x] Listen to the `log` config option and:
    - [x] Initialize a file logger if `log` is "file", which is the default.
    - [x] Initialize an stdout logger if `log` is "stdout".

tombruijn:
I added an issue to document how the logger works: https://github.com/appsignal/integration-guide/issues/99
Some more information can also be found there

tombruijn:
I've documented this in https://github.com/appsignal/integration-guide/pull/102
That has to be approved before working on this.

PR #620, opened 2022-02-22
Add centralized logger

luismiramirez:
## Add centralized logger using Winston

This commits adds a new central logger using Winston. It is set up when
the client initializes. It will fallback to stdout if the provided log
file path is not accessible.

There's a new static method in client that gives access to the logger
through the globally stored client after initialization.

## Send log messages through centralized logger

Log messages that need to be recorded are now sent through the
centralized logger.

---

### ToDo

- [x] Move logger operations from BaseClient to its own class
- [x] Proper integration testing

Fixes: #520

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -628,6 +628,24 @@
         "node": ">=0.1.95"
       }
     },
+    "node_modules/@colors/colors": {
+      "version": "1.5.0",
+      "resolved": "https://registry.npmjs.org/@colors/colors/-/colors-1.5.0.tgz",
+      "integrity": "sha512-ooWCrlZP11i8GImSjTHYHLkvFDP48nS4+204nGb1RiX/WXYHmJA2III9/e2DWVabCESdW7hBAEzHRqUn9OUVvQ==",
+      "engines": {
+        "node": ">=0.1.90"
+      }
+    },
+    "node_modules/@dabh/diagnostics": {
+      "version": "2.0.3",
+      "resolved": "https://registry.npmjs.org/@dabh/diagnostics/-/diagnostics-2.0.3.tgz",
+      "integrity": "sha512-hrlQOIi7hAfzsMqlGSFyVucrx38O+j6wiGOf//H2ecvIEqYN4ADBSS2iLMh5UFyDunCNniUIPk/q3riFv45xRA==",
+      "dependencies": {
+        "colorspace": "1.1.x",
+        "enabled": "2.0.x",
+        "kuler": "^2.0.0"
+      }
+    },
     "node_modules/@istanbuljs/load-nyc-config": {
       "version": "1.1.0",
       "resolved": "https://registry.npmjs.org/@istanbuljs/load-nyc-config/-/load-nyc-config-1.1.0.tgz",
@@ -1790,6 +1808,11 @@
         "node": ">=8"
       }
     },
+    "node_modules/async": {
+      "version": "3.2.3",
+      "resolved": "https://registry.npmjs.org/async/-/async-3.2.3.tgz",
+      "integrity": "sha512-spZRyzKL5l5BZQrr/6m/SqFdBN0q3OCI0f9rjfBzCMBIP4p75P620rR3gTmaksNOhmzgdxcaxdNfMy6anrbM0g=="
+    },
     "node_modules/asynckit": {
       "version": "0.4.0",
       "resolved": "https://registry.npmjs.org/asynckit/-/asynckit-0.4.0.tgz",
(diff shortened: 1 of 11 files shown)


packages/nodejs/src/client.ts
```diff
@@ -159,6 +170,52 @@ export class BaseClient implements Client {
...
+  private setUpLogger(): winston.Logger {
+    const logFilePath = this.config.logFilePath
+    const logLevel = String(this.config.data["logLevel"])
+    let transport
+
+    const logFormat = printf(({ level, message, timestamp }) => {
+      return `[${timestamp} (process) #${process.pid}][appsignal][${level}] ${message}`
+    })
+
+    if ((this.config.data["log"] = "file" && logFilePath)) {
```
tombruijn, line 186:
```suggestion
    if (this.config.data["log"] = "file" && logFilePath) {
```

packages/nodejs/src/client.ts
```diff
@@ -159,6 +170,52 @@ export class BaseClient implements Client {
...
+      case "error":
+        return "error"
+      case "warning":
+        return "warn"
+      case "info":
+        return "info"
+      case "debug":
+        return "debug"
+      case "trace":
+        return "silly"
```
tombruijn, line 213: sorry what? 😅

package-lock.json
```diff
@@ -16755,6 +17116,62 @@
         }
       }
     },
+    "winston": {
+      "version": "3.6.0",
+      "resolved": "https://registry.npmjs.org/winston/-/winston-3.6.0.tgz",
+      "integrity": "sha512-9j8T75p+bcN6D00sF/zjFVmPp+t8KMPB1MzbbzYjeN9VWxdsYnTB40TkbNUEXAmILEfChMvAMgidlX64OG3p6w==",
+      "requires": {
```
tombruijn, line 17123: It has quite a bit more dependencies than I would have hoped 🤷‍♂️

packages/nodejs/src/client.ts
```diff
@@ -159,6 +170,52 @@ export class BaseClient implements Client {
...
+      return `[${timestamp} (process) #${process.pid}][appsignal][${level}] ${message}`
+    })
+
+    if ((this.config.data["log"] = "file" && logFilePath)) {
+      transport = new winston.transports.File({ filename: logFilePath })
+    } else {
+      transport = new winston.transports.Console()
+    }
+
+    return winston.createLogger({
```
tombruijn, line 192: It may be better to create a wrapper class for this, so we control our own interface.

packages/nodejs/src/client.ts
luismiramirez: I also forgot an equal sign 🤦🏼 . I really need to find a way of testing this when I write the logger class.

packages/nodejs/src/client.ts
luismiramirez: Yep it's on to do

packages/nodejs/src/client.ts
luismiramirez: Yup, 😄 . Apparently, it is: "log a silly amount of information" and it's similar to "trace".  

https://github.com/winstonjs/winston#logging-levels

packages/nodejs/src/client.ts
tombruijn: Ah ok, I misunderstood that then.

unflxw (approved):

packages/nodejs/src/logger.ts
```diff
@@ -0,0 +1,71 @@
...
+    let transport
+
+    if (type == "file") {
+      transport = new winston.transports.File({ filename: filename })
+    } else {
+      transport = new winston.transports.Console()
+    }
+
+    const logFormat = printf(({ level, message, timestamp }) => {
+      return `[${timestamp} (process) #${process.pid}][appsignal][${level}] ${message}`
```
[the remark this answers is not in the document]
luismiramirez: Yep, I decided to keep it here to avoid adding more logic to the Logger. Do we really need to add that bit only to the stdout logger 🤔 ? I can't see a benefit for it.

unflxw:
Another nitpick (sorry 😅) -- the spec says the timestamp should be given in local time, not in UTC time. This one can be a bit confusing to the user when reading logs for the file, as the process logs with UTC time will be mixed with the agent and extension logs in local time.

packages/nodejs/src/logger.ts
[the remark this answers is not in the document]
unflxw: I don't really mind either way. I see how consistency with the format used by the agent and extension is a nice to have, but I don't think this is a deal breaker.

luismiramirez:
Thanks for catching that one @unflxw

packages/nodejs/.changesets/add-centralized-logger.md
```diff
@@ -0,0 +1,8 @@
+---
+bump: "patch"
+type: "add"
+---
+
+Log messages are now sent through a new centralized logger that writes to `/tmp/appsignal.log` by default.
+A warning is printed to STDERR when the default or provided logPath is not accessible, and the Logger
+automatically falls back to STDOUT.
```
tombruijn, line 8:
```suggestion
automatically falls back to STDOUT. Use config option `log: stdout` to log AppSignal messages to STDOUT instead.
```

packages/nodejs/src/logger.ts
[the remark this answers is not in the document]
tombruijn: The `[appsignal]` prefix was added back in the day to more easily be able to tell from which system the message comes from, in what can be a mix of different log messages.

packages/nodejs/src/interfaces/client.ts
```diff
@@ -22,6 +23,8 @@ export interface Client {
 
   config: Configuration
 
+  logger: Logger
```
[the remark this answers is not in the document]
luismiramirez: It'll increase "safety", yep.

tombruijn (approved):

Issue #604, opened 2022-02-08
Using ESLint for linting


tombruijn:
We currently use prettier to format our code, but we don't really "lint" our code base. Let's set up ESLint in this project as well. We need to use a [TypeScript compatible plugin](https://typescript-eslint.io/) probably. I also know there's a package like [prettier-eslint](https://www.npmjs.com/package/prettier-eslint). It's confusing to me. I don't think we necessarily want to remove prettier in favor of ESLint, even though ESLint also does some formatting, they just shouldn't disagree with each other.

## To do

- [x] Set up ESLint with TypeScript in a compatible way with prettier.
- [ ] When we are happy with the result, create a similar issue for the JavaScript repo
- [ ] Update handbook (issue) with new state: https://github.com/appsignal/handbook/issues/58. Leave a comment or update the issue description.

tombruijn:
@unflxw are you working on adding ESLint to the JavaScript repo as well? Or can we create a new issue for that?

unflxw:
@tombruijn Not actively working on it at the moment, but I do consider it part of this issue and want to get around to it. Now that we've figured out for Node, it hopefully won't be too hard to get it working on the JavaScript repo.

tombruijn:
Ok. Then I'm moving it back to "To do".

PR #622, opened 2022-03-04
Use ESLint

unflxw:
[skip changeset]

Part of #604.

## [Use ESLint](https://github.com/appsignal/appsignal-nodejs/commit/8b9c9dc110759111b29823e26178058c60fc9945)

Add the ESLint linter to the project, along with the Typescript, Prettier and Jest plugins for it, and configure ESLint to use them.

Replace uses of Prettier in the `lint` and `lint:write` actions and in the `lint-staged` pre-commit hook with ESLint.

Configure ESLint so that all rules that currently trigger errors or warnings are turned off. Subsequent commits will re-enable these rules along with the necessary fixes to the code.

## [Fix Jest linter warnings and errors](https://github.com/appsignal/appsignal-nodejs/commit/68fa649be5d0d8fe3169801f478f7828da8799ca)

Re-enable the following linter warnings and errors, fixing all issues detected:

- `jest/expect-expect`: In tests where a function is merely called to ensure calling it does not throw an error, wrap the call in an explicit `expect(() => call()).not.toThrow()`. To allow `expect` blocks to be reused across tests, configure the linter so that any function whose name begins with `expect` is considered an expect.

- `jest/no-done-callback`: In tests where the usage of the `done` callback is superfluous, remove it. In tests where it is called at the end of a callback, as an implicit assertion that the callback is called, wrap the callback in a `jest.fn()` and `expect` it to be called.

- `jest/no-identical-title`: Ensure different tests within the same describe block have different names. The only occurrence of this was a copy-paste mistake.

## [Apply automatic fixes for all lint rules](https://github.com/appsignal/appsignal-nodejs/commit/41f61833e4a210961ae031f27fc711f5120e2811)

As all the errors have been fixed, re-enable the `prefer-const` and `@typescript-eslint/no-inferrable-types` rules.

Fixes for `@typescript-eslint/ban-types` are also applied, replacing uses of object types (i.e. `Number`, the type for `new Number(10)`) with primitive types (i.e. `number`, the type for the literal `10`). Since other errors relating to this rule cannot be automatically fixed, the rule remains disabled.

## [Fix "no unused vars" linter warnings](https://github.com/appsignal/appsignal-nodejs/commit/5939622e03d6adffe2ad38ddfdff3322abd5043e)

Re-enable the `@typescript-eslint/no-unused-vars` rule.

Configure an exception to allow argument names starting with `_` to remain unused.

Prefix unused argument names with `_` and remove all other unused variables.

## [Fix JavaScript linter errors](https://github.com/appsignal/appsignal-nodejs/commit/18eb84d65145e20d626d92f26d75ce6a1a93742d)

Re-enable the `no-unreachable`, `no-useless-escape`, `no-var`, `prefer-rest-params` and `no-prototype-builtins` linter rules, and fix the linter errors from these rules.

In the `global.d.ts` file, declare the `__APPSIGNAL__` global variable in `NodeJS.Global`, not in `declare global`.

## [Fix "ban types" linter errors](https://github.com/appsignal/appsignal-nodejs/commit/e6074c684f17ae864cb847a3fe69ae5d17c72fc7)

Re-enable the `@typescript-eslint/ban-types` rule. Configure it to allow the use of the `Function` type, as it's used very sparingly in our codebase with no good alternative to it.

Fix other usages, where `{}` is used to denote an unknown type, by using `unknown` instead.

## Remaining disabled rules

- `@typescript-eslint/ban-ts-comment`: Reports an error on the use of `@ts-ignore`. There's only one usage of it, and we should look into it, but fixing it may require changes beyond the scope of this PR.

- `@typescript-eslint/no-empty-function`: Reports an error on the use of empty functions, such as `() => {}`, requiring at least a comment that makes it explicit that the function is intentionally empty.

- `@typescript-eslint/no-explicit-any`: Reports a warning on the explicit use of `any`. While some of our uses of `any` could be replaced with more specific types, I couldn't find an alternative for others, such as the return type of `JSON.parse`.

- `@typescript-eslint/no-non-null-assertion`: Reports a warning on the use of `!` to tell the TypeScript compiler that a nullable value will not be null in this case.

- `@typescript-eslint/no-this-alias`: Reports an error when a variable is assigned `this` as a value, as in `const that = this`. There are only three uses of this pattern in our codebase. I tried fixing them by renaming shadowing variables and using arrow functions, which preserve `this`, but the tests fail. I don't entirely understand why, so I'm leaving the fixes out of this PR, just in case.

- `@typescript-eslint/no-var-requires`: Reports an error on the use of require statements outside of import statements. Some of our uses of this are due to using `require` in unusual ways. Others (`require("package.json")`) should not be requires at all, and could be replaced with `JSON.parse`.

- `@typescript-eslint/triple-slash-reference`: Reports an error on the use of triple slash references, as in `/// <reference path="foo.d.ts" />`, which tell the TypeScript compiler to pull certain files. I tried removing this, as the file it references is also referenced in the `files` property in `tsconfig.json`, so it shouldn't be necessary. Works on my machine, but fails in CI. We should look into this.


tombruijn:
> @typescript-eslint/ban-ts-comment: Reports an error on the use of @ts-ignore. There's only one usage of it, and we should look into it, but fixing it may require changes beyond the scope of this PR.

We can look into this in another issue/PR yes. You already created issue #624 looks like.

> @typescript-eslint/no-empty-function: Reports an error on the use of empty functions, such as () => {}, requiring at least a comment that makes it explicit that the function is intentionally empty.

I think this is mostly a problem for "noop" functions, can we ignore the "noop" paths to make that easier? Or add an `eslintrc.js` to the `noop` dirs?

> @typescript-eslint/no-explicit-any: Reports a warning on the explicit use of any. While some of our uses of any could be replaced with more specific types, I couldn't find an alternative for others, such as the return type of JSON.parse.

It seems fine to keep `any` around for our use cases. We can leave it off.

> @typescript-eslint/no-non-null-assertion: Reports a warning on the use of ! to tell the TypeScript compiler that a nullable value will not be null in this case.

They want us to write it like this instead? `foo.bar?.includes('baz') ?? false;` That seems fine? Makes the "or" value more explicit. Does it not return a falsy value otherwise?

https://typescript-eslint.io/rules/no-non-null-assertion/

> @typescript-eslint/no-this-alias: Reports an error when a variable is assigned this as a value, as in const that = this. There are only three uses of this pattern in our codebase. I tried fixing them by renaming shadowing variables and using arrow functions, which preserve this, but the tests fail. I don't entirely understand why, so I'm leaving the fixes out of this PR, just in case.

Ah yes the two different type of arrow functions. If we can't fix in this PR we can revisit it at a later time. Shouldn't be a huge problem.

> @typescript-eslint/no-var-requires: Reports an error on the use of require statements outside of import statements. Some of our uses of this are due to using require in unusual ways. Others (require("package.json")) should not be requires at all, and could be replaced with JSON.parse.

Is this because we want to dynamically require the modules and patch them?
Or is this also bad? `const { DiagnoseTool } = require("../diagnose")`

Not sure if there would be another way to solve it for monkeypatches. It may be another thing we can't go around because of what our library does.

> @typescript-eslint/triple-slash-reference: Reports an error on the use of triple slash references, as in /// <reference path="foo.d.ts" />, which tell the TypeScript compiler to pull certain files. I tried removing this, as the file it references is also referenced in the files property in tsconfig.json, so it shouldn't be necessary. Works on my machine, but fails in CI. We should look into this.

I hope it's like a cache issue locally, that the project needs to be cleaned beforehand. What is the error you see on the CI? Is it important enough to create a separate issue for it, so that we can look at it after merging this PR?

tombruijn (approved):
The PR so far seems good. I left a comment with question about the disabled rules.

luismiramirez (approved):

unflxw:
@tombruijn Added commits fixing `@typescript-eslint/no-empty-function` and `@typescript-eslint/no-non-null-assertion`. Thank you!

Regarding the non-null assertion operator (`!`), it does not make the whole statement return false; it simply suppresses the type error about doing things with `null` or `undefined`. It forces the compiler to believe that the value preceding it is not `null` or `undefined`, but it doesn't itself compile to anything on the resulting JavaScript code.

That is, if `foo.bar.baz` in JavaScript would throw an exception when `foo.bar` is `undefined`, `foo.bar.baz` should not compile in properly-typed TypeScript, but `foo.bar!.baz` will compile -- but the resulting code will still throw an exception when it's undefined.

packages/nodejs/src/__tests__/config.test.ts
```diff
@@ -146,7 +146,9 @@ describe("Configuration", () => {
 
   describe("apiKey option", () => {
     it("sets the pushApiKey config option with the apiKey value", () => {
-      const warnMock = jest.spyOn(console, "warn").mockImplementation(() => {})
+      const warnMock = jest.spyOn(console, "warn").mockImplementation(() => {
+        return
```
tombruijn, line 150: To avoid having to do this for these functions, it looks like we can also only allow these arrow functions. I would be fine with that. We use these types of empty functions a bunch. For "methods" explicitly calling `return` is good.

```js
    "@typescript-eslint/no-empty-function": [
      "error",
      { allow: ["arrowFunctions"] }
    ],
```

tombruijn:
> Regarding the non-null assertion operator (!), it does not make the whole statement return false; it simply suppresses the type error about doing things with null or undefined. It forces the compiler to believe that the value preceding it is not null or undefined, but it doesn't itself compile to anything on the resulting JavaScript code.

For some reason I thought this would throw an error if the property was undefined 🤔 (Probably because of the `!` usage in Ruby method naming.)

So replacing `!` with an explicit `?? "alternative value"` we should get less weird behavior when an attribute is not set 👍 

> Yep, that's also disallowed. It wants us to do either import { DiagnoseTool } from "../diagnose" or import { DiagnoseTool } from require("../diagnose"). I don't know if there's a difference between the two.

Best I can think of right now is the difference between ESM and CommonJS? `require` should be CommonJS`.

> And then there's places like the diagnose tool CLI itself, where we do stuff like require("./node_modules/@appsignal/nodejs/dist/cli/diagnose"), which is definitely not good. I'm not saying it's impossible to comply with this one, but it may take some effort.

Yeah.. That may require some more work. Which we can do another time. It would make things a lot easier if we decide to move the cli files into another package.

tombruijn (approved):

packages/nodejs/src/__tests__/config.test.ts
unflxw: That seems like a good compromise to me. 👍

jeffkreeftmeijer (approved):

PR #625, opened 2022-03-09
Allow nested values in setSampleData

### LINKED ISSUE #624: Fix `setSampleData` TypeScript signature
As mentioned in #622, [we use the `@ts-ignore` annotation to ignore a signature mismatch on the Express middleware](https://github.com/appsignal/appsignal-nodejs/blob/main/packages/express/src/middleware/index.ts#L51). This is not actually any sort of "parse error", just a good old type mismatch: we're passing an arbitrarily-nested object to a function whose declared signature says it accepts objects with one or two levels of nesting. More specifically:

```typescript
// `query` is a `ParsedQs` from the `qs` library, which is defined as:
interface ParsedQs { [key: string]: undefined | string | string[] | ParsedQs | ParsedQs[] }
// see https://github.com/DefinitelyTyped/DefinitelyTyped/blob/5a2c5035e657b5b7a20ac14377e48344c3e00bd5/types/qs/index.d.ts#L57
// this is a recursive type, meaning ParsedQs is an arbitrarily-nested object

// we pass it to `span.setSampleData`, which is defined as:
public function setSampleData(
  key: string,
  data:
    | Array<
        HashMapValue | Array<HashMapValue> | HashMap<HashMapValue> | undefined
      >
    | HashMap<
        HashMapValue | Array<HashMapValue> | HashMap<HashMapValue> | undefined
      >
): this
// see https://github.com/appsignal/appsignal-nodejs/blob/ae012e70f6aa001f0e79ccf7aab1d23506e9e03d/packages/nodejs/src/interfaces/span.ts#L90-L99
// which is not recursive, it only allows one or two levels of nesting

// therefore, the following fails:
span.setSampleData("params", { ...params, ...query })
```

We should probably address this by fixing the type signature on `span.setSampleData`, as it should accept recursive types. Afterwards, we should remove the `@ts-ignore` annotation and re-enable the ESLint rule that disallows its use.

unflxw:
Fixes #624.

Fix the signature of setSampleData so that it accepts any array or object, instead of only arrays or objects with up to two levels of nesting. This allows the user to pass arbitrary objects to sample data properties that accept them, such as `custom_data`.

Note that not all sample data properties accept nested values.

### Notes

I tried to implement only allowing arbitrary nesting for some keys, while enforcing one level of nesting for other keys in the signature, [as discussed in the issue](https://github.com/appsignal/appsignal-nodejs/issues/624#issuecomment-1060437678). In principle, it is possible to do so using function signature overloads:

```typescript
type SampleDataValue = string | number | bigint | boolean | null | undefined

type FlatSampleData = {
  [key: string]: SampleDataValue
} | SampleDataValue[]

type NestedSampleData = {
  [key: string]: SampleDataValue | NestedSampleData
} | (SampleDataValue | NestedSampleData)[]

interface Span {
  public setSampleData(
    key: "environment" | "tags",
    data: FlatSampleData
  ): this
  
  public setSampleData(
    key: "params" | "session_data" | "custom_data",
    data: NestedSampleData
  ): this
}
```

There's a few reasons why I opted for a simpler function signature:

- It is not possible to specify that "non-plain objects" should also be allowed in the `FlatSampleData`. Given our current implementation of `Data.generate`, passing a non-plain object to it will call `.toString` on it. So you should be able to pass those to `FlatSampleData`, but it's not possible, as far as I know, to discriminate between these kinds of objects at the type signature level.
- [Overloads are tricky!](https://www.typescriptlang.org/docs/handbook/2/functions.html#writing-good-overloads) Specifically, each invocation of an overloaded function can only represent a call to one of the possible overloads. That is, you couldn't do something like `{"tags": {}, "params": {}}.entries().forEach(([key, data]) => span.setSampleData(key, data))`, because each invocation inside the loop corresponds to a different overload. The TypeScript handbook itself says: "Always prefer parameters with union types instead of overloads when possible".
- [We have a pending task to replace `setSampleData(key, data)` with functions of the form `setKey(data)`.](https://github.com/appsignal/integration-guide/issues/80) That seems like a much less surprising way to enforce different signatures for different sample data values.

### DIFF
--- a/packages/express/src/middleware/index.ts
+++ b/packages/express/src/middleware/index.ts
@@ -45,11 +45,6 @@ export function expressMiddleware(appsignal: Client): RequestHandler {
           span.setName(`${method} ${baseUrl}${req.route.path}`)
         }
 
-        // defeated the type checker here because i'm pretty sure the error
-        // `tsc` returns is actually a parse error
-        // @TODO: keep an eye on this
-        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-        // @ts-ignore
         span.setSampleData("params", { ...params, ...query })
         span.setSampleData("environment", filteredHeaders)
 
--- a/packages/nodejs/.changesets/allow-nested-values-in-span-setsampledata.md
+++ b/packages/nodejs/.changesets/allow-nested-values-in-span-setsampledata.md
@@ -1,0 +1,9 @@
+---
+bump: "patch"
+type: "fix"
+---
+
+Allow nested values in `Span.setSampleData`. This change also allows
+values other than strings, integers and booleans to be passed as values
+within the sample data objects. Note that not all sample data keys allow
+nested values to be passed.
--- a/packages/nodejs/src/interfaces/span.ts
+++ b/packages/nodejs/src/interfaces/span.ts
@@ -1,4 +1,4 @@
-import { HashMap, HashMapValue } from "@appsignal/types"
+import { HashMap } from "@appsignal/types"
 
 /**
  * The state of a `Span` at initialization time.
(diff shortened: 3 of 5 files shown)


tombruijn (approved):
Annoying change looks like. This result is better than what we had before, now people can send arbitrary data at least. Unfortunately they can also do that for tags, which we ignore.

jeffkreeftmeijer (approved):

luismiramirez (approved):

thijsc (approved):

PR #635, opened 2022-03-28
Delete job specific cache in CI jobs

### LINKED ISSUE #631: Use Semaphore artifacts instead of the cache
Currently, our build system uses the Semaphore cache system to pass the build artifact from the Build stage to the testing stages. [The Semaphore artifact system would be a better fit for this use case.](https://docs.semaphoreci.com/essentials/artifacts/#workflow-artifacts)

tombruijn:
To move build artifacts around from job to job in the same workflow,
make sure it uses the artifacts generated by this workflow's run.

Caches are not overwritten if it already exists. Make sure to always
remove the cache if it exists. Otherwise a retry of a build may to
continue to use the cache from the first time the workflow was run,
which is not what we want and may introduce silent bugs.

I did not choose to use artifacts here as artifacts may incur an
addition charge later on. By deleting the cache we get the same
behavior.

Closes #631
[skip changeset]

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -87,6 +87,7 @@ blocks:
     - name: Build
       commands:
       - mono build
+      - cache delete $_PACKAGES_CACHE-packages-$SEMAPHORE_GIT_SHA-v$NODE_VERSION
       - cache store $_PACKAGES_CACHE-packages-$SEMAPHORE_GIT_SHA-v$NODE_VERSION packages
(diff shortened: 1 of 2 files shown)


tombruijn:
Waiting for #636 to be merged to fix the build.

tombruijn:
You can now see in the build output the cache is being deleted: https://appsignal.semaphoreci.com/jobs/fed7013f-1640-4b3c-9ed0-cb9e530a5606

PR #639, opened 2022-03-30
Follow redirects on transmitter requests

### LINKED ISSUE #505: Checksum verification fails some of the time
Every so often a job on the CI fails because the checksum verification we do on the agent & extension tar.gz file download fails.

We should figure out why because it's getting annoying, and affecting some of our customers.

This is made difficult because the `nodejs-ext` package will delete itself upon failure, deleting the downloaded tar file. So it's not possible to look at the downloaded tar file later at the moment.

We should store the downloaded tar file somewhere outside of the package directory (such as `/tmp`) upon failure so we can inspect it.

Related issue in https://github.com/appsignal/appsignal-nodejs/issues/471, merging the `nodejs` and `nodejs-ext` packages, which means the package cannot fail the install and the downloaded file is still accessible after a failed checksum verification. We can choose to pick this up first, so we don't have to write something special for the current structure that will be unnecessary later on.

~🚨 Blocked on: https://github.com/appsignal/appsignal-nodejs/issues/471~ See https://github.com/appsignal/appsignal-nodejs/pull/545

## Support conversations

- https://app.intercom.com/a/apps/yzor8gyw/inbox/inbox/conversation/16410700073232 (private link)

## To do

- [x] Print the install report on extension installation failure.
- [x] Debug a job or (user) install when it fails and take a look at the downloaded archive file.

unflxw:
Implement following redirects for `Transmitter` requests. As per the HTTP spec and common browser behavior, redirects using status codes 301, 302 and 303 have the method changed to `GET`, and redirection loops are detected.

Fixes #505.

### DIFF
--- a/packages/nodejs/.changesets/follow-redirects-when-downloading-the-appsignal-agent.md
+++ b/packages/nodejs/.changesets/follow-redirects-when-downloading-the-appsignal-agent.md
@@ -1,0 +1,6 @@
+---
+bump: "patch"
+type: "fix"
+---
+
+Follow redirects when downloading the AppSignal agent. This fixes an issue where the redirect page would be downloaded instead of the agent, causing the agent installation to fail.
--- a/packages/nodejs/src/__tests__/transmitter.test.ts
+++ b/packages/nodejs/src/__tests__/transmitter.test.ts
@@ -15,6 +15,7 @@ describe("Transmitter", () => {
       nock.activate()
     }
 
+    nock.cleanAll()
     nock.disableNetConnect()
 
     originalEnv = { ...process.env }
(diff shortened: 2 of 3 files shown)


tombruijn (approved):
Looks good. I only did a quick review on GitHub. I don't have time anymore to check it out and test it today.

tombruijn (approved):
I tested it manually locally and triggered a redirect. Then I saw it worked based on the output below.

I triggered the redirect by switching to another geographic location using a VPN. I think that may cause cloudfront or fastly to point to another location.

```
!!! handleRedirectsCallback call
!!! handleRedirectsCallback fn call
!!! handleRedirectsCallback status: 307
!!! handleRedirectsCallback go redirect
!!! handleRedirectsCallback call
!!! handleRedirectsCallback fn call
!!! handleRedirectsCallback status: 200
!!! handleRedirectsCallback done
```

```diff
+++ packages/nodejs/src/transmitter.ts
@@ -134,12 +134,16 @@ export class Transmitter {
     callback,
     onError
   }: TransmitterRequestOptions): (stream: http.IncomingMessage) => void {
+    console.log("!!! handleRedirectsCallback call")
     return stream => {
+      console.log("!!! handleRedirectsCallback fn call")
       const responseStatus = stream.statusCode ?? 999
+      console.log("!!! handleRedirectsCallback status:", responseStatus)
       const isRedirect = REDIRECT_STATUS_CODES.indexOf(responseStatus) !== -1
       const newURL = stream.headers?.location

       if (isRedirect && typeof newURL !== "undefined") {
+        console.log("!!! handleRedirectsCallback go redirect")
         const redirectCount = callback[REDIRECT_COUNT] ?? 0

         if (redirectCount >= MAX_REDIRECTS) {
@@ -164,6 +168,7 @@ export class Transmitter {
           })
         }
       } else {
+        console.log("!!! handleRedirectsCallback done")
         callback(stream)
       }
     }
```

tombruijn:
@unflxw can this be merged?

unflxw:
@tombruijn Thanks for testing it on a real-world scenario! 👍 Merging it now.

PR #640, opened 2022-03-31
Fix `datamap.setFloat` implementation

### LINKED ISSUE #638: Error on `setFloat` for Data object
In PR https://github.com/appsignal/appsignal-agent/pull/756, we ran into the issue that `setFloat` caused an error about an invalid argument. This happened in the [spec/integration/opentelemetry_spec.rb](https://github.com/appsignal/appsignal-agent/pull/756/files#diff-a16593e93a9506767a8e72ea2bfd038b254f1b916f50753855a205fadbecc7ab).

The attributes we set where the following:

```
{ "string": "example string", "bool": true, "int": 10, "double": 12.3 },
```

We should check if this is a bug in the C++ extension.

```
Runner #1
Sudo: false
Output:
DEPRECATED: File names are no longer supported in the 'logPath' config option. Changing the filename to 'appsignal.log'
DEPRECATED: File names are no longer supported in the 'logPath' config option. Changing the filename to 'appsignal.log'
/Users/tombruijn/appsignal/agent/integrations/appsignal-nodejs/packages/nodejs/dist/internal/data.js:32
                        extension_wrapper_1.datamap.setFloat(key, value, map);
                                                    ^

Error: Invalid argument
    at /Users/tombruijn/appsignal/agent/integrations/appsignal-nodejs/packages/nodejs/dist/internal/data.js:32:53
    at Array.forEach (<anonymous>)
    at Function.mapObject (/Users/tombruijn/appsignal/agent/integrations/appsignal-nodejs/packages/nodejs/dist/internal/data.js:22:36)
    at Function.generate (/Users/tombruijn/appsignal/agent/integrations/appsignal-nodejs/packages/nodejs/dist/internal/data.js:8:25)
    at Extension.importOpenTelemetrySpan (/Users/tombruijn/appsignal/agent/integrations/appsignal-nodejs/packages/nodejs/dist/extension.js:29:166)
    at BaseClient.importOpenTelemetrySpan (/Users/tombruijn/appsignal/agent/integrations/appsignal-nodejs/packages/nodejs/dist/client.js:123:24)
    at run (/Users/tombruijn/appsignal/agent/spec/integration/runners/nodejs/opentelemetry.js:31:13)
    at Object.<anonymous> (/Users/tombruijn/appsignal/agent/spec/integration/runners/nodejs/opentelemetry.js:46:1)
    at Module._compile (node:internal/modules/cjs/loader:1101:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
```

unflxw:
The existing implementation for `datamap.setFloat` mistakenly attempted to use the second argument as both the value to add to the map and the map to add it to. This resulted in an "Invalid argument" error thrown when trying to cast the value as a map.

The tests for `Data.generate` tested sending floating point values to the agent using the value `99.0`. Since JavaScript has a unified `number` type for integers and floats alike, `99.0` is the same number as `99`.

The implementation of `Data.generate` decides whether to call `datamap.setInteger` or `datamap.setFloat` based on whether `Number.isInteger` returns `true` or `false` -- and it will return `true` for `99.0`, as it would for `99`. This means that, in practice, `datamap.setFloat` was not covered by the tests.

Fixes #638.

> _Three character diff_
> _Tricky to spot but quite_
> _Boring to explain_

### DIFF
--- a/packages/nodejs/.changesets/allow-sending-non-integer-values-in-data-maps.md
+++ b/packages/nodejs/.changesets/allow-sending-non-integer-values-in-data-maps.md
@@ -1,0 +1,6 @@
+---
+bump: "patch"
+type: "fix"
+---
+
+Fix an issue where the AppSignal extension would throw an error when an object containing a non-integer number is sent to it. This would be triggered when calling `setSampleData` with an object containing a non-integer number, or when the values for a metric's tags are non-integer numbers.
--- a/packages/nodejs/ext/appsignal_extension.cpp
+++ b/packages/nodejs/ext/appsignal_extension.cpp
@@ -182,7 +182,7 @@ Napi::Value SetFloatToDataMap(const Napi::CallbackInfo &info) {
   Napi::Number num = info[1].As<Napi::Number>();
 
   Napi::External<appsignal_data_t> map =
-      info[1].As<Napi::External<appsignal_data_t>>();
+      info[2].As<Napi::External<appsignal_data_t>>();
 
   appsignal_data_map_set_float(map.Data(), MakeAppsignalString(key_utf8),
                                num.DoubleValue());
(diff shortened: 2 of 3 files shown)


luismiramirez (approved):
🎣

PR #641, opened 2022-04-07
Use high-resolution host language timer

jeffkreeftmeijer:
Instead of falling back to the agent’s internal timer, use a high-resolution host language timer to match timings in OpenTelemetry.


thijsc (approved):

luismiramirez (approved):

unflxw (approved):
After this change, `span.closeSpan` (from the extension) is no longer used anywhere. Should it be removed?

packages/nodejs/.changesets/use-high-resolution-host-language-timer.md
```diff
@@ -0,0 +1,6 @@
+---
+bump: "patch"
+type: "change"
+---
+
+Use high-resolution host language timer
```
tombruijn, line 6:
```suggestion
Use the Node.js performance module as high-resolution timer for span creation and close times. This provides more accurate times and improves compatibility with other systems in the future.
```

packages/nodejs/src/__tests__/span.test.ts
```diff
@@ -11,6 +11,8 @@ describe("RootSpan", () => {
 
     expect(span).toBeInstanceOf(RootSpan)
     expect(internal.closed).toBeFalsy()
+    console.log(internal.start_time)
```
tombruijn, line 14:
```suggestion
```

packages/nodejs/src/utils.ts
```diff
@@ -42,3 +46,27 @@ export function isWritable(path: string) {
     return false
   }
 }
+
+/**
+ * Returns a high-resolution time tuple.
```
tombruijn, line 51:
```suggestion
 * Returns a high-resolution time tuple containing the current time in seconds and nanoseconds.
```

packages/nodejs/src/__tests__/utils.test.ts
```diff
@@ -9,4 +10,26 @@ describe("Utils", () => {
...
+        timeOrigin: 1648719688126.149,
+        now: function () {
+          return 116931.83823999763
+        },
+        clearMarks: perf_hooks.performance.clearMarks,
+        mark: perf_hooks.performance.mark,
+        measure: perf_hooks.performance.measure,
+        nodeTiming: perf_hooks.performance.nodeTiming,
+        timerify: perf_hooks.performance.timerify,
+        eventLoopUtilization: perf_hooks.performance.eventLoopUtilization
```
tombruijn, line 26:
```suggestion
        ...perf_hooks.performance,
        timeOrigin: 1648719688126.149,
        now: function () {
          return 116931.83823999763
        }
```

Easier to only customize what you really want to change without the noise of all the things you want to keep the default. And if any new things are added in the future we don't need to manually update it there.

tombruijn (approved):

tombruijn:
> After this change, `span.closeSpan` (from the extension) is no longer used anywhere. Should it be removed?

It would help to avoid confusion in the future about which function to use if we remove it.

jeffkreeftmeijer:
Took your suggestions, and removed the `closeSpan` function from the extension.

tombruijn (approved):

PR #651, opened 2022-04-29
Add OpenTelemetry span processor

jeffkreeftmeijer:
This patch adds an OpenTelemetry span processor that takes spans from OpenTelemetry and exports them to the AppSignal agent via its newly added `importOpenTelemetrySpan` function.

Blocked on a new agent release: https://github.com/appsignal/appsignal-agent/pull/756

### DIFF
--- a/packages/nodejs/.changesets/add-opentelemetry-span-processor.md
+++ b/packages/nodejs/.changesets/add-opentelemetry-span-processor.md
@@ -3,4 +3,4 @@ bump: "minor"
 type: "add"
 ---
 
-Add OpenTelemetry span processor
+Add OpenTelemetry span processor for the mysql and mysql2 packages. These are experimental integrations and require additional set up to integrate with OpenTelemetry.
--- a/packages/nodejs/.changesets/add-opentelemetry-span-processor.md
+++ b/packages/nodejs/.changesets/add-opentelemetry-span-processor.md
@@ -1,0 +1,6 @@
+---
+bump: "minor"
+type: "add"
+---
+
+Add OpenTelemetry span processor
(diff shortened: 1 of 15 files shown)


luismiramirez (approved):

packages/nodejs/.changesets/add-opentelemetry-span-processor.md
```diff
@@ -0,0 +1,6 @@
+---
+bump: "minor"
+type: "add"
+---
+
+Add OpenTelemetry span processor
```
tombruijn, line 6: This can use a bit more explanation what this means for the user, if it means anything. If users cannot use it yet, let's remove the changeset for now.

tombruijn (approved):

packages/nodejs/src/interfaces/span_processor.ts
```diff
@@ -0,0 +1,44 @@
+import { HrTime, Span, SpanAttributes, SpanContext } from "@opentelemetry/api"
+
+/**
+ * OpenTelemetrySpanProcessor is based on OpenTelemetry's internal span processor.
+ */
+export interface OpenTelemetrySpanProcessor {
```
unflxw, line 6: Instead of copying the types, I believe it should be possible to add the package with the types as a dev dependency and use `import type` to import only the type definitions.

unflxw (approved):

packages/nodejs/src/interfaces/span_processor.ts
tombruijn: @unflxw not sure what you mean here. If it's a dev dependency how can we use it once the package is shipped?

Issue #649, opened 2022-04-20
Add Node 18 to CI


unflxw:
[Node 18 is now available!](https://nodejs.org/en/blog/announcements/v18-release-announce/) It's the newest LTS version, so we should ensure we support it. Let's add it to the CI matrix.

[Node 12 will be EOL in ten days (2022-04-30).](https://github.com/nodejs/release) We could remove it from the CI matrix then.

PR #652, opened 2022-05-02
Add Node 18 to build matrix

luismiramirez:
Node 18 is now available and Node 12 is in EOL. This commits adds Node
18 to the build matrix and removes Node 12.

As Node 18 requires glibc 2.28 or newer, the matrix needs to run on at
least Ubuntu 20.04, so it's been updated.

We don't need to specify C 8 anymore for any build as all versions work
by default with the version included in the new OS image.

Closes: #649 

[skip changeset]

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -8,7 +8,7 @@ name: AppSignal for Node.js
 agent:
   machine:
     type: e1-standard-2
-    os_image: ubuntu1804
+    os_image: ubuntu2004
 auto_cancel:
   running:
     when: branch != 'main' AND branch != 'develop'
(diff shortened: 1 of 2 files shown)


unflxw (approved):
This is also meant to remove Node 17, right? Just pointing it out because it's not mentioned explicitly in the description.

Thanks for updating the image and getting rid of the GCC 8 hack in the process! 🎉

luismiramirez:
> This is also meant to remove Node 17, right? Just pointing it out because it's not mentioned explicitly in the description.

It doesn't go into EOL until 2022-06-01, I'll set up a reminder to remove it then.

tombruijn (approved):

Issue #650, opened 2022-04-27
Express Integration: Track request body by default


matiasgarcia:
I am wondering if there is room for tracking the response body, automatically, in the express integration.

We are doing this manually right now, but we would prefer to this come out of the box as other integrations (i.e ruby on rails one).

```
    span.setSampleData(
        "params", {
            "url": req.path,
            "method": req.method ,
            "params": req.params,
            "body": req.body,
            "query": JSON.stringify(req.query)
        }
    )
```

I took a look at the code, and maybe, this can be added in the following line?

https://github.com/appsignal/appsignal-nodejs/blob/main/packages/express/src/middleware/index.ts#L48

For example:

Rather than

`span.setSampleData("params", { ...params, ...query })`

Do

`span.setSampleData("params", { ...params, ...query, body: req.body })`

shairyar:
https://app.intercom.com/a/apps/yzor8gyw/inbox/inbox/4356044/conversations/16410700116213

tombruijn:
Hi @matiasgarcia, Thanks for the issue. We're aware of the problem. We're going to take a look at it, hopefully this week.

luismiramirez:
Hey @matiasgarcia! You can find the fix in the latest [@appsignal/express](https://www.npmjs.com/package/@appsignal/express/v/1.0.30) version.

Thank you very much for your collaboration!

PR #653, opened 2022-05-05
Send HTTP Post requests body params for Express

luismiramirez:
Body params were not bein sent to AppSignal in Express' HTTP
transactions. With this change, they're now properly sent.

Fixes: #650

### DIFF
--- a/packages/express/src/middleware/index.ts
+++ b/packages/express/src/middleware/index.ts
@@ -34,8 +34,20 @@ export function expressMiddleware(appsignal: Client): RequestHandler {
       res.end = function (this: Response, ...args: any) {
         res.end = originalEnd
 
-        const { method = "GET", params = {}, query = {}, headers } = req
+        const {
+          method = "GET",
+          params = {},
+          query = {},
+          body = {},
+          headers
+        } = req
         const filteredHeaders = filterHeaders(headers, appsignal)
+        const objectBody =
+          body instanceof Object &&
+          Object.keys(body).length > 0 &&
+          !(body instanceof Buffer)
+            ? { body: body }
+            : {}
 
         // if there is no error passed to `next()`, the span name will
         // be updated to match the current path
@@ -45,7 +57,7 @@ export function expressMiddleware(appsignal: Client): RequestHandler {
           span.setName(`${method} ${baseUrl}${req.route.path}`)
         }
 
-        span.setSampleData("params", { ...params, ...query })
+        span.setSampleData("params", { ...params, ...query, ...objectBody })
         span.setSampleData("environment", filteredHeaders)
 
         return res.end.apply(this, args)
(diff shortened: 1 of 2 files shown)


packages/express/src/middleware/index.ts
```diff
@@ -45,7 +51,7 @@ export function expressMiddleware(appsignal: Client): RequestHandler {
           span.setName(`${method} ${baseUrl}${req.route.path}`)
         }
 
-        span.setSampleData("params", { ...params, ...query })
+        span.setSampleData("params", { ...params, ...query, ...body })
```
tombruijn, line 54: I think we should check if the `body` is an object or not. And skip if it it's not. According to the types it can be `any`: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/ec3525c11ef9f60e731b9a0e9bd64bfc4bf9fdab/types/express/index.d.ts#L115

packages/express/src/middleware/index.ts
```diff
@@ -45,7 +52,7 @@ export function expressMiddleware(appsignal: Client): RequestHandler {
           span.setName(`${method} ${baseUrl}${req.route.path}`)
         }
 
-        span.setSampleData("params", { ...params, ...query })
+        span.setSampleData("params", { ...params, ...query, ...objectBody })
```
matiasgarcia, line 55: Not sure this is the best workaround. My suggestion was pretty much limited not considering edge cases.

Just to give you an example, what if params or query or body have a key with the same name? In that case they will be overriden.
Maybe body should be elsewhere?

luismiramirez: Hi @matiasgarcia, thanks for your feedback.

That would be unlikely, and an odd API design, but it's correct HTTP, and a POST request with query params and a body at the same time could happen.

I added the key `body` when the object is present as you suggested.

packages/express/src/middleware/index.ts
```diff
@@ -34,8 +34,18 @@ export function expressMiddleware(appsignal: Client): RequestHandler {
...
+        const {
+          method = "GET",
+          params = {},
+          query = {},
+          body = {},
+          headers
+        } = req
         const filteredHeaders = filterHeaders(headers, appsignal)
+        const objectBody =
+          body instanceof Object && Object.keys(body).length > 0
```
unflxw, line 46: This could also check that `!(body instanceof Buffer)`, to avoid sending the raw body bytes if the Express application uses [`express.raw()`](http://expressjs.com/en/api.html#express.raw).

unflxw (approved):

packages/express/src/middleware/index.ts
tombruijn: Probably a good idea to check it 👍 . If it's a Buffer the `Object.keys` function will return length > 0, but the keys are numbers (indexes).

luismiramirez: Thanks for noticing! Feedback applied.

tombruijn (approved):

PR #655, opened 2022-05-06
Update apollo-server-plugin-base dependency in CI

luismiramirez:
Our dependency for apollo-server-plugin-base was a few steps behind the
latest releases. This commit updates them.

Release tracker update: [private link](https://github.com/appsignal/release-tracker/pull/14)

[skip changeset]


unflxw (approved):
Change looks good. Future improvement maybe: I wonder if we should we check against different Apollo versions, and not just different Apollo plugin base versions?

tombruijn (approved):

PR #657, opened 2022-05-09
Ignore parent `NoopSpan` in `createSpan`

unflxw:
If a `NoopSpan` is passed as the parent span to be used by `tracer.createSpan`, instead of assuming it to be a `SpanContext` and creating a broken child span from it, behave as if no span has been passed.

This edge case can be accidentally triggered by passing the output of `tracer.currentSpan` to `tracer.createSpan`, which is something you might want to do if you want to create a span whose parent is the current span, as opposed to `tracer.createSpan`'s default behaviour of creating a span whose parent is the current **root** span.

Rewrite the tests for `.createSpan` so that it is possible to understand what behaviour is being tested without going back and forth between the tests and the implementation, and add a test for the above behaviour change.

### DIFF
--- a/packages/nodejs/src/__tests__/tracer.test.ts
+++ b/packages/nodejs/src/__tests__/tracer.test.ts
@@ -1,5 +1,6 @@
 import { BaseTracer as Tracer } from "../tracer"
 import { RootSpan, ChildSpan } from "../span"
+import { NoopSpan } from "../noops"
 
 describe("Tracer", () => {
   const name = "test"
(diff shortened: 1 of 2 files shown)


unflxw:
[Intercom conversation](https://app.intercom.com/a/apps/yzor8gyw/inbox/inbox/conversation/16410700118007#part_id=comment-16410700118007-14654219434) for context.

packages/nodejs/src/__tests__/tracer.test.ts
```diff
@@ -11,29 +11,69 @@ describe("Tracer", () => {
...
+
+        expect(childSpan).toBeInstanceOf(ChildSpan)
+        expect(childSpanData.parent_span_id).toEqual(rootSpanData.span_id)
+        expect(childSpanData.name).toEqual("childSpan")
+      })
+    })
+
+    describe("from a NoopSpan", () => {
+      it("behaves as if no span was passed", () => {
+        const noopSpan = tracer.currentSpan()
```
tombruijn, line 57:
```suggestion
        const noopSpan = tracer.currentSpan()
        expect(noopSpan).toBeInstanceOf(NoopSpan)
```

Let's check if it's the type of value we expect here. Or create a `NoopSpan` object directly.

```suggestion
        const noopSpan = new NoopSpan()
```

tombruijn (approved):

luismiramirez (approved):

PR #659, opened 2022-05-11
Import OpenTelemetry types rather than duplicate

tombruijn:
Use Typescript's `import type` and `export type` to include the
OpenTelemetry types into our package without having to copy the code
into our project directly.

This was suggested by @unflxw in PR https://github.com/appsignal/appsignal-nodejs/pull/651#discussion_r866916531

I'm not sure if this is more or less fragile. If the OpenTelemetry
package changes the interface we'd be shipping an older version that
doesn't work with older versions anymore.

The package from which we import is a dev dependency only, and used
during the compile step, and won't ship as a dependency in the final
release.

I got an issue when using the OpenTelemetry SpanProcessor type for the
`onEnd` function. It doesn't have a second argument.

The `onStart` also needed another type, from `SpanContext` to `Context`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1329,6 +1329,63 @@
         "node": ">=8.0.0"
       }
     },
+    "node_modules/@opentelemetry/core": {
+      "version": "1.2.0",
+      "resolved": "https://registry.npmjs.org/@opentelemetry/core/-/core-1.2.0.tgz",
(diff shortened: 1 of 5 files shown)


unflxw:
@tombruijn Yes, this is what I had in mind :) I hoped we could remove the file in `interfaces/` entirely, but if we want to re-export those types, that's okay.

If possible, I'd wait for @jeffkreeftmeijer's input on this as well? There may be something I'm missing that makes the previous approach preferable.

tombruijn:
Ok, asking for a review from Jeff as well.

luismiramirez (approved):

jeffkreeftmeijer (approved):
I added that interface to save us a dependency on `@opentelemetry/sdk-trace-base`, to allow us to only depend on the types package, but this seems to make more sense indeed.

Issue #632, opened 2022-03-25
Fix Next.js test example failure on Node <= 14


unflxw:
[See sample failed build from the `main` branch on Semaphore.](https://appsignal.semaphoreci.com/workflows/eca33c5f-22f8-47ff-8355-9fe64bdc336e?pipeline_id=709d6e85-a603-4c64-abcd-111cf1575d56)

This was uncovered after cleaning the caches and fixing the build step for Node <= 14; see #628, #629 and [this Slack conversation](https://appsignal.slack.com/archives/CNPP953E2/p1648126339853399). This failure may or may not be related to the switch to NPM version 7 for these Node versions, implemented in #629.

tombruijn:
If it takes too much work to fix right now. I would first disable the test, mark it pending, and get back to it after issue #631 again.

PR #661, opened 2022-05-16
Reenable Next.js flaky test

luismiramirez:
Remove skip clause from the integration test that failed for Node.js
v14 on CI

The most recent changes on the build matrix (Upgrading the base SO image, removing some old hotfixes for CPP) have probably fixed this? The test now works.

Fixes: #632 

[skip changeset]


unflxw (approved):
If it's working, then it's working! 😅

PR #664, opened 2022-05-20
Add options for disabling default instrumentation

### LINKED ISSUE #519: Figure out what to do with the ignoreInstrumentation option
I found that we have a config option called `ignoreInstrumentation`. This config option is undocumented in our docs page: https://docs.appsignal.com/nodejs/configuration/options.html
And it doesn't have a description in the code itself.

It seems to determine if we [automatically activate some plugins](https://github.com/appsignal/appsignal-nodejs/blob/1a2bef0a710ff9b40129fc90aa25bd2f4230f400/packages/nodejs/src/bootstrap.ts#L16-L37) or not: httpPlugin, httpsPlugin, redisPlugin, pgPlugin

- [ ] Do we want to keep this? Is there a scenario where we don't want to automatically enable those plugins?
- [ ] If we want to keep it, does it currently work for all plugins?
- [ ] Document the config option on the docs page if it is a public config option
- [ ] Document the config option in the code base if it's a private config option

tombruijn:
We previously had the undocumented configuration option
`ignoreInstrumentation` that didn't follow our previously set out naming
and behavior conventions for configuration options.

Split up this configuration option in separate configuration options for
the different instrumentations.

Closes #519


packages/nodejs/src/client.ts
```diff
@@ -73,7 +78,14 @@ export class BaseClient implements Client {
 
     this.instrumentation = new Instrumentation(this.tracer(), this.metrics())
 
-    initCorePlugins(this.instrumentation, { ignoreInstrumentation })
+    initCorePlugins(this.instrumentation, {
+      instrumentationConfig: {
+        http: instrumentHttp || true,
+        https: instrumentHttps || true,
+        pg: instrumentPg || true,
+        redis: instrumentRedis || true
```
unflxw, line 86: P.S. to the comment above: when reading the options from the `Configuration` object, their default `true` value from the `Configuration` defaults will apply, so it won't be necessary to add `|| true` to each of them.

packages/nodejs/src/client.ts
```diff
@@ -55,7 +55,12 @@ export class BaseClient implements Client {
...
   constructor(options: Partial<AppsignalOptions> = {}) {
-    const { ignoreInstrumentation } = options
+    const {
+      instrumentRedis,
+      instrumentHttp,
+      instrumentHttps,
+      instrumentPg
+    } = options
 
     this.config = new Configuration(options)
```
unflxw, line 65: If these configuration options are read directly from the options passed to the constructor, changing their value using environment variables will have no effect. (This bug was already present on the existing `ignoreInstrumentation` implementation)

That is, as it's currently implemented, setting any of the `APPSIGNAL_INSTRUMENT_*` environment variables to `false` won't actually disable that instrumentation.

For the environment variable values to be taken into account as well, these options should be read from the Configuration object, after it is initialised.

```suggestion
    this.config = new Configuration(options)
    
    const {
      instrumentRedis,
      instrumentHttp,
      instrumentHttps,
      instrumentPg
    } = this.config.data
```

unflxw (changes requested):
Requesting changes due to the environment variables issue, which is a problem now that these configuration options are going to be documented. 👍✅ otherwise.

unflxw:
A thought: do we want to have separate configuration options for instrumenting HTTP and HTTPS? This seems to be an implementation detail, that they're two separate instrumentations on our internals, due to how Node has two separate modules for them.

I can't think of a scenario where you'd want one but not the other, and I can imagine users setting `instrumentHttp` to `false` and being surprised that HTTPS requests still show up in AppSignal.

It might be better to have just the one `instrumentHttp` configuration toggle for disabling both HTTP and HTTPS instrumentation?

luismiramirez (approved):
LGTM after applying Noemi's feedback. I also agree with not separating the `http` and `https` instrumentation enabling.

packages/nodejs/src/client.ts
tombruijn: Good point. Looking at this with fresh eyes you can't disable it through the client. I run into the issue that if I remove `|| true` I get this error:

```
src/client.ts:82:9 - error TS2322: Type 'boolean | undefined' is not assignable to type 'boolean'.
  Type 'undefined' is not assignable to type 'boolean'.

82         http: instrumentHttp,
           ~~~~
```

I'll take a better look

jeffkreeftmeijer (approved):

packages/nodejs/src/client.ts
unflxw: Ah, yes, the types on the config. Since `this.config.data` is a `Partial<AppsignalOptions>`, every key in it could be undefined -- even if we know it won't be, because it has a default value. I wonder if it really needs to be declared as a partial. If every configuration option has a default, then it wouldn't need to be a partial.

Anyway, feel free to ignore my original comment. No need to mess with the types at this point.

packages/nodejs/src/config/configmap.ts
```diff
@@ -17,6 +17,10 @@ export const ENV_TO_KEY_MAPPING: { [key: string]: string } = {
   APPSIGNAL_IGNORE_ACTIONS: "ignoreActions",
   APPSIGNAL_IGNORE_ERRORS: "ignoreErrors",
   APPSIGNAL_IGNORE_NAMESPACES: "ignoreNamespaces",
+  APPSIGNAL_INSTRUMENT_HTTP: "instrumentHttp",
+  APPSIGNAL_INSTRUMENT_HTTPS: "instrumentHttps",
```
unflxw, line 21:
```suggestion
```

Since the HTTP and HTTPS settings have been merged into `instrumentHttp`, the `instrumentHttps` entry can be removed from this map.

packages/nodejs/src/config/configmap.ts
```diff
@@ -80,6 +84,10 @@ export const JS_TO_RUBY_MAPPING: { [key: string]: string } = {
   ignoreActions: "ignore_actions",
   ignoreErrors: "ignore_errors",
   ignoreNamespaces: "ignore_namespaces",
+  instrumentHttp: "instrument_http",
+  instrumentHttps: "instrument_https",
```
unflxw, line 88:
```suggestion
```

Since the HTTP and HTTPS settings have been merged into `instrumentHttp`, the `instrumentHttps` entry can be removed from this map.

packages/nodejs/src/config.ts
```diff
@@ -127,6 +127,10 @@ export class Configuration {
       ignoreActions: [],
       ignoreErrors: [],
       ignoreNamespaces: [],
+      instrumentHttp: true,
+      instrumentHttps: true,
```
unflxw, line 131:
```suggestion
```

Since the HTTP and HTTPS settings have been merged into `instrumentHttp`, the `instrumentHttps` entry can be removed from this map.

packages/nodejs/src/__tests__/config.test.ts
```diff
@@ -27,6 +27,10 @@ describe("Configuration", () => {
     ignoreActions: [],
     ignoreErrors: [],
     ignoreNamespaces: [],
+    instrumentHttp: true,
+    instrumentHttps: true,
```
unflxw, line 31:
```suggestion
```

Since the HTTP and HTTPS settings have been merged into `instrumentHttp`, the `instrumentHttps` entry can be removed from this map.

packages/nodejs/.changesets/add-config-options-for-disabling-default-instrumentation.md
```diff
@@ -0,0 +1,24 @@
...
+bump: "patch"
+type: "add"
+---
+
+Add config options for disabling default instrumentation like HTTP, HTTPS, PostgreSQL (pg package) and Redis (node-redis package).
+
+The following configuration options have been added:
+
+- `instrumentHttp`
+- `instrumentHttps`
```
unflxw, line 11:
```suggestion
```

Since the HTTP and HTTPS settings have been merged into `instrumentHttp`, the `instrumentHttps` entry should be removed from the changelog.

packages/nodejs/src/client.ts
```diff
@@ -73,7 +71,14 @@ export class BaseClient implements Client {
 
     this.instrumentation = new Instrumentation(this.tracer(), this.metrics())
 
-    initCorePlugins(this.instrumentation, { ignoreInstrumentation })
+    const { instrumentRedis, instrumentHttp, instrumentPg } = this.config.data
+    initCorePlugins(this.instrumentation, {
+      instrumentationConfig: {
+        http: instrumentHttp,
```
unflxw, line 77:
```suggestion
        http: instrumentHttp,
        https: instrumentHttp,
```

The value of the `instrumentHttp` configuration option should be passed in this `instrumentationConfig` as the value for both the `http` and `https` keys. Otherwise, the HTTPS integration will never be disabled.

unflxw (commented):
When merging the `instrumentHttp` and `instrumentHttps` instrumentation options, some HTTPS-related things were left in there.

unflxw (approved):

PR #670, opened 2022-05-25
Lock apollo-server package versions in CI

tombruijn:
Make sure that the apollo-server package uses the same version as the
apollo-server-plugin-base package.

The apollo-server package is configured as matching any version (`*`) in
the test app's `package.json`. This should install the right version.

The apollo-server and apollo-server-plugin-base packages do not follow
the same release scheme so that's why the versions do not appear to
match.

I've also updated the versions against we test to be the last three
ones, except the 3.8 release (3.7 was skipped).

Related issue https://github.com/apollographql/apollo-server/issues/6469
[skip changeset]

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -117,19 +117,19 @@ blocks:
       - LANGUAGE=nodejs test/integration/diagnose/bin/test
     - name: "@appsignal/apollo-server - apollo-server@latest - integrations"
       commands:
-      - script/install_test_example_packages apollo-server apollo-server-plugin-base@latest
+      - script/install_test_example_packages apollo-server apollo-server@latest apollo-server-plugin-base@latest
+      - script/test_package_integration apollo-server
+    - name: "@appsignal/apollo-server - apollo-server@3.6.0 - integrations"
+      commands:
+      - script/install_test_example_packages apollo-server apollo-server@3.6.0 apollo-server-plugin-base@3.5.0
+      - script/test_package_integration apollo-server
+    - name: "@appsignal/apollo-server - apollo-server@3.5.0 - integrations"
+      commands:
+      - script/install_test_example_packages apollo-server apollo-server@3.5.0 apollo-server-plugin-base@3.4.0
       - script/test_package_integration apollo-server
     - name: "@appsignal/apollo-server - apollo-server@3.4.0 - integrations"
       commands:
-      - script/install_test_example_packages apollo-server apollo-server-plugin-base@3.4.0
-      - script/test_package_integration apollo-server
-    - name: "@appsignal/apollo-server - apollo-server@3.3.0 - integrations"
-      commands:
-      - script/install_test_example_packages apollo-server apollo-server-plugin-base@3.3.0
-      - script/test_package_integration apollo-server
-    - name: "@appsignal/apollo-server - apollo-server@3.2.0 - integrations"
-      commands:
-      - script/install_test_example_packages apollo-server apollo-server-plugin-base@3.2.0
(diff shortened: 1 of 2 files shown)


.semaphore/semaphore.yml
```diff
@@ -117,19 +117,19 @@ blocks:
       - LANGUAGE=nodejs test/integration/diagnose/bin/test
     - name: "@appsignal/apollo-server - apollo-server@latest - integrations"
       commands:
-      - script/install_test_example_packages apollo-server apollo-server-plugin-base@latest
+      - script/install_test_example_packages apollo-server apollo-server@latest apollo-server-plugin-base@latest
       - script/test_package_integration apollo-server
     - name: "@appsignal/apollo-server - apollo-server@3.4.0 - integrations"
       commands:
-      - script/install_test_example_packages apollo-server apollo-server-plugin-base@3.4.0
+      - script/install_test_example_packages apollo-server apollo-server@3.4.0 apollo-server-plugin-base@3.4.0
```
unflxw, line 124:
```suggestion
      - script/install_test_example_packages apollo-server apollo-server@3.5.0 apollo-server-plugin-base@3.4.0
```

.semaphore/semaphore.yml
```diff
@@ -117,19 +117,19 @@ blocks:
...
       - script/test_package_integration apollo-server
     - name: "@appsignal/apollo-server - apollo-server@3.3.0 - integrations"
       commands:
-      - script/install_test_example_packages apollo-server apollo-server-plugin-base@3.3.0
+      - script/install_test_example_packages apollo-server apollo-server@3.3.0 apollo-server-plugin-base@3.3.0
```
unflxw, line 128:
```suggestion
      - script/install_test_example_packages apollo-server apollo-server@3.4.0 apollo-server-plugin-base@3.3.0
```

unflxw (approved):
Unlike most other `apollo-server-*` packages, which release their versions in sync with each other (that is, `apollo-server@3.8.0` works with `apollo-server-express@3.8.0`), versions of `apollo-server-plugin-base` follow a different pace -- in fact, the latest version is `apollo-server-plugin-base@3.6.0`. This means that the right version of `apollo-server-plugin-base` to use with a given `apollo-server` version is not necessarily the one with the same version number. 🤦

Going by [the release history](https://github.com/apollographql/apollo-server/commits/main/packages/apollo-server-plugin-base/package.json):

- `apollo-server@3.8.0` [was released along with](https://github.com/apollographql/apollo-server/commit/c072d9f9a32ee67f4fd8f717cbbff6fe59a3be4e) `apollo-server-plugin-base@3.6.0` (these are the latest minors)
- `apollo-server@3.5.0` [was released along with](https://github.com/apollographql/apollo-server/commit/e6c3cbbf820b3ffff1e7e98d41b3bfc08f99b9b0) `apollo-server-plugin-base@3.4.0`
- `apollo-server@3.4.0` [was released along with](https://github.com/apollographql/apollo-server/commit/7861ba22c9e4caf037e7a1cec71191b9b0deb71b) `apollo-server-plugin-base@3.3.0`
- `apollo-server@3.3.0` was released without a corresponding release of `apollo-server-plugin-base`, so I assume it still used `apollo-server-plugin-base@3.2.0`
- `apollo-server@3.2.0` [was released along with](https://github.com/apollographql/apollo-server/commit/4dc19131aaea2416b5b2ccab295b06272f94c996) `apollo-server-plugin-base@3.2.0`

So, the version numbers need to be changed (see suggestions) -- other than that, LGTM 👍

tombruijn:
Thanks for spotting that @unflxw! I've also updated the packages to the latest couple versions.

luismiramirez (approved):

PR #672, opened 2022-05-27
Check if span is still open in ScopeManager.active

tombruijn:
## Check if span is still open in ScopeManager.active

The ScopeManager.active and ScopeManager.root function returned spans as
active or as a valid root span when they were closed. This is a
different state from a NoopSpan where no span has been active.

When a span closes it doesn't necessary reset the scope (remove the
executionAsyncId) from the scopes and roots maps on the ScopeManager.

Keep track if a span is open or not by keeping an `open` attribute on
the Span that is set to `false` whenever `Span.close()` is called. We
should not rely on the `isObject() => { closed: true }`, because that's
a more intensive operation (creating and parsing a JSON object) and we
should only use it for testing.

With the new `Span.isOpen()` function we can check if the Span is stil
open and only return it if that's the case. If it's not, also
immediately remove it from the ScopeManager's maps.

I've updated the `createRootSpan` test to close the span after the
assertions, otherwise it will now fail because of returning undefined.

We ran into this issue occurring in PR #669. It will not necessarily fix
the issue we ran into there, but it's an issue nonetheless.

## Refactor ScopeManager deleting ids from maps

Use a function to delete the executionAsyncId from the scopes and roots
maps, rather than repeating this behavior in several functions in the
same class. Keeps it more DRY.


packages/nodejs/src/span.ts
```diff
@@ -180,6 +186,15 @@ export class BaseSpan implements Span {
     }
   }
 
+  /**
+   * Returns if the span is still open or not.
+   *
+   * A closed span can't be modified or fetched attributes from.
+   */
+  public isOpen() {
```
luismiramirez, line 194: Why not making this a plain public property?

luismiramirez (approved):

packages/nodejs/src/span.ts
tombruijn: How can I do that without also making it writable? I don't want users to be able to modify it.

The definition `public readonly open: boolean` doesn't work.

```
error TS2540: Cannot assign to 'open' because it is a read-only property.

177     this.open = false
             ~~~~
```

luismiramirez: I see. I've found a couple of workarounds to make it mutable from the inside but it's probably too much code for a single property. 

LGTM

unflxw: For actual privacy enforced on the JavaScript side, we should use JavaScript private fields, not TypeScript's `private`. We already use those in the `#scopes` and `#roots` properties of the ScopeManager.

It's described as "hard private" in [the TypeScript handbook](https://www.typescriptlang.org/docs/handbook/2/classes.html#caveats):
> Unlike TypeScripts’s `private`, JavaScript’s [private fields](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields) (`#`) remain private after compilation and do not provide the previously mentioned escape hatches like bracket notation access, making them hard private.

Here's [a TypeScript playground](https://www.typescriptlang.org/play?#code/FAYwNghgzlAEAqBPADgUwMogE4EtkBcAFXANwn1QDEdUwATWAb2FldmVPNVgHs0A7AFywARjx5hUEfizYge-KPiwBXEPh5YAFAEomstm3wALHFAB0fVP1gBeWMpWoDsAL7B3oBUtgAzcXaw-KgA7ggoGNh4RJwU1LR0ANzAwAD0qQjG3BxWWPiIsGawSjhgYLAQICCoMDgikgA0QTwUDsbkbUXyALbIpTUOPGkZAAb+PJYCI7w2JtwAUujFOHTO8ooSqOZgPADmWgBEldUwfuKT1gdNWuMVcNKIOhf8OsMV-Aw4+LAg0qLcqDoX0BwESNwC0Hej2egV8EDAUDW3k22z2h2ONTg4xhEF8FCwsG6PCBvkQV1g4J4dyhTysLxS6UKnX4AGtYCFuFBjDwVPRYCpEbBFuxYtxfDR6HAcIoKBA6KBIKd5hAyJhcARiDgyHEJQxmIYAMR04RiTbSFKGdZKVTqTS6fSGQwmMzmI0CQKOZyGTyGZAqeo4ECwXaob50+2MWBYUMqLCzUwWN3WNweFJW74iCAE+zBMLK1VRDWi+L0BkZZ1SrHS7giVC-AXcDmwVbi4IMCDB0P4rwbSSo-ZHKqY0RZ57kzNYZ56MttbjyVb-HZhExYHghe42VBYVcE2v1wVzaMAcjg-CpIfw3YnMPscIRzjecx+xLnPG63WsFAYtaXhXuxUQfh8AgAAPWAtx3Jo9wgBtYBGEJ2goEgt1dEY3jMfgj2+eEl0Bfl+EkU5EB5I9o0KGwO3AaANwYOCEK4ZCsGmIoRgrNDGStFEdn2bEkxeIA) demonstrating the difference.

tombruijn: I've updated it.

luismiramirez (approved):

packages/nodejs/src/span.ts
```diff
@@ -17,6 +17,11 @@ import { BaseClient } from "./client"
  */
 export class BaseSpan implements Span {
   protected _ref: unknown
+  #_open: boolean
```
unflxw, line 20: Nitpick: the `#` already means "private" -- I'd drop the `_`, which also means "private".

unflxw (approved):
Looks good to me! :+1:

Feel free to ignore this: it might be desirable to apply the same check (`span && span.open`) that this PR adds to the `.active` or `.root` methods in the scope manager, to the logic that keeps track of which span is the active or root span in the first place. That is, when a new asynchronous context is created (in the `ScopeManager`'s constructor, inside the `init` function) do that same check against the currently known spans for `currentId`/`triggerId`, before propagating them to the newly initialised asynchronous context.

packages/nodejs/src/span.ts
tombruijn: Good catch!

tombruijn:
> Feel free to ignore this: it might be desirable to apply the same check (span && span.open) that this PR adds to the .active or .root methods in the scope manager, to the logic that keeps track of which span is the active or root span in the first place. That is, when a new asynchronous context is created (in the ScopeManager's constructor, inside the init function) do that same check against the currently known spans for currentId/triggerId, before propagating them to the newly initialised asynchronous context.

Let's address this in another PR. I think it's a valid concern. I didn't check this, and I hoped that it would all be routed through these functions, but I guess not. We need these checks there too.

PR #676, opened 2022-05-31
Check if span is open in ScopeManager

tombruijn:
Based on #675.

## Add tests for ScopeManager.root() function

These unit tests were missing, especially after the recent changes to
allow `root()` to only return open spans.

## Update bindContext to use internal helpers

Don't do the whole `this.#scopes.get(asyncHooks.executionAsyncId())` if
we can simply call `this.active()`.

## Check if span is active when set as active or root

Use the `setRoot` function to set the root span. Make sure that when the
root span is set, it's still open. If it's not open, don't set it.
Setting a closed root span is a worse experience and may break things,
because it introduces a broken state.

## Check if span is open in ScopeManager.withContext

Before passing through the given span as the current context, make
sure it's still open. If not, do not set it and reuse the existing
active span.

If no span is active, return NoopSpan. Returning a NoopSpan seems like
better behavior than returning a closed span where people can't tell if
it does anything or not.

I've also added a private `setActive` helper that does the same thing as
rootSpan but for active spans. That way it's more centralized how we set
the active and root spans.

### DIFF
--- a/packages/nodejs/src/scope.ts
+++ b/packages/nodejs/src/scope.ts
@@ -13,6 +13,7 @@ import { Span } from "./interfaces/span"
 import * as asyncHooks from "async_hooks"
 import { EventEmitter } from "events"
 import shimmer from "shimmer"
+import { NoopSpan } from "./noops/span"
 
 // A list of well-known EventEmitter methods that add event listeners.
 const EVENT_EMITTER_METHODS: Array<keyof EventEmitter> = [
@@ -111,11 +112,24 @@ export class ScopeManager {
     return this
   }
 
+  private setActive(span: Span) {
+    if (span.open) {
+      const uid = asyncHooks.executionAsyncId()
+      this.#scopes.set(uid, span)
+    }
+  }
+
   /**
    * Returns the current active `Span`.
    */
   public active(): Span | undefined {
-    const uid = asyncHooks.executionAsyncId()
+    return this.activeFor(asyncHooks.executionAsyncId())
+  }
+
+  /*
+   * Returns the current active `Span` for the given executionAsyncId.
+   */
+  private activeFor(uid: number): Span | undefined {
     const span = this.#scopes.get(uid)
     // Perform check if the span is not closed. A span that has been closed
     // can't be considered an active span anymore.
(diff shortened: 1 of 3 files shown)


luismiramirez (approved):

packages/nodejs/src/scope.ts
```diff
@@ -186,8 +212,12 @@ export class ScopeManager {
...
         this.removeSpanForUid(uid)
       } else {
-        this.#scopes.set(uid, oldScope)
-        this.#roots.set(uid, rootSpan)
+        if (rootSpan) {
+          this.setRoot(rootSpan)
+        }
+        if (oldScope) {
+          this.setActive(oldScope)
+        }
```
unflxw, line 220: Note that, since `.setRoot` and `.setActive` are used, and they only set spans if they're not closed, the old root and active spans will not be restored if they've been closed at this point in the code, meaning the span passed to `.withContext` will continue to be the active span after the `.withContext` callback has finished executing.

I don't think this is the desired behaviour. A closer-to-desired behaviour might be to also call `.removeSpanForUid` if either the root span or active span are closed.

Also note that, if the root span has been closed but the active span hasn't been closed, or the other way around, it will restore one but not the other, leading to a situation where the current root span may not be the root of the current active span. I don't think this is the desired behaviour either, I think that it should either restore both or remove both.

I think a possible implementation for the `finally` block that avoids these issues could be:

```typescript
finally {
  this.removeSpanForUid(uid)
  if (rootSpan && rootSpan.open && oldScope && oldScope.open) {
    this.setRoot(rootSpan)
    this.setActive(oldScope)
  }
}
```

Not blocking the PR on this, because there's other ways to end up in similarly inconsistent states using `.withContext`. It's kind of inevitable, at least with its current signature, as it takes an active span as an argument, but not a root span, and spans don't know their root spans.

But I think it would be best to avoid creating additional situations by which users of `.withContext` (or `tracer.withSpan` by proxy) can accidentally end up with an inconsistent state.

unflxw (approved):
I really like how this pull request re-arranges the ScopeManager's internals, not only so that it has more reasonable behaviour with closed spans, but so that it's easier to follow.

Also, tests. 🥳

packages/nodejs/src/scope.ts
tombruijn: I'm wondering if the root span is even something withContext should concern itself. Its behavior is to only overwrite the active span, why should it concern itself with the root span? If the root span gets closed during this time, it's not the withContext block responsibility to restore to a working state.

PR #678, opened 2022-06-01
Fix removing wrapped event listeners

unflxw:
### [Fix removing listeners in wrapped event emitters](https://github.com/appsignal/appsignal-nodejs/commit/2b812750a0a893f4d18cb8f3f8a9d8c6a30030e2)

The `.emitWithContext` function in the `ScopeManager`, which is called from the `.wrapEmitter` function in the `Tracer`, modifies a given `EventEmitter` so that the callbacks passed to it when registering a listener are bound to the active span of the asynchronous context in which the listener was registered. That is, calling `scopeManager.active()` or `tracer.currentSpan()` from the listener should return the span that was active at the point where the listener was registered. This is done by wrapping the callback in a function that closes over the current span, calling the original callback from within `scopeManager.withContext` with the closed over span, and registering this wrapper function as the callback for the event emitter's listener.

Since the callback that is registered for an event emitter's listener is a wrapper function over the callback, and not the original callback, future attempts to remove the listener from the event emitter silently fail, as the callback function that is passed to `.removeListener` is not the callback function that was originally registered.

This commit fixes this by keeping track of which wrapper functions were created for which events and callbacks, and modifying the given event emitter further, intercepting calls to methods that remove listeners and replacing the given callback with its previously created wrapper.

This commit also adds tests for the .emitWithContext function.

### [Copy original function name to wrapper](https://github.com/appsignal/appsignal-nodejs/commit/37e162c0cf065856262110b3ed0516453fe5e722)

When wrapping a function in `.bindContext`, the `length` property of the original function is copied over to the wrapper function.

This commit changes it so that, in addition to the `length` property, the `name` property is also copied.

Closes appsignal/support#173, for real this time.

### DIFF
--- a/packages/nodejs/src/scope.ts
+++ b/packages/nodejs/src/scope.ts
@@ -227,13 +227,15 @@ export class ScopeManager {
     // prevent re-wrapping
     contextWrapper[WRAPPED] = true
 
-    // explicitly inherit the original function's length, because it is
-    // otherwise zero-ed out
+    // explicitly inherit the original function's length and name
     Object.defineProperty(contextWrapper, "length", {
-      enumerable: false,
       configurable: true,
-      writable: false,
       value: fn.length
+    })
+
+    Object.defineProperty(contextWrapper, "name", {
+      configurable: true,
+      value: fn.name
     })
 
     return contextWrapper
--- a/packages/nodejs/src/scope.ts
+++ b/packages/nodejs/src/scope.ts
@@ -15,12 +15,18 @@ import { EventEmitter } from "events"
 import shimmer from "shimmer"
 
 // A list of well-known EventEmitter methods that add event listeners.
-const EVENT_EMITTER_METHODS: Array<keyof EventEmitter> = [
+const EVENT_EMITTER_ADD_METHODS: Array<keyof EventEmitter> = [
   "addListener",
   "on",
   "once",
   "prependListener",
   "prependOnceListener"
+]
+
+// A list of well-known EventEmitter methods that remove event listeners.
+const EVENT_EMITTER_REMOVE_METHODS: Array<keyof EventEmitter> = [
+  "off",
+  "removeListener"
 ]
 
 const WRAPPED = Symbol("@appsignal/nodejs:WRAPPED")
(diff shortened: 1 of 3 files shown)


luismiramirez (approved):

packages/nodejs/src/scope.ts
```diff
@@ -221,30 +227,85 @@ export class ScopeManager {
...
+      callbacks = []
+      eventMap.set(cb, callbacks)
+    }
+
+    callbacks.push(boundCallback)
+  }
+
+  pop(event: string, cb: Func<void>): Func<void> | undefined {
+    const eventMap = this.get(event)
+    if (!eventMap) return undefined
```
tombruijn, line 304:
```suggestion
    if (!eventMap) return
```

Nitpick: undefined is the default when not returning anything

packages/nodejs/src/scope.ts
```diff
@@ -221,30 +227,85 @@ export class ScopeManager {
...
+
+    const callbacks = eventMap.get(cb)
+    if (!callbacks) return undefined
```
tombruijn, line 307:
```suggestion
    if (!callbacks) return
```

Nitpick: undefined is the default when not returning anything

packages/nodejs/src/scope.ts
```diff
@@ -221,30 +227,85 @@ export class ScopeManager {
...
-    // otherwise zero-ed out
+    // explicitly inherit the original function's length and name
     Object.defineProperty(contextWrapper, "length", {
-      enumerable: false,
       configurable: true,
-      writable: false,
       value: fn.length
     })
 
+    Object.defineProperty(contextWrapper, "name", {
```
tombruijn, line 236: What problem does copying over the name property solve?
Is this something we can test for?

packages/nodejs/src/scope.ts
```diff
@@ -221,30 +227,85 @@ export class ScopeManager {
...
+            const maybeBoundCallback = boundCallbacks.pop(event, cb) ?? cb
+            return oldFn.call(this, event, maybeBoundCallback)
+          }
+        })
+      }
+    })
+  }
+}
+
+class BoundCallbacks extends Map<string, WeakMap<Func<void>, Func<void>[]>> {
```
tombruijn, line 285: A bit of a nitpick: With this class extending Map it also exposes a lot of other public functions, but it looks like only `push` and `pop` should really be used. Functions that aren't part of the Map API, e.g. `new BoundCallbacks().get("foo")`. Should we instead create a new class that wraps around the `Map` and only exposes the `push` and `pop` functions?

```class
class BoundCallbacks {
  #map: Map<...>

  push(...) { this.#map.set(...) }
  pop(...) { this.#map.get(...) }
}
```

packages/nodejs/src/scope.ts
unflxw: There definitely should be a test!

It doesn't really directly solve a problem, but if someone's debugging their application with good ol' `console.log`, the wrapper function will have the same name as the wrapped function, instead of being an unnamed function. I'm happy to remove this commit, it's a minor improvement and it's not really related to the actual fix. I just spent a lot of time staring at various `console.log` statements while trying to figure out the fix.

packages/nodejs/src/scope.ts
unflxw: Thanks! You're right, it shouldn't really expose methods other than `push` and `pop`.

tombruijn (approved):

PR #679, opened 2022-06-02
Remove Node 17 from the build matrix

unflxw:
### [Remove Node 17 from the build matrix](https://github.com/appsignal/appsignal-nodejs/commit/54282ee3f7add65d233bfbb9dc8a3b7265339f2e)

Node 17 has reached its end-of-life, so we don't want to support it anymore. This commit removes Node 17 builds from the Semaphore build matrix.

### [Remove OpenSSL workaround for Next.js](https://github.com/appsignal/appsignal-nodejs/commit/78cee208a68b385eea60b8a79b8cc7ee9bfb8ae2)

When Node.js version 17 was released, it included an upgrade to OpenSSL 3.0.0. This version of OpenSSL did not support some cyphers that were commonly used in Webpack workflows to hash file contents.

A workaround was added then to force Node.js to use an older OpenSSL version. This means we've been running all our Node.js tests with an uncommon configuration setting since then.

This commit removes the workaround. Instead, it adds a mechanism to the build matrix that allows certain variations of Node.js packages under test to be excluded from certain Node.js versions, and it uses it so that the Next.js versions whose bundled Webpack configurations use this cypher are not tested on Node 18.

[skip changeset]


tombruijn:
> This commit removes the workaround. Instead, it adds a mechanism to the build matrix that allows certain variations of Node.js packages under test to be excluded from certain Node.js versions, and it uses it so that the Next.js versions whose bundled Webpack configurations use this cypher are not tested on Node 18.

But we should test Next.js on Node.js 18, right? Can we apply the workaround only for Next.js on Node.js 18? Is the workaround still necessary on the latest webpack?

tombruijn (approved):

luismiramirez (approved):

PR #681, opened 2022-06-03
Check if spans are open in the ScopeManager before setting as root or active

tombruijn:
Previous iteration of this PR was PR #676. Closed in favor of this more incremental change per commit with feedback addressed from that PR.

## Add assertion for storing error in withContext

The ScopeManager.withContext tests did not assert if the error that was
rethrown was also set on the root span or not.

## Reuse scope helpers to fetch active and root span

Do not reimplement fetching the root and active spans again in the
`withContext` function.

## Check if span is active when set as root

Use the `setRoot` function to set the root span. Make sure that when the
root span is set, it's still open. If it's not open, don't set it.
Setting a closed root span is a worse experience and may break things,
because it introduces a broken state.

## Don't set closed spans as active in withContext

When a span is given to `ScopeManager.withContext` check if it is is
open. If it's not don't set it as active. Instead pass in the already
currently active span, or a NoopSpan if no span is active.

This prevents the usage of closed spans, that cannot be used to store
any data after they are closed.

## Don't restore the root span for withContext

The ScopeManager.withContext only temporary sets the current active
span. It then restores the active span afterwards.

But it also restores the root span, even though that should not have
changed. At least not by anything we've done in this withContext
function.

If the user has set another rootSpan, that is their decision, but it's
not up to us to restore it.

## Don't restore active span if closed in withContext

When a span gets closed for whatever reason before getting restored in a
withContext callback, do not restore it. We do not restore closed spans
as active spans.

This is difficult to test, because the `active()` function also does
this check, but removing that check this test will still pass.

## Refactor ScopeManager to use internal helpers

Use internal helpers to set and unset the active span so that it doesn't
repeat this behavior in multiple places.

### DIFF
--- a/packages/nodejs/.changesets/do-not-restore-closed-spans-from-within-withspan.md
+++ b/packages/nodejs/.changesets/do-not-restore-closed-spans-from-within-withspan.md
@@ -1,0 +1,6 @@
+---
+bump: "patch"
+type: "change"
+---
+
+Do not restore closed spans from within the `withSpan` helper. If a previously active span gets closed while `withSpan` has another span as currently active, do not restore the closed span when the callback has finished.
--- a/packages/nodejs/.changesets/do-not-restore-root-span-after-withcontext-callback-has-finished.md
+++ b/packages/nodejs/.changesets/do-not-restore-root-span-after-withcontext-callback-has-finished.md
@@ -1,0 +1,16 @@
+---
+bump: "patch"
+type: "change"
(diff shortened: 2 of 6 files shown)


unflxw (approved):
Nice! So now `withSpan` only switches the active span around, and the only thing it does with the root span is report errors. Neat 👍

luismiramirez (approved):

PR #683, opened 2022-06-07
Add OpenTelemetry wrapper client

tombruijn:
Hey team,

This is my alternative to what we discussed in the meeting this morning. I wasn't sure if it was understood what I meant so I figured I'd make the change and show it. Let's discuss the options that we have some time. I'd still like to see the alternative with another package.

## Add OpenTelemetry wrapper client

Add a client that disables the auto instrumentation when using this
package for OpenTelemetry. This inherits from the BaseClient but
provides the default instrumentation necessary for better OpenTelemetry
compatibility.

For the user this means they no longer need to configure the
`instrumentHttp`, `instrumentRedis` and `instrumentPg` options manually
anymore. We can update the docs to no longer mention them either for the
OpenTelemetry config. We do need to update the docs to use the new
OpenTelemetry client instead.

```js
// appsignal.js
const { AppsignalOpenTelemetry } = require("@appsignal/nodejs");

const appsignal = new AppsignalOpenTelemetry({
  // Your config
});

module.exports = { appsignal };
```

## Add tests for initCorePlugins in client test file

The client test did not test if the core plugins were called or not, so
I've added tests for it.

### DIFF
--- a/packages/nodejs/.changesets/add-opentelemetry-client-wrapper-for-easy-setup.md
+++ b/packages/nodejs/.changesets/add-opentelemetry-client-wrapper-for-easy-setup.md
@@ -1,0 +1,25 @@
+---
+bump: "patch"
+type: "add"
+---
+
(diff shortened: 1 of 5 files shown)


thijsc (commented):
This makes sense to me too, I will hold of forming an opinion until we see the package alternative too.

PR #689, opened 2022-06-14
Convert to OpenTelemetry package

### LINKED ISSUE #599: Spans created by `Span.child()` and `Tracer.createSpan()` are not tracked in the scope as the currentSpan
If you create a span using `span.child()` and then later call `tracer.currentSpan()`, these spans will not be the same.

```js
const tracer = appsignal.tracer()
const span = tracer.currentSpan()
console.log(tracer.currentSpan())
// RootSpan { _ref: [External: 17b70cd0] }
const child1 = span.child()
console.log(tracer.currentSpan())
// RootSpan { _ref: [External: 17b70cd0] }
console.log(child1)
// ChildSpan { _ref: [External: 179c4a00] }
const child2 = tracer.createSpan()
console.log(tracer.currentSpan())
// RootSpan { _ref: [External: e282200] }
console.log(child2)
// ChildSpan { _ref: [External: e2829a0] }
```

Basically, nesting of spans, relying on `tracer.currentSpan()`, is not possible if we create  with `span.child()` etc. We should wrap them in `tracer.withSpan` type of functions. The spans will only be nested under the root span or last span wrapped with `tracer.withSpan`.

tombruijn:
See commits
See https://github.com/appsignal/appsignal-nodejs/pull/688 PR description for things to do.

## Closes issues

- Closes #599
- Closes https://github.com/appsignal/appsignal-docs/issues/537

### DIFF
--- a/scripts/extension/support/constants.js
+++ b/scripts/extension/support/constants.js
@@ -3,7 +3,7 @@
 // appsignal-agent repository.
 // Modifications to this file will be overwritten with the next agent release.
 
-const AGENT_VERSION = "06391fb"
+const AGENT_VERSION = "8dd80e5"
 const MIRRORS = [
   "https://appsignal-agent-releases.global.ssl.fastly.net",
   "https://d135dj0rjqvssy.cloudfront.net"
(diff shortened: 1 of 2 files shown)


tombruijn:
Moving this to blocked as we're not actively working on this PR specifically.

PR #691, opened 2022-06-15
Add demo function to restore demo functionality

tombruijn:
This was previously removed in the OpenTelemetry only step.

I've used the new `createOpenTelemetrySpan` function to create the
spans, link them together with generated ids.
If I hardcode the ids you can see spans from child span all being
reported under the same root span and only one root span becomes visible
in the UI.

The `randomBytes` function has been available for a very long time in
Node.js. I would want to use `randomUUID`, but that's only been
available in Node.js 14.17, and not 14.0.

[skip changeset]

### DIFF
--- a/src/demo.ts
+++ b/src/demo.ts
@@ -1,0 +1,67 @@
+import { randomBytes } from "crypto"
+import { Client } from "./client"
+import { hrTime } from "./utils"
+
+/**
+ * Sends a demonstration/test sample for a exception and a performance issue.
+ *
+ * Note that the agent must be active for at least 60 seconds in order for the payload
+ * to be sent to AppSignal.
+ */
+export function demo(client: Client) {
+  const { sec: startSec, nsec: startNsec } = hrTime()
+  const traceId = randomBytes(10).toString("base64")
+  const rootSpanId = randomBytes(10).toString("base64")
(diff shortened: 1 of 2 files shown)


jeffkreeftmeijer (approved):

src/demo.ts
```diff
@@ -0,0 +1,70 @@
...
+import { hrTime } from "./utils"
+
+/**
+ * Sends a demonstration/test sample for a exception and a performance issue.
+ *
+ * Note that the agent must be active for at least 60 seconds in order for the payload
+ * to be sent to AppSignal.
+ *
+ * This is also a good place to start if you want to see what custom instrumentation
+ * looks like in the wild!
```
unflxw, line 12: I'm not sure this comment is true anymore? We're using a function here that isn't meant to be used outside of the span processor.

We haven't really discussed what custom instrumentation post-OpenTelemetry is going to look like yet, but I don't think we'll have users call `extension.createOpenTelemetrySpan` directly. I think we've mentioned the possibility of exposing the OT tracer.

unflxw (approved):

src/demo.ts
tombruijn: I don't think it was ever true. I just copy-pasted it. I will remove it.
I don't know what we're going to do with the tracer.

PR #705, opened 2022-06-30
Test the SpanProcessor's error handling behaviour

unflxw:
I'm doing this work as part of https://github.com/appsignal/opentelemetry/issues/15. This PR tests how the existing implementation of the SpanProcessor in the `opentelemetry` branch transforms OpenTelemetry exception events into AppSignal span errors.

[skip changeset]

### [Add back span.toObject for testing](https://github.com/appsignal/appsignal-nodejs/commit/e039e1c779eb353e62c252abf7b85e0e03fe7242)

This commit adds back `span.toObject` and `spanToJSON`, as they are
currently implemented in the `main` branch, to the `Span` object in
the `opentelemetry` branch.

### [Make the ref property in Span private](https://github.com/appsignal/appsignal-nodejs/commit/c9b2fad1f860938af194011b8cab52fc8fd89dfe)

This property is not meant to be accessed from outside the Span,
so it should be private.

### [Add back and adapt SpanTestRegistry](https://github.com/appsignal/appsignal-nodejs/pull/705/commits/ca2648d824e94be595f0032879669a327925f996)

This commit adds back the `SpanTestRegistry` used by the HTTP
instrumentation tests in the `main` branch, adapting it to be used
with the `Span` objects in the `opentelemetry` branch.

### [Add tests for the span processor](https://github.com/appsignal/appsignal-nodejs/commit/28b129891b337fbabe9a546176ffe84a7ab2191c)

Tests the span processor's current behaviour using the
`SpanTestRegistry`, checking that exception events in OpenTelemetry
spans are transformed into errors in AppSignal spans.

[http]: https://github.com/appsignal/appsignal-nodejs/blob/8c893a4fc6134fd912adbd61900416913fa14b68/packages/nodejs/test/registry.ts

### DIFF
--- a/src/__tests__/span_processor.test.ts
+++ b/src/__tests__/span_processor.test.ts
@@ -1,0 +1,122 @@
+import { SpanTestRegistry } from "../../test/registry"
+import { Client } from "../client"
+import { SpanProcessor } from "../span_processor"
+import { Tracer, BasicTracerProvider } from "@opentelemetry/sdk-trace-base"
+import { trace, context } from "@opentelemetry/api"
+
+describe("Span processor", () => {
+  let tracer: Tracer
+  let tracerProvider: BasicTracerProvider
+  let client: Client
+
+  function createError(message: string): any {
+    try {
+      throw new Error(message)
+    } catch (e) {
+      return e
+    }
+  }
+
+  beforeEach(() => {
+    SpanTestRegistry.clear()
+
+    client = new Client({
+      name: "TEST APP",
+      pushApiKey: "PUSH_API_KEY",
+      active: true,
+      enableMinutelyProbes: false
+    })
+
+    const spanProcessor = new SpanProcessor(client)
+
+    tracerProvider = new BasicTracerProvider()
+    tracerProvider.addSpanProcessor(spanProcessor)
+
+    tracer = tracerProvider.getTracer("unknown-instrumentation")
+  })
+
+  afterEach(() => {
+    tracerProvider.shutdown()
+    client.stop()
+  })
+
+  it("processes unknown OpenTelemetry spans", () => {
+    tracer.startSpan("unknownSpan").setAttribute("foo", "bar").end()
+
+    expect(SpanTestRegistry.count()).toEqual(1)
+    expect(SpanTestRegistry.lastSpan()?.toObject()).toMatchObject({
+      name: "unknownSpan",
+      closed: true,
+      attributes: expect.objectContaining({
+        foo: "bar",
+        "appsignal:body": expect.stringContaining("unknown-instrumentation")
+      }),
+      parent_span_id: "",
+      error: null
+    })
+  })
+
+  it("processes an exception event in an OpenTelemetry span as the error", () => {
+    const span = tracer.startSpan("errorSpan")
+    span.recordException(createError("first"))
+    span.end()
+
+    expect(SpanTestRegistry.count()).toEqual(1)
+    expect(SpanTestRegistry.lastSpan()?.toObject()).toMatchObject({
+      name: "errorSpan",
+      closed: true,
+      parent_span_id: "",
+      error: {
+        name: "Error",
+        message: "first",
+        backtrace: expect.any(String)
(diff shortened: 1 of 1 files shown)


src/span.ts
```diff
@@ -1,22 +1,52 @@
...
+   * Returns a SpanData object representing the internal Span in the extension.
+   *
+   * @private
+   */
+  public toObject(): SpanData {
+    const json = span.spanToJSON(this.#ref)
+
+    // if this is true, then the span has been garbage collected
+    // @TODO: i feel that this could have better ergonomics on the agent
+    // side. come up with something better here later.
```
tombruijn, line 45:
```suggestion
    // If the span JSON is empty, the span has been closed.
```

I know it's an old comment, but let's update it.
We're not going to change it returning an empty string any time soon I think.

src/__tests__/span_processor.test.ts
```diff
@@ -0,0 +1,157 @@
...
+    })
+
+    const backtrace = JSON.parse(spans[0].error.backtrace)
+    expect(backtrace).toBeInstanceOf(Array)
+
+    expect(backtrace[0]).toEqual("Error: first")
+    expect(backtrace[1]).toMatch(/^\s+at createError \(/)
+  })
+
+  it("ignores exception events in child spans", async () => {
```
tombruijn, line 121: Didn't we say the last error would be stored on the span instead of the first? Or does OpenTelemetry work different in this area of error handling?

src/__tests__/span_processor.test.ts
```diff
@@ -0,0 +1,157 @@
...
+    jest
+      .spyOn(client.extension, "createOpenTelemetrySpan")
+      .mockImplementation((...args) => {
+        const span = originalCreateOpenTelemetrySpan(...args)
+        mockSpan(span)
+        return span
+      })
+  }
+
+  function mockSpan(span: Span) {
```
tombruijn, line 25: What did you learn writing this about how this solution is better? Did the span registry you referenced have gaps anywhere?

What I like about the span registry is that we're able to load it into other test files as well. In case we ever want to assert span values in any other test files.

src/__tests__/span_processor.test.ts
unflxw: I thought so too! I've written the test for what the currently existing behavior is.

Looking at it now, the test is kind of a lie: all errors _are_ being saved to the span, but the test can only assert the last one due to [a limitation of the `SpanInProgress.to_json()` method on the agent](https://github.com/appsignal/appsignal-agent/blob/d573c9b59bfadb546354cd995827e87e74d190ef/extension/src/tracing/span.rs#L440-L447) used for the assertions, which will only output the first error.

I'll look for a way to address this.

luismiramirez (approved):

src/__tests__/span_processor.test.ts
tombruijn: The advantage of replacing the class implementation (and by replacing we mean extending it with some additional testing behavior), is that we don't need to spy on a specific instance of the AppSignal client/extension. It works without spies! What you say is true, spans are only ever created by the `createOpenTelemetrySpan` function on the extension, but it's also never created without a class instance of `Span`. Which I why opted for the Span class replacement solution.

As [I said in my commit back then](https://github.com/appsignal/appsignal-nodejs/commit/86b42af3704a9ded22365e706ba65667a3ea3600), I was inspired by what I wrote for the Ruby gem, but I also see that even that is not a perfect implementation and a test mode in the extension would be even better. Find a way for `spanToJson` to return the attributes even when the span is closed rather than hack our way around it.
I remember seeing a spec in the OpenTelemetry package that they assert what's being sent through a (HTTP) exporter, which I don't know if we could do reliably in every test setup.

unflxw: I'll grab the SpanTestRegistry from the main branch and use it on these tests, modifying it to work with the new Span class. I liked the original implementation and it should be easy to adapt it for this use case.

tombruijn (approved):

PR #709, opened 2022-07-14
Configure OpenTelemetry HTTP protobuf exporter

tombruijn:
Send data directly to the agent rather than through the extension using
the OpenTelemetry HTTP protobuf exporter.


tombruijn:
Closing because Thijs doesn't want to go this route for Node.js at the moment.

PR #715, opened 2022-08-03
Add Express error handler middleware for OTel

### LINKED ISSUE #713: Error tracking middleware for Express
Create and export a middleware for OpenTelemetry error tracking similar to the current one so users can use it on their Express apps. It'll need to receive the AppSignal client object to access OpenTelemetry's tracer to handle spans and record the exception if any.

[skip blogpost]

luismiramirez:
This Express error handler works similarly to what we already have in
the current integration. It allows users to automatically record
exceptions raising in web request to the current active span.

[skip blogpost]
[skip changeset]

Closes #713

---

Tests will be added in the coming integration tests PR


src/instrumentation/express/error_handler.ts
```diff
@@ -0,0 +1,31 @@
...
+      activeSpan &&
+      err &&
+      (!err.status || (err.status && err.status >= 500))
+    ) {
+      if (activeSpan) {
+        activeSpan.recordException(err)
+        activeSpan.setStatus({
+          code: opentelemetry.SpanStatusCode.ERROR,
+          message: err.message
+        })
```
unflxw, line 26: Discussed this with @luismiramirez and we don't think it's a problem in practice, but just mentioning it for reference: in practice, the active span will be the HTTP span, and this status that is being set here will be overwritten with a `SpanStatusCode.OK` when the request is closed by the HTTP integration.

In practice, this doesn't matter to us, because we don't do anything with the span status. We just care about the exception event that is recorded. This is how you're supposed to record an exception, by recording the exception and setting the status -- if the status is later overriden by the instrumentation, that's out of our control.

src/instrumentation/express/error_handler.ts
```diff
@@ -0,0 +1,31 @@
...
+  ) {
+    const activeSpan = opentelemetry.trace.getSpan(
+      opentelemetry.context.active()
+    )
+    // if there's no `status` property, forward the error
+    // we also ignore client errors here
+    if (
+      activeSpan &&
+      err &&
+      (!err.status || (err.status && err.status >= 500))
```
unflxw, line 19: Curious: who sets this `status` property on the Error object? Is this an Express thing? Can't find it on their docs.

unflxw (approved):

src/instrumentation/express/error_handler.ts
luismiramirez: Yes! It's "mentioned" here: https://expressjs.com/en/guide/error-handling.html#the-default-error-handler

_The res.statusCode is set from err.status (or err.statusCode). If this value is outside the 4xx or 5xx range, it will be set to 500._

`status` is a Express alias for `statusCode` that is added by node to HTTP errors.

That's why we "enrich" the type with: `err: Error & { status?: number }`

unflxw: Huh. Thanks! Should we also be looking at `statusCode` then?

luismiramirez: I don't think so, we should be safe using `status` as Express docs recommend. If we were outside of Express, yes, we should be looking at `statusCode`

jeffkreeftmeijer (approved):

src/instrumentation/express/error_handler.ts
tombruijn: I thought you were going to create an additional child span for it. That would be useful if we later have an event timeline where you can see which error occurred where, if you have multiple errors.

luismiramirez: Recent processor changes don't support that behavior yet. This middleware doesn't know where an error happened because the active span is always the root span, so we don't know what middleware it comes from. Also, this middleware will catch the first HTTP error, and then the request will break, so it's unlikely that more than one error raises within the same trace.

unflxw: Luismi and I have discussed a very un-cooked, future nice-to-have idea on how to make the error appear on the span that raised it.

The gist of it is, instead of having this error middleware, having functionality inside the Express instrumentation that intercepts calls to `next(err)` in middlewares and request handlers. From a cursory look at the Express internals, it seems that `next` is always called with the error, both when an error is explicitly passed to `next` in the request handler, and implicitly in the Express internals when an error is thrown in the handler.

But for now, I think the error middleware will do. Problem with it is, of course, that it can't know where the error came from in the middleware chain.

PR #717, opened 2022-08-09
Otel integration test apps

luismiramirez:
This PR contains a proposal of an integration test architecture for all the supported instrumentations. If approved, it'll be our standard and will open a new PR for each instrumentation combination we already have. After that, we'll add Postgres and GraphQL instrumentations.

---
## [Remove old integration test helpers](https://github.com/appsignal/appsignal-nodejs/commit/5fc3b4020c8ad8fb354215bf1824c1bf4eede618)

Old `test/integration/support` are no longer needed for the
OpenTelemetry version of the integration.

## [Add testMode and testModeFilePath config opts](https://github.com/appsignal/appsignal-nodejs/commit/298a4a164167f0bb2ad6d2d62d7fc22f9f07d0b0)

When in test mode (`testMode:true`) the `TestModeSpanProcessor` is added
to the tracer provider and it saves spans in a JSON file (path based on
the contents of `testModeFilePath`) for the integration tests to read.

## [Add Express + Redis test app](https://github.com/appsignal/appsignal-nodejs/commit/d5bd6ba37f1553e7de48183cb226a21d598438bf)

This commit adds a test app using Express and Redis plus the structure
future test apps will have.

It includes a module with common methods that will be shared in all the
apps.

To run the tests position in `test/express-redis/` and run
`docker-compose up`.

## [Add Express Redis integration tests to CI](https://github.com/appsignal/appsignal-nodejs/commit/24fe23404a3e45343c145c9f11aa8500f6aa375d)

This commit adds a script to run integration test apps and the addition
of the Express Redis test app to the Semaphore build matrix.

## [Comply with linter suggestions](https://github.com/appsignal/appsignal-nodejs/commit/e486e2ff3cf291d4e6568eecf3c780973f3cc24e)

Due to the nature of the PR, some minor linter offenses needed to be
addressed.

## [Stop integration tests on container exit](https://github.com/appsignal/appsignal-nodejs/commit/50c5bbca61a409c8c20728c7b024ed44e14ec44a)

Integration tests are run in their own container so docker-compose needs
to know when to exit to tell the CI the result code.

This commit adds the flags --abort-on-container-exit to exit with code >
0 when something goes wrong and --exit-code-from tests to base the

[skip changeset]


test/integration_helper.rb
```diff
@@ -0,0 +1,114 @@
...
+    while child && child["parentSpanId"]
+      return true if child["parentSpanId"] == parent["spanId"]
+
+      child = span(child["parentSpanId"])
+    end
+
+    false
+  end
+
+  def expect_http_root_span(name)
```
luismiramirez, line 62: I'll keep all these instrumentation-specific helpers here until we have all the apps in place, because I can see all of them being reused. Even the Redis ones.

src/config/options.ts
```diff
@@ -30,6 +30,8 @@ export type AppsignalOptions = {
   runningInContainer: boolean
   sendParams: boolean
   skipSessionData: boolean
+  testMode: boolean
+  testModeFilePath: string
```
tombruijn, line 34: Other test specific config we've only added as env vars, like `APPSIGNAL_DIAGNOSE_ENDPOINT` (should have a `_` prefix) for pointing to the local endpoint in the test suite. I don't think this needs to be part of the config class because it doesn't need to be configurable by our users. You can use a `_APPSIGNAL_TEST_MODE` and `_APPSIGNAL_TEST_MODE_FILE_PATH` env var for this.

src/client.ts
```diff
@@ -199,6 +201,12 @@ export class Client {
 
     const tracerProvider = new NodeTracerProvider()
     tracerProvider.addSpanProcessor(new SpanProcessor(this))
+    if (this.config.data.testMode && this.config.data.testModeFilePath) {
+      const filePath = this.config.data.testModeFilePath
+      const spanProcessor = new TestModeSpanProcessor(filePath)
+
+      tracerProvider.addSpanProcessor(spanProcessor)
```
tombruijn, line 208: I was a bit unsure what was tested here, because this doesn't use our real SpanProcessor. It's testing all the automatically added instrumentation up to the SpanProcessor part, right?

Like I mentioned in an earlier call, we can also call `toObject()` on an OpenTelemetry Span in the SpanProcessor when this test mode is active and write that extension parsed result to a file.

test/express-redis/tests/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM ruby:3.0
+
+RUN bundle config --global frozen 1
+
+WORKDIR /usr/src/app
+
+COPY Gemfile Gemfile.lock ./
+RUN bundle install
+
+COPY . .
```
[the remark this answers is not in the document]
unflxw: @tombruijn There is a point to copying the Gemfile and its lock first, then the rest of the application. It allows the Docker cache to reuse the `RUN bundle install` layer if the Gemfile and its lock haven't changed.

src/config/options.ts
unflxw: If it's not declared on the config type, then TS won't let you access it from `this.config.data`. I don't think our config object currently supports having configuration options that can't be set via the constructor.

Regarding the `_` prefix at the beginning of the environment variable, I thought that was used only for variables that are passed on to the agent.

We could, of course, just read `process.env["APPSIGNAL_TEST_MODE"]` where needed, instead of getting it from the config object, if that's preferable.

test/express-redis/app/run.sh
```diff
@@ -0,0 +1,26 @@
...
+)
+
+cd /app
+
+echo "Installing app dependencies"
+rm -rf node_modules
+npm install
+
+echo "Linking integration"
+npm link @appsignal/nodejs
```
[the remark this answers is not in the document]
luismiramirez: I think it's best to remove the `file://..` and keep the link.

test/express-redis/tests/Dockerfile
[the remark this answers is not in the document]
tombruijn: Ok, but that's only locally. On Semaphore it doesn't cache Docker layers.

If we use a volume we could cache the bundle directory on the CI and locally, so it doesn't need to download all gems all the time. We'd take small performance hit because it's mounted on a volume, but not as bad as on local macOS machines.

src/config/options.ts
tombruijn: > Regarding the _ prefix at the beginning of the environment variable, I thought that was used only for variables that are passed on to the agent.

It's for private env vars, which include the extension's env vars.

> We could, of course, just read process.env["APPSIGNAL_TEST_MODE"] where needed, instead of getting it from the config object, if that's preferable.

I would prefer it, just because it's not a config option any of our users need to be able to set. (That they can potentially via the env var is a minimal risk.)

src/client.ts
luismiramirez: I'll take some time to try the `toObject()` approach.

tombruijn: No need to block the PR on it. You can look at it in another PR if you want.

src/instrumentation/express/error_handler.ts
```diff
@@ -0,0 +1,31 @@
+import { Request, Response, NextFunction, ErrorRequestHandler } from "express"
```
tombruijn, line 1: This branch branched of the OpenTelemetry branch before #715 was merged so this commit shouldn't be here. Can you rebase on the latest OpenTelemetry branch state?

tombruijn (approved):
other than that one commit that shouldn't be in this PR, looks good

package.json
```diff
@@ -18,11 +18,11 @@
     "@opentelemetry/sdk-trace-base": "^1.4.0",
     "node-addon-api": "^3.1.0",
     "node-gyp": "^9.0.0",
+    "npm": "^7.18.1",
```
tombruijn, line 21: wait why is npm itself a dependency here? I see this directly after hitting approve.

Can you remove it?

src/client.ts
luismiramirez: As talked earlier this week, I think we should keep this out of scope as the plans are relying on the upcoming HTTP server in the agent and get the spans from there.

unflxw (approved):

PR #718, opened 2022-08-09
Do not instrument Koa generator middleware

unflxw:
In Koa versions 0.x and 1.x, it was possible to implement middleware using generator functions. In Koa version 2.x, these are referred as "legacy middleware" and their use is deprecated, emitting a warning when they are used. However, it is still possible to use them, as Koa uses an utility called `koa-convert` internally to transform legacy middleware into new-style middleware. In Koa version 3.x, support for generator middleware will be removed entirely.

Our instrumentation does not support generator-based middleware, and it will break the request if a generator-based middleware is used early in the middleware chain, by failing to yield to the following middleware. Unfortunately, low-quality tutorial websites continue to recommend using generator-based middleware for error handling, and these sites rank high for relevant keywords.

This commit changes our Koa instrumentation so that it does not attempt to instrument generator-based middleware. On Koa version 2.x, this allows the Koa internals to convert the middleware and emit the deprecation warning. On Koa version 3.x, this will outright fail, which is the best possible deprecation warning.

The docs are also updated to point out that generator-based middleware will not be instrumented, and it points users to `koa-convert`, which they can use to convert the middleware _before_ passing it to the Koa application's `.use` method, which would allow AppSignal to instrument it.

### DIFF
--- a/packages/koa/README.md
+++ b/packages/koa/README.md
@@ -50,6 +50,12 @@ app.on("error", (error) => {
 });
 ```
 
+Note that generator-based middleware was deprecated in Koa version 2.x, and the next major Koa version will remove support for them entirely.
+
+Our instrumentation does not instrument generator-based middleware. The [Koa 2.x migration guide][guide] explains how you can use the `koa-convert` library to convert them to new-style async middleware.
+
+[guide]: https://github.com/koajs/koa/blob/v2.x/docs/migration.md#using-v1x-middleware-in-v2x
+
 ## Contributing
 
 Thinking of contributing to this repo? Awesome! 🚀
--- a/packages/koa/.changesets/do-not-break-requests-using-legacy-generator-middleware.md
+++ b/packages/koa/.changesets/do-not-break-requests-using-legacy-generator-middleware.md
@@ -1,0 +1,6 @@
+---
+bump: "patch"
+type: "fix"
+---
+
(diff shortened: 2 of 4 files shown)


tombruijn (approved):

packages/koa/README.md
```diff
@@ -50,6 +50,12 @@ app.on("error", (error) => {
 });
 ```
 
+Note that generator-based middleware are deprecated in Koa version 2.x, and support for them will be removed in the next major Koa version.
+
+Our instrumentation does not instrument generator-based middleware. As described in [the Koa 2.x migration guide][guide], you can use the `koa-convert` library to convert them to new-style async middleware.
```
noccor, line 55: Looks good, some very minor changes:

Note that generator-based middleware was deprecated in Koa version 2.x, and the next major Koa version will remove support for them entirely. 

Our instrumentation does not instrument generator-based middleware. Our [Koa 2.x migration guide][guide] explains how you can use the `koa-convert` library to convert them to new-style async middleware.

tombruijn:
@unflxw I see this PR has no changesets, but I assume it's a change we do want to ship?

unflxw:
@tombruijn Ah, yes, thank you! I'll add one now. I wonder why the bot didn't get mad at me...

tombruijn:
I don't know. No errors occurred 🤔 But I guess the backlog-helper has a difficult time waking up this morning, because it posted the comment right after being redeployed.

unflxw:
@tombruijn Changeset added, if you wish to review it.

packages/koa/README.md
unflxw: Thank you! Applied with a minor change as discussed ("Our" -> "The")

PR #721, opened 2022-08-23
update readme: syntax error in config object

zt4ff:
Hi, I added a comma after a property in the appSignal configuration file.


tombruijn:
Thanks for the fix!

PR #729, opened 2022-09-15
Add Prisma support

luismiramirez:
This commit adds Prisma automatic instrumentation. It uses Prisma
official OpenTelemetry package and includes the usual integration test
app.

[skip changeset]

---

To run the test app, go to `test/express-prisma` and run `$ docker-compose up`

Followed official docs:

- OpenTelemetry: https://www.prisma.io/docs/concepts/components/prisma-client/opentelemetry-tracing
- Installation guide: https://www.prisma.io/docs/getting-started/setup-prisma/start-from-scratch/relational-databases-typescript-postgres


unflxw:
Looks good. It's not clear to me if this will require an extractor in the agent, or the existing Postgres and MySQL extractors suffice?

luismiramirez:
@unflxw the extractors are triggered by the instrumentation library name. In this case, it doesn't matter which database you're using, everything will be wrapped with Prisma spans, so at least we need to trigger the SQL extraction by checking if the span instrumentation library name is `prisma`. It's also important to check that the whole tree is displayed because there's some useful information there apart from the query.

Tree sample: https://github.com/appsignal/opentelemetry/issues/44#issuecomment-1248170833

tombruijn (approved):

jeffkreeftmeijer (approved):

unflxw (approved):

PR #734, opened 2022-09-27
Additional instrumentations config option

unflxw:
### [Require parent span for IORedis](https://github.com/appsignal/appsignal-nodejs/commit/89927c3151aba5c538cec2d599182faf553b478e)

This is a left-over from the 2.x OpenTelemetry instrumentation,
which required spans to be created even if there was no root span
(because the root span would be an AppSignal span)

### [Remove AppsignalOptions interface](https://github.com/appsignal/appsignal-nodejs/commit/8c70fd9a2c551eedd655e430dd3a9b3fe08a0e16)

This is a duplicate of the `AppsignalOptions` type found in
`config/options.ts`, except as an interface. It is not used and,
as far as I can tell, serves no purpose.

### [Add additionalInstrumentations config option](https://github.com/appsignal/appsignal-nodejs/commit/0da9ec81f9b1503625c6ca423acdf35bd19cb755)

This commit adds the `additionalInstrumentations` config option.
Since it doesn't conform to the rules of other config options, as
its value is not serialisable and cannot be set via environment
variables, it's defined on a separate object.

### Thoughts

Fixes appsignal/opentelemetry#45.

This is a "good enough" approach to the problem -- it lets you add your own instrumentations. The name is `additionalInstrumentations` to make it clear that it does not override the default instrumentation list, but merely adds to it.

In the future, we may want to allow for a way to modify the existing instrumentation list. It might be worth looking at how `@opentelemetry/auto-instrumentations-node` does it -- they provide a function that generates the instrumentation list, and allows you to pass modifications to the instrumentations' config to it:

```js
{
  instrumentations: [
    getNodeAutoInstrumentations({
      // load custom configuration for http instrumentation
      '@opentelemetry/instrumentation-http': {
        applyCustomAttributesOnSpan: (span) => {
          span.setAttribute('foo2', 'bar2');
        },
      },
      // disable express instrumentation entirely
      '@opentelemetry/instrumentation-express': {
        enabled: false
      }
    }),
  ]
}
```

[skip changeset]


src/client.ts
```diff
@@ -179,11 +181,13 @@ export class Client {
...
    */
-  private initOpenTelemetry() {
+  private initOpenTelemetry(
+    additionalInstrumentations?: OpenTelemetryOptions["additionalInstrumentations"]
+  ) {
     const sendParams = this.config.data.sendParams
     const sendSessionData = this.config.data.sendSessionData
     const sdk = new NodeSDK({
-      instrumentations: [
+      instrumentations: (additionalInstrumentations ?? []).concat([
```
tombruijn, line 190: I would turn this around, let `additionalInstrumentation` overwrite our defaults. That way, if users want to, they override the HttpInstrumentation or PrismaInstrumentation (and any others) with their own config. How it's currently implemented any conflicting additionalInstrumentation is silently ignored.

src/diagnose.ts
```diff
@@ -205,7 +205,9 @@ export class DiagnoseTool {
 
     Object.keys(options).forEach(key => {
       const newKey = JS_TO_RUBY_MAPPING[key]
-      config[newKey] = options[key]
+      if (newKey !== undefined) {
```
tombruijn, line 208: I guess this is because the OpenTelemetry config options aren't in our JS_TO_RUBY_MAPPING list?

tombruijn:
> In the future, we may want to allow for a way to modify the existing instrumentation list. It might be worth looking at how @opentelemetry/auto-instrumentations-node does it -- they provide a function that generates the instrumentation list, and allows you to pass modifications to the instrumentations' config to it:

Yes, I would want us to take a closer look at that solution. Either use the function that auto package offers, or make something similar ourselves.

unflxw:
@tombruijn Alright, closing this and implementing something similar to that instead. 👍

tombruijn:
@unflxw shouldn't the redis and options commits still be merged? Can you submit them in a PR of their own?

PR #739, opened 2022-09-28
Implement `instrumentations` config option

unflxw:
When initialising AppSignal, it is now possible to pass an `instrumentations` config option.

This option takes a map as an argument. The keys of the map are instrumentation package names.

To disable a default instrumentation, pass `{enabled: false}` as the value.

To change the configuration of a default instrumentation, pass the desired configuration as the value. The given configuration will be merged with the default configuration for that instrumentation.

To add a new instrumentation, pass the initialised instrumentation as the value.

```
  instrumentations: {
    // disable koa instrumentation
    "@opentelemetry/instrumentation-koa": {
      enabled: false
    },
    // change the config for the http instrumentation
    "@opentelemetry/instrumentation-http": {
      ignoreIncomingPaths: [/test/i]
    },
    // add a custom instrumentation
    "@someone/custom-instrumentation": new CustomInstrumentation()
  }
})
```

To provide an upgrade path for adding more default instrumentations in the future without breaking the application for users who already added the instrumentation on their own, if an instrumentation instance is given instead of a config hash, instead of crashing, the instance is used.

---

The type incantations are based on those from the `@opentelemetry/auto-instrumentations-node` package. The configuration merge is done for top-level keys only. The `{enabled: false}` thing comes from that package as well -- it may be familiar to existing OpenTelemetry users.

I wonder if this is making it too easy for users to shoot themselves in the foot. I'm not sure if we should aim to make it easy for users to change the configuration of the default instrumentations -- they're configured the way they are for a reason. I could see confused users wondering why setting `headersToSpanAttributes` breaks the `requestHeaders` config option, or why setting the Express `requestHook` breaks the params in AppSignal. This configuration option also takes too many possible argument shapes in order to do too many possible things, and has an unwieldy type signature.

Perhaps we could instead have two configuration options: `disableInstrumentations`, which takes an array of package names of default instrumentations to disable, and `addInstrumentations`, which takes an array of initialised instrumentation instances. If users _really_ want to change the config of a default instrumentation, they can disable it and re-add it with the new config. It would be simpler to implement on our end, and to reason about on the users' end.

Fixes appsignal/opentelemetry#45.

[skip changeset]


src/client.ts
```diff
@@ -28,9 +28,43 @@ import { RedisInstrumentation as Redis4Instrumentation } from "@opentelemetry/in
...
+
+type AppsignalInstrumentationOption = DefaultInstrumentationConfigMap & {
+  [key: string]: InstrumentationOption
+}
+
+export type AppsignalOpenTelemetryOptions = {
+  instrumentations: AppsignalInstrumentationOption
+}
+
+type Options = Partial<AppsignalOptions & AppsignalOpenTelemetryOptions>
```
unflxw, line 66:
```suggestion
type AppsignalOpenTelemetryOptions = {
  instrumentations: AppsignalInstrumentationOption
}

export type Options = Partial<AppsignalOptions & AppsignalOpenTelemetryOptions>
```

Whoops.

src/client.ts
```diff
@@ -178,73 +214,129 @@ export class Client {
...
+
+    const additionalInstrumentations = Object.entries(instrumentationsOption)
+      .filter(([key, instrumentation]) => {
+        const isDefault = defaultInstrumentationNames.includes(key)
+        const isInstance = instrumentation instanceof InstrumentationBase
+
+        return !isDefault || isInstance
+      })
+      .map(
+        ([_key, instrumentation]) => instrumentation as InstrumentationOption
```
tombruijn, line 326: > "@someone/custom-instrumentation": new CustomInstrumentation()

You gave the above example for additional instrumentation. For users the key in the instrumentation object won't matter right? They can give the key any name.

src/client.ts
```diff
@@ -178,73 +214,129 @@ export class Client {
...
+    const defaultInstrumentations = Object.entries(DefaultInstrumentationMap)
+      .filter(([key, _instrumentationConstructor]) => {
+        const instrumentationName =
+          key as keyof typeof DefaultInstrumentationMap
+
+        const userInstrumentationConfig =
+          instrumentationsOption[instrumentationName] || {}
+
+        if (userInstrumentationConfig instanceof InstrumentationBase) {
+          return false
```
tombruijn, line 294: This means it gets filtered out if it's an instance of the instrumentation class?
Let's log a warning that it's being ignored in that case.

It's a bit weird because the additional instrumentation needs to be an instance of the instrumentation, but I understand it's not possible to update the instance with our default config afterwards.

src/client.ts
```diff
@@ -178,73 +214,129 @@ export class Client {
...
+      })
+
+    const defaultInstrumentationNames = Object.keys(DefaultInstrumentationMap)
+
+    const additionalInstrumentations = Object.entries(instrumentationsOption)
+      .filter(([key, instrumentation]) => {
+        const isDefault = defaultInstrumentationNames.includes(key)
+        const isInstance = instrumentation instanceof InstrumentationBase
+
+        return !isDefault || isInstance
```
tombruijn, line 323: This allows for default instrumentation that are configured with an instance to be added as part of the additionalInstrumentations, right? Would they be added twice in that case?

src/client.ts
unflxw: They could give it any key, yes. We should probably tell them to give it the name of the package -- for consistency, but also for compatibility if we add that package as a default in the future.

src/client.ts
unflxw: They shouldn't be added twice, because they're filtered out from `defaultInstrumentations` above if they're instances.

tombruijn:
> How do you feel about the disableInstrumentations + addInstrumentations duo option? Those would each have understandable, helpful signatures.

I like the split in the configuration.

I think `disableInstrumentations` would work better as `disableDefaultInstrumentations`, so it's even more clear what you're disabling, and not also disabling "added" instrumentations.

About `addInstrumentations`, the name `additionalInstrumentations` can also work here to clarify it adds new instrumentations to an existing list of defaults.

What would happen if someone added a new instance of an instrumentation that's also in our default instrumentations, but they didn't disable it first? Does it get added twice?

unflxw:
> What would happen if someone added a new instance of an instrumentation that's also in our default instrumentations, but they didn't disable it first? Does it get added twice?

Yes, unless we do something specifically to prevent it. I suppose we could check if any of the additional instrumentations provided by the user has the same constructor name as one of our default instrumentations, and if so, drop one of the two.

unflxw:
Closing in favor of the approach described in the comments.

Issue #731, opened 2022-09-23
Magic attribute helpers


unflxw:
[As discussed on Slack](https://appsignal.slack.com/archives/C035QHVGM7Y/p1663930125649939), we could write some simple magic attribute helpers so that the simplest custom instrumentation use cases do not require thinking about spans, tracers or OpenTelemetry. Here's the gist of the proposal -- these two code snippets should do the same thing:

```js
// Before:
const activeSpan = opentelemetry.trace.getActiveSpan();
activeSpan.setAttribute(
  "appsignal.custom_data",
  JSON.stringify({"chunky": "bacon"})
)
activeSpan.setAttribute("appsignal.tag.color", "blue")

// After:
appsignal.setRootName("Action name for incident")
appsignal.setName("Action name for the child span")
appsignal.setCategory("event.category")
appsignal.setBody("SELECT * FROM users")
appsignal.setRequestHeader("HEADER-NAME", "Header value")
appsignal.setParams({"chunky": "bacon"})
appsignal.setSessionData({"chunky": "bacon"})
appsignal.setCustomData({"chunky": "bacon"})
appsignal.setTag("color", "blue")
```

Similar `setHeaders`, `setParams`... helpers would work in the same way. An additional `setError` would do the whole `recordException` dance for you.

PR #742, opened 2022-10-03
Fix dependency hell

unflxw:
### [Update config global declaration](https://github.com/appsignal/appsignal-nodejs/commit/2ff85131f7feff665dfbb438084223e61231f1d9)

The way we declared the config here does not work with the latest
TypeScript versions. This is the new recommended way of doing it.

### [Bump dependencies and add overrides](https://github.com/appsignal/appsignal-nodejs/commit/7c6e12638b4398389814fe3cfa6a52e4390803fd)

Bump OpenTelemetry library dependencies as to be able to use
`trace.getActiveSpan()`, which was added in version 1.2.0 of the
`@opentelemetry/api` library. Bump other OpenTelemetry libraries,
as they are all in a weird intricate lockstep with each other.

To avoid type errors where TypeScript complains that two types
from different versions of the same package are different types,
add overrides to force package deduplication.

### [Pin specific instrumentation versions](https://github.com/appsignal/appsignal-nodejs/commit/036cf7f36c5c9dcf9589b06172de31b1392d8d73)

Instead of using whatever version of an instrumentation is bundled
with `@opentelemetry/auto-instrumentations-node`, pin specific
versions of the specific instrumentations we actually care about.

---

Started doing this as part of #731, but it kind of became its own thing, so I split it into a separate PR, as it's worth having a separate discussion about these changes independently of the magic attribute helpers. That said, the magic attribute helpers do need the `@opentelemetry/api` version bump in order to use `trace.getActiveSpan()`

[skip changeset]


tombruijn (approved):
Looks good to me. And if new versions of those packages are released we can remove the overrides again later.

unflxw:
@tombruijn Ideally, since the overrides are `^version`, they would allow to upgrade to the newer version, but also _force the upgraded version on every package_. But I'm not sure that's how it works since the documentation isn't very clear on it.

PR #743, opened 2022-10-03
Add magic attributes

unflxw:
Implement the magic attribute helpers as described in #731, except:

- The helpers are imported from the library, not from the client, meaning users don't need to pass the client around:

```js
import { setTag, setParams } from '@appsignal/nodejs'
```

- The helper for `setRootName` is not implemented yet, as its corresponding magic attribute is not implemented in the processor. A task bullet point has been created in appsignal/opentelemetry#46 to implement it later.

---

Requires the dependency bump in #742, so it's built on top of it -- ignore the other commits.


src/helpers.ts
```diff
@@ -0,0 +1,57 @@
...
+  setAttribute("appsignal.name", name)
+}
+
+export function setCategory(category: string) {
+  setAttribute("appsignal.category", category)
+}
+
+export function setBody(body: string) {
+  setAttribute("appsignal.body", body)
+}
```
tombruijn, line 57: Can we add some basic unit tests for these helpers?

package.json
```diff
@@ -12,16 +12,31 @@
...
     "node-gyp": "^9.0.0",
     "tslib": "^2.0.3",
     "winston": "^3.6.0"
   },
+  "overrides": {
+    "@opentelemetry/api-metrics": "^0.33.0",
+    "@opentelemetry/core": "^1.7.0",
+    "@opentelemetry/instrumentation": "^0.33.0",
+    "@opentelemetry/semantic-conventions": "^1.7.0"
+  },
```
tombruijn, line 39: > Requires the dependency bump in https://github.com/appsignal/appsignal-nodejs/pull/742, so it's built on top of it -- ignore the other commits.

You know you can change the base branch to your other PR so these commits don't show in this PR?

unflxw: Ah, should've done that, yes 😅

unflxw:
TO-DO before merge:
- [ ] Add unit tests (thanks @tombruijn!)
- [ ] Add `setNamespace` (thanks @noccor!)

tombruijn (approved):

PR #744, opened 2022-10-03
Add error helpers

### LINKED ISSUE #741: Set error and send error helpers for OpenTelemetry
Related issue https://github.com/appsignal/appsignal-nodejs/issues/731
```js
// Stores an error event as an event on the active span
appsignal.setError(some error object)
// Creates a new trace with a single span that has an error event
appsignal.sendError(some error object, callback)
```

unflxw:
### [Add `{set,send}Error` helpers](https://github.com/appsignal/appsignal-nodejs/pull/744/commits/e7f80f4e796d96049431681da06e61bb3f0baf6d)

Adds the `setError` and `sendError` helpers, as described in https://github.com/appsignal/appsignal-nodejs/issues/741.

Note that the `sendError` callback does not receive a span as an
argument. It creates an active span and runs your callback inside
it. As the other helpers do not require a span as an argument, you
can use those helpers from the callback without having a span
passed as an argument.

### [Print all spans on failure](https://github.com/appsignal/appsignal-nodejs/pull/744/commits/25dc929f1a08d6d282e388458df015981a801d39)

When the integration tests fail, they should print all the spans
they have seen, to help with debugging.

### [Use helpers to set magic attributes](https://github.com/appsignal/appsignal-nodejs/pull/744/commits/3932265f5e65fddde37648cf52ae6c322e970a38)

Use our helpers internally when setting errors, params or magic
attributes for the Express and Koa instrumentations. As a result,
magic attributes for Express params and cookies are now set in the
root span, not in the request handler span.

Fix existing tests for the params and add tests for the session and
the error in Express, as well as a test for custom instrumentation.

###  [Warn if there is no active span](https://github.com/appsignal/appsignal-nodejs/pull/744/commits/13359068c5d950e131b5019c08a56338c60408ef)

Instead of silently failing, emit a debug-level warning to the logs
when a helper attempts to set an attribute, but there is no active
span to set it on.

Also emit a debug-level warning to the logs when a helper that takes
an Error-like object as argument receives an value which is not
Error-like, as it is common to throw non-Error values.

---

Fixes #741.

Moved the "Remove tracer provider" bit from the draft to a separate PR.
[skip changeset]

### DIFF
--- a/src/helpers.ts
+++ b/src/helpers.ts
@@ -1,9 +1,16 @@
 import { SpanStatusCode, AttributeValue, trace } from "@opentelemetry/api"
+import { Client } from "./client"
 
 function setAttribute(attribute: string, value: AttributeValue) {
   const activeSpan = trace.getActiveSpan()
   if (activeSpan) {
     activeSpan.setAttribute(attribute, value)
+  } else {
+    const splitAttributes = attribute.split(".")
+    const attributeSuffix = splitAttributes[splitAttributes.length - 1]
+    Client.logger.debug(
+      `There is no active span, cannot set \`${attributeSuffix}\``
+    )
   }
 }
 
@@ -87,7 +94,13 @@ export function setError(error: Error) {
         code: SpanStatusCode.ERROR,
         message: error.message
       })
+    } else {
+      Client.logger.debug(
+        `There is no active span, cannot set \`${error.name}\``
+      )
     }
+  } else {
+    Client.logger.debug("Cannot set error, it is not an `Error`-like object")
   }
 }
 
@@ -100,5 +113,7 @@ export function sendError(error: Error, fn: () => void = () => {}) {
         fn()
         span.end()
       })
+  } else {
+    Client.logger.debug("Cannot send error, it is not an `Error`-like object")
   }
 }
--- a/src/helpers.ts
+++ b/src/helpers.ts
@@ -1,9 +1,9 @@
-import { AttributeValue, trace } from "@opentelemetry/api"
+import { SpanStatusCode, AttributeValue, trace, Span } from "@opentelemetry/api"
 
 function setAttribute(attribute: string, value: AttributeValue) {
-  const currentSpan = trace.getActiveSpan()
-  if (currentSpan) {
-    currentSpan.setAttribute(attribute, value)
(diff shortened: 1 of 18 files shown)


src/helpers.ts
```diff
@@ -77,3 +77,28 @@ export function setBody(body: string) {
...
+      })
+    }
+  }
+}
+
+export function sendError(error: Error, fn: () => void = () => {}) {
+  if (error && error.name && error.message) {
+    trace
+      .getTracer("Appsignal.sendError")
+      .startActiveSpan(error.name, { root: true }, span => {
```
tombruijn, line 111: The changing of the active span doesn't break anything down the line? Like if you send an error this way and then want to continue working with the previous active span, that will work?

For the AppSignal span and transaction APIs I think it "broke" the current span if we did the flow like this with changing the active span.

src/helpers.ts
```diff
@@ -77,3 +77,28 @@ export function setBody(body: string) {
...
+        message: error.message
+      })
+    }
+  }
+}
+
+export function sendError(error: Error, fn: () => void = () => {}) {
+  if (error && error.name && error.message) {
+    trace
+      .getTracer("Appsignal.sendError")
```
tombruijn, line 110: If this name determines the "instrumentationLibrary" name, we may want to create an extractor for it, or at least not have it warn about a missing extractor.

unflxw: I believe that, since it will always create a root span, it should not have a visible warning about a missing extractor. But happy to add an extractor for it.

src/helpers.ts
unflxw: I think not -- OpenTelemetry docs say that the span will be active _for the duration of the span_ -- but I'll add a test to check.

src/helpers.ts
unflxw: I've checked the extractor and I'm not sure what would need to change from the default handling. Let me know your thoughts on what the extractor should do.

tombruijn: It looks better than I thought. It doesn't say anything about a missing incident action name. The other warnings only really show on performance samples in the event timeline.

tombruijn (approved):

PR #746, opened 2022-10-04
Add instrumentation config options

unflxw:
### [Move test mode span processor](https://github.com/appsignal/appsignal-nodejs/commit/020b2736aa25f188f191a09224cc5cba07f6fa22)

Move the test mode span processor out of the client and with the
other span processor.

### [Implement instrumentation config options](https://github.com/appsignal/appsignal-nodejs/commit/beb680413086fa8a7e6035c7842f3dfe65a83218)

Implement the `disableDefaultInstrumentations` and
`additionalInstrumentations` config option.

`disableDefaultInstrumentations` takes either a list of
instrumentation names to disable as an argument, or a true/false
boolean.

If set to true, all default instrumentations will be disabled,
providing an easy way for people who would rather configure
OpenTelemetry instrumentations on their own. If set to `false`,
no default instrumentations are disabled.

`additionalInstrumentations` takes any `InstrumentationOption`, as
defined by the `instrumentations` config option in the `NodeSDK`
configuration object.

### [Remove instrument{Http,Pg,Redis} config options](https://github.com/appsignal/appsignal-nodejs/commit/86987805707e27b682d1ce6510a5b2e4aae6133f)

Remove the configuration options to disable specific
instrumentations, as that's taken care of by the new
`disableInstrumentations` config option.

A re-do of #739. Fixes appsignal/opentelemetry#45.


src/span_processor.ts
```diff
@@ -52,3 +52,48 @@ export class SpanProcessor implements OpenTelemetrySpanProcessor {
     return Promise.resolve()
   }
 }
+
+import * as fs from "fs"
```
tombruijn, line 56: Let's move that to the top of the file.

src/__tests__/client.test.ts
```diff
@@ -102,4 +104,83 @@ describe("Client", () => {
...
+      expect(instrumentationNames).toContain(
+        "@opentelemetry/instrumentation-http"
+      )
+    })
+
+    it("can disable all default instrumentations", () => {
+      client = new Client({
+        ...DEFAULT_OPTS,
+        active: true,
+        disableDefaultInstrumentations: true
```
tombruijn, line 126: Great addition! That will be very convenient.

tombruijn (approved):
Looks like you need to update the diagnose tests submodule to fix the CI build.

src/span_processor.ts
unflxw: 🤦 Thank you for spotting this!

PR #750, opened 2022-10-06
Add root name magic attribute helper

tombruijn:
Allow users change the incident's action name by overriding the root span name for OpenTelemetry traces. This is the counterpart to the `setName` helper that only changes the active span's name. The "name" translates to the "title" you see on the box when you hover over an event in the event timeline.

This helper does the same as `rootSpan.setName("Name")`, but not always can users access the root span of the trace. Which is where this helper comes in to make it still possible to overwrite the root span's name. If the `setRootName` function is used, it overwrites any `rootSpan.setName` value.

I'm also okay with changing this name to `setTraceName` if that would make it more clear what is the root in this context.

This also requires the processor change I submitted earlier.

[skip changeset]

Part of https://github.com/appsignal/opentelemetry/issues/46

### DIFF
--- a/src/__tests__/helpers.test.ts
+++ b/src/__tests__/helpers.test.ts
@@ -17,6 +17,7 @@ import {
   setSessionData,
   setTag,
   setNamespace,
+  setRootName,
   setError,
   sendError
 } from "../helpers"
@@ -85,6 +86,7 @@ describe("Helpers", () => {
       setHeader("content-type", "application/json")
       setTag("something", true)
       setNamespace("web")
+      setRootName("Root name")
 
       span.end()
     })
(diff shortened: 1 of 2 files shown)


jeffkreeftmeijer (approved):

unflxw:
> This helper does the same as rootSpan.setName("Name"), but not always can users access the root span of the trace.

At the moment, we're only documenting `setName` as a "Child span helper". I think that's how we should talk about them -- `setName` goes with `setCategory` and `setBody` and it's only for child spans, and this one goes with the other ones that apply to the whole sample (`setParams`, etc...)

> I'm also okay with changing this name to `setTraceName` if that would make it more clear what is the root in this context.

👍 to that -- but wouldn't the AppSignal-y word for this be "sample"? So maybe `setSampleName`?

tombruijn:
> 👍 to that -- but wouldn't the AppSignal-y word for this be "sample"? So maybe setSampleName?

@unflxw The most accurate name would be `setIncidentActionName`. We internally call it the action name for the incident. "Action name" has not been used in any naming since the Transaction API, instead using `setName` in the Span API.

I'd rather not cling to old naming if we're going to change it anyway in the future. That's why my first suggestion was `setRootName` because it's clear it updates the root span's name, but `setTraceName` is also an option because the root span determines the trace name.

unflxw:
I'm happy with any of these names.

PR #757, opened 2022-10-11
Only set params on router layer

unflxw:
To avoid setting the params in many different spans, potentially interfering with custom instrumentation for the params, only set the params once, on the router layer.

TO-DO:
- [ ] Update the dependency on `@opentelemetry/instrumentation-koa` to a released version that contains the changes in this PR: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1226
- [ ] Update the `koa-mysql` integration test as necessary


luismiramirez (approved):
I think this one needs a changeset

PR #771, opened 2022-10-24
Add OpenTelemetry SpanKind as a magic attribute

luismiramirez:
To identify the origin of the HTTP requests from their spans, the `SpanKind` attribute is needed. As adding a completely new attribute will add a lot of overhead requiring modifications in all steps, the attribute is now added from the SpanProcessor as a magic attribute.

More info on OpenTelemetry's SpanKind:
https://opentelemetry.io/docs/reference/specification/trace/api/#spankind

Fixes: https://github.com/appsignal/opentelemetry/issues/42

### DIFF
--- a/src/span_processor.ts
+++ b/src/span_processor.ts
@@ -1,5 +1,6 @@
 import * as fs from "fs"
 import type { Context } from "@opentelemetry/api"
+import { SpanKind } from "@opentelemetry/api"
 import type {
   Span,
   ReadableSpan,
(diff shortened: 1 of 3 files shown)


src/span_processor.ts
```diff
@@ -22,6 +23,9 @@ export class SpanProcessor implements OpenTelemetrySpanProcessor {
   onStart(_span: Span, _parentContext: Context): void {}
 
   onEnd(span: ReadableSpan): void {
+    // Add OpenTelemetry kind enum value as a magic attribute
+    span.attributes["appsignal.kind"] = SpanKind[span.kind]
```
unflxw, line 27: Nitpick: I'm wondering if modifying the given OpenTelemetry span would cause problems if there were other span processors -- as in, if it would add our magic attribute for those other span processors as well.

It may be safer to rebuild the attributes object, instead of modifying it in-place in the ReadableSpan object.

```js
const spanAttributes = {
  ...span.attributes,
  "appsignal.kind": SpanKind[span.kind]
}
// then pass this to createOpenTelemetrySpan
```

unflxw (approved):
The idea sounds good to me -- I don't think we need to change the signatures of all the methods just to pass this attribute on to the extractor.

src/span_processor.ts
luismiramirez: Thanks for this. I prefer your approach; changing it

PR #772, opened 2022-10-24
Add Fastify support

luismiramirez:
Fastify is now instrumented via OpenTelemetry contrib package.


unflxw (approved):
Should we wait to merge this with an agent release from https://github.com/appsignal/appsignal-agent/pull/842?

luismiramirez:
> Should we wait to merge this with an agent release from https://github.com/appsignal/appsignal-agent/pull/842?

Yep, I'm currently trying to fix that flaky integration spec, but we can probably make a release without fixing it.

PR #773, opened 2022-10-25
Add Mongoose support

unflxw:
Add the OpenTelemetry instrumentation for Mongoose and add a test app that checks for the span generated by it.

Part of https://github.com/appsignal/appsignal-nodejs/issues/758.

### On database statements

By default, the Mongoose instrumentation does not emit database statements. However, it is possible to provide a `dbStatementSerializer` in the config for the instrumentation, whose return value will be set as the `db.statement` attribute in the span.

This config option is undocumented, and it receives differently shaped objects depending on the Mongoose method that is being instrumented. It should be possible to write a serializer in order to extract more informatio from it, if we choose to, but it looks like it wouldn't be easy.

Given that the span names are relatively well-formed, showing the model and the operation that are queried (similarly to what we obtain from Prisma) I've not taken on the task of creating a statement serializer.

### On dependency versions

The instrumentation has a development dependency on the `mongoose` library, as it uses some of the library-defined types for its own typing.

Due to this, I needed to use the same `mongoose` version (`6.5.2`) as a dev-dependency in the integration, and as a dependency in the test application. Using other versions would result on typing errors.

I'm not sure to what extent this would be a problem in real-world applications for our users.

I suspect the issue would be caused by the TypeScript compiler trying to compile the `appsignal.ts` initialiser file (which depends on `mongoose` version `6.5.2`) and the app (which depends on a different `mongoose` version) as a single compilation unit. I also suspect that the integration being linked into the app using `npm link` is part of the problem, as that means the integration has its own `node_modules` folder with a different `mongoose` version in it.

We should test this against the released package. Regardless, using an `appsignal.cjs` file (excluded from the compilation process) instead of an `appsignal.ts` file should prevent these compilation errors due to version mismatches.


luismiramirez (approved):
Apart from the super minor things I commented, I think a patch changeset is missing.

Issue #763, opened 2022-10-11
FS module instrumentation


tombruijn:
The Node.js module that does things with the file system.

## To do
- [x] Add instrumentation to the client by default (https://github.com/appsignal/appsignal-nodejs/pull/777)
  - Only if the instrumentation library is part of the OpenTelemetry core or auto instrumentation package
- [x] Add extractor for OpenTelemetry spans (https://github.com/appsignal/appsignal-agent/pull/846)
- [x] Add integration test for spans (https://github.com/appsignal/appsignal-nodejs/pull/777)

luismiramirez:
Reason why this issue is blocked: ttps://github.com/appsignal/appsignal-nodejs/pull/777#issuecomment-1295054358

PR #777, opened 2022-10-27
Add fs module support

luismiramirez:
All fs module usage is now automatically instrumented.

Tests are in the express-redis sample application.

Automatic instrumentation is manually suppresed in the test span processor because it uses the fs module to handle the spans.json file. It is done using the `supressTracing` helper provided by OpenTelemetry core library.

Part of #763.

### DIFF
--- a/.changesets/add-fs-module-support.md
+++ b/.changesets/add-fs-module-support.md
@@ -1,0 +1,6 @@
+---
+bump: "patch"
+type: "add"
+---
+
+Add fs module support
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,18 +1,19 @@
 {
   "name": "@appsignal/nodejs",
-  "version": "3.0.0-beta.3",
+  "version": "3.0.0-beta.4",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@appsignal/nodejs",
-      "version": "3.0.0-beta.3",
+      "version": "3.0.0-beta.4",
       "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
         "@opentelemetry/api": "^1.2.0",
         "@opentelemetry/instrumentation-express": "^0.31.2",
         "@opentelemetry/instrumentation-fastify": "^0.30.0",
+        "@opentelemetry/instrumentation-fs": "^0.5.1",
         "@opentelemetry/instrumentation-graphql": "^0.31.0",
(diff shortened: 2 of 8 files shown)


jeffkreeftmeijer (approved):

unflxw:
I have looked into the issue where the Prisma test applications fail to boot when the FS instrumentation is added.

TL;DR: We gotta wait for a release containing [the changes in this PR](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1222) to be published (should be included in [the next release](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1214)) and update the `@opentelemetry/instrumentation-fs` package to that version.

The core of the issue is that the Prisma client calls `util.promisify(fs.exists)`. It turns out that `fs.exists` is a bit unusual. It's deprecated by the Node team, because its callback signature does not follow the standard `(err, result) => { ... }` signature of Node callbacks. Instead, since `fs.exists` can't fail, its signature is just `(result) => { ... }`.

Since `util.promisify` expects the first argument of the callback to be an error, this means that promisifying `fs.exists` would result on the promise always being rejected, with the actual result being the rejection error.

This is consistent with the error message we see:
> ```
> error2.clientVersion = this.client._clientVersion;
> ```
> ```
> TypeError: Cannot create property 'clientVersion' on boolean 'true'
> ```

Prisma is trying to set the clientVersion property on the error object it received, but the error object isn't an object, but a boolean `true`, the return value of `fs.exists`, turned into the rejection value of a promise by `util.promisify`.

Now, the reason why this bug only happens for the instrumented version of the function is that the original `fs.exists` provides a custom promisified implementation of itself. If there is an `util.promisify.custom` symbol as a property of the function, `util.promisify` will use its value as the promisified implementation. The original `fs.exists` provides this, but the OpenTelemetry-instrumented version of the function doesn't.

Or rather, it doesn't _yet_, as there's [a merged PR implementing this custom behaviour](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1222) for the instrumented version of the function. This should fix our problems; we just need to wait for [a release including the change](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1214).

I have not been able to verify the actual fix directly (npm's "install from GitHub" does not play well with monorepos) but I have confirmed that manually excluding `fs.exists` from being instrumented by commenting the relevant line of code fixes the issue, so I expect the fix above to also fix the issue.

unflxw (approved):

Issue #781, opened 2022-11-07
Logging


thijsc:
Support for AppSignal logging from the Node integration. See https://github.com/appsignal/appsignal-ruby/pull/890 for an exampe implementation.

PR #790, opened 2022-11-10
Implement logging feature

unflxw:
This PR implements the logging feature (#781), closely mimicking [its Ruby implementation](https://github.com/appsignal/appsignal-ruby/pull/890).

There's [an integration test on the agent side](https://github.com/appsignal/appsignal-agent/compare/logging...wip-logging-nodejs?expand=1) that needs fixing -- it works, but it currently requires a sleep in order for the log lines to be flushed to the endpoint. When it's fixed there will be a PR for that, or it will be added to [the existing logging PR for the agent](https://github.com/appsignal/appsignal-agent/pull/847).

The build of the agent used in this PR at the moment is a build from [commit `9b62288`](https://github.com/appsignal/appsignal-agent/pull/847/commits/9b6228857df2d405ab6769a4a1596ad272298bcd) in the agent logging PR. We may want to make a new build of the agent and update this, once the agent logging PR has been merged.

### [Rename Logger to IntegrationLogger](https://github.com/appsignal/appsignal-nodejs/commit/97d41361f6ec9c4d574a7668602fe303ffa19a43)

This follows the rename performed [in the Ruby logging PR][pr].
The existing logger used by the integration is renamed to the
integration logger, so that a new user-facing logger using the logging
API can be implemented.

### [Implement Logger class](https://github.com/appsignal/appsignal-nodejs/commit/67e42d034b073ebd2d87c4fce89d268982b9d3a0)

Implements a logger class that uses the C API bridge to log messages
through the Logging API. It follows the same user-facing API as the
one implemented in the [Ruby logging PR][pr], except that it uses
the logging levels that were in use in the Node.js logger.

On the user facing API, the logger level is implemented as a string.
The attributes are typed as a flat map whose values must be strings,
numbers or booleans, in concordance with how the agent traverses the
resulting data map.

[pr]: https://github.com/appsignal/appsignal-ruby/pull/890

### [Add logging endpoint config option](https://github.com/appsignal/appsignal-nodejs/commit/75b6162281858e850006bfd50755997f689857c4)

The logging endpoint must be configurable, and the value given to
it must be passed on to the agent.


unflxw:
Oh, gotta add it to the diagnose tests.

src/logger.ts
```diff
@@ -0,0 +1,72 @@
+import { Client } from "./client"
+
+export type LoggerLevel = "trace" | "debug" | "info" | "warn" | "error"
```
luismiramirez, line 3: What about implementing this one as an enum?

src/logger.ts
```diff
@@ -1,75 +1,72 @@
...
+  private log(
+    severity: number,
+    message: string,
+    attributes: LoggerAttributes = {}
+  ) {
+    if (severity < this.severityThreshold) {
+      return
     }
+
+    this.#client.extension.log(this.#group, severity, `${message}`, attributes)
```
tombruijn, line 70:
```suggestion
    this.#client.extension.log(this.#group, severity, message, attributes)
```
Does the message need to be wrapped like this?

tombruijn (approved):

src/logger.ts
unflxw: This is meant as a bit of defensive programming. The message should already be a string, so this does nothing most of the time.

But if it's not a string, this performs [string coercion](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String#string_coercion) on it. This ensures that whatever is passed to the extension is definitely a string.

Actually, looking at those docs now, it would be preferable to use `String()` to avoid raising a TypeError if a symbol is passed.

src/logger.ts
unflxw: My initial implementation for this was an enum! I'm willing to be convinced that it should be an enum.

I decided against it because I thought it was nicer for users (especially JavaScript users) to not have to import the LoggerLevel enum:

```js
// with enum
import { Appsignal, LoggerLevel } from "@appsignal/nodejs";

Appsignal.logger("groupname", LoggerLevel.trace)

// without enum
import { Appsignal } from "@appsignal/nodejs";

Appsignal.logger("groupname", "trace")
```

luismiramirez: True, I tend to forget that not everyone is using TS.

luismiramirez (approved):

src/logger.ts
unflxw: Resolving with https://github.com/appsignal/appsignal-nodejs/pull/790/commits/c151a11b183c95c7d2bbd5cd1f505409456f75c8.

PR #799, opened 2022-11-21
Add request params support for Fastify

luismiramirez:
Fastify apps now record parameters in the HTTP spans.

Closes #778

Issue #762, opened 2022-10-11
Knex instrumentation


tombruijn:
Knex is a query builder used by the Bookshelf ORM.

## To do
- [x] Add instrumentation to the client by default
  - Only if the instrumentation library is part of the OpenTelemetry core or auto instrumentation package
- [x] Add extractor for OpenTelemetry spans
- [x] Add integration test for spans
- [x] Add colors and grouping to server repo
- [x] Add the instrumentation to the docs

unflxw:
@tombruijn Added a TO-DO point to add docs to it.

PR #803, opened 2022-11-22
Add Knex.js instrumentation

tombruijn:
- Add the Knex.js instrumentation package as a dependency.
- Load it by default.
- Add an integration test app for Knex.js.

Part of https://github.com/appsignal/appsignal-nodejs/issues/762

---

Based on #799, because it needs the latest `@opentelemetry/api` package version.


test/express-knex/app/package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "express-apollo-example",
```
luismiramirez, line 2: Nit: 
```suggestion
  "name": "express-knex-example",
```

luismiramirez (approved):

unflxw (approved):

PR #806, opened 2022-11-23
Bump agent to version 0c870c0

unflxw:
Update the agent version to perform hoisting on the agent instead of in the processor, and to use the merged version of the logging branch.


tombruijn (approved):

tombruijn:
I usually add a changeset to also include things not otherwise mentioned, like the GraphQL extractor improvements, but it's okay for now.

luismiramirez (approved):

PR #817, opened 2022-12-27
Only instrument `fs` with parent span

### LINKED ISSUE #813: Implement requireParentSpan for fs instrumentation
Other OTel instrumentations (such as Mongoose) have a `requireParentSpan` option. When active, they will only create a span if there's already an active span in the current context that can be their span's parent -- that is, their span will never be the root span.

This could be useful in reducing noise in our users' dashboards, as users' applications and libraries often read files at points which are not already in an instrumentation context. A trace consisting only of "fs openFile" every time you boot your app is not particularly useful to the user.

unflxw:
Fixes #813.

After implementing the OTel PR with the `requireParentSpan` configuration option, while ungarbling the README file in order to document it, I realised that there was an easy way to do this already available within the instrumentation. 🤦

This is that easy way. It might be good to replace this with `requireParentSpan` once/if it's merged, but it's merely a nice-to-have.

### DIFF
--- a/.changesets/only-instrument-fs-with-parent-span.md
+++ b/.changesets/only-instrument-fs-with-parent-span.md
@@ -1,0 +1,6 @@
+---
+bump: "patch"
+type: "fix"
+---
+
+Prevent creating fs-only samples by only creating an fs span if there is a parent span present.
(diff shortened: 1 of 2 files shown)


luismiramirez (approved):
LGTM after adding a changeset

jeffkreeftmeijer (approved):

PR #819, opened 2023-01-02
Track restify request parameters

### LINKED ISSUE #787: Restify instrumentation
Web framework.

https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/plugins/node/opentelemetry-instrumentation-restify

## To do
- [x] Add instrumentation to the client by default
  - Only if the instrumentation library is part of the OpenTelemetry core or auto instrumentation package
- [x] Add extractor for OpenTelemetry spans
- [x] Add integration test for spans
- [x] Add docs
- [x] Add `requestHook` config option to restify instrumentation: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1312
- [x] Update restify instrumentation with requestHook config option to set request parameters

tombruijn:
Upgrade to the latest restify package that includes the `requestHook` config option to report the parameters.

We only track them for the request handler layer type and not all middlewares that may be present to avoid it being set multiple times and it overwriting one another.

This relies on the `params` and `query` properties on the request. These are the parsed request params and query params for the request that are only available when the restify queryParser plugin is loaded. This avoids us having to parse it separately, and I assume it will only be necessary to track parameters if the plugin is added to the app.

Closes #787

### DIFF
--- a/test/restify/app/src/app.ts
+++ b/test/restify/app/src/app.ts
@@ -6,6 +6,7 @@ function respond(req: any, res: any, next: any) {
 }
 
 const server = restify.createServer()
+server.use(restify.plugins.queryParser())
 server.get("/", function (req: any, res: any, next: any) {
   res.send("home")
   return next()
--- a/.changesets/report-request-parameters-for-restify.md
+++ b/.changesets/report-request-parameters-for-restify.md
@@ -1,0 +1,6 @@
+---
+bump: "patch"
+type: "add"
+---
+
+Report request parameters for restify apps. This is reported automatically. You can disable parameter reporting with the AppSignal `sendParams: false` configuration option.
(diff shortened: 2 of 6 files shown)


test/restify/app/src/app.ts
```diff
@@ -6,6 +6,7 @@ function respond(req: any, res: any, next: any) {
 }
 
 const server = restify.createServer()
+server.use(restify.plugins.queryParser())
```
unflxw, line 9: Would it make sense for the integration test to look for the presence of the `appsignal.request.params` attribute?

unflxw (approved):

test/restify/app/src/app.ts
tombruijn: I've updated the tests for it.

luismiramirez (approved):

PR #820, opened 2023-01-04
Log error on appsignal.tracer function call

### LINKED ISSUE #815: Add tracer deprecation notice
In the 3.x integration `Appsignal.tracer` is no longer a thing. Code that uses it, such as users' custom instrumentation for 2.x and our own 2.x integrations, will receive `undefined` as a return value and fail shortly thereafter.

We've seen cases on support of people mixing up 3.x and 2.x installation instructions, as well as confusion during upgrades. [As discussed on Slack](https://appsignal.slack.com/archives/CNPP953E2/p1671457206045279?thread_ts=1671454381.770839&cid=CNPP953E2), we could log a message when `Appsignal.tracer` is accessed, explaining the issue and pointing users to [our migration guide](https://docs.appsignal.com/nodejs/3.x/migration-guide.html).

## To do

- [x] Add `Appsignal.tracer` function
- [x] Print an upgrade message when called

tombruijn:
We notice that some people, after upgrading to version 3 of our package, run into crashes with undescriptive error messages. This happens when `appsignal.tracer()` is called.

To help with the upgrade process, log a message with a link to the upgrade guide.

Closes #815

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -387,9 +387,10 @@ export class Client {
     global.__APPSIGNAL__ = this
   }
 
-  tracer(): void {
+  get tracer(): any {
     console.error(
       "The `appsignal.tracer()` function was called, but it has been removed in AppSignal for Node.js package version 3.x. Please read our migration guide to upgrade to this new version of our package: https://docs.appsignal.com/nodejs/3.x/migration-guide.html. It is also possible to downgrade to version 2.x, after which this code will work again."
     )
+    return () => {}
   }
 }


src/client.ts
```diff
@@ -386,4 +386,10 @@ export class Client {
   private storeInGlobal(): void {
     global.__APPSIGNAL__ = this
   }
+
+  tracer(): void {
```
unflxw, line 390: With the previous behaviour (where `tracer` is undefined) the application crashes on the function call itself ("TypeError: x is not a function"), but with the changes in this PR, it might not error until it's used later (by calling `.createSpan()` on it, for example)

I think that the earlier the error is thrown, the better. Consider either explicitly throwing an error when this function is called, or turning it into a getter (or both?)

```suggestion
  get tracer(): void {
```

src/client.ts
```diff
@@ -386,4 +386,10 @@ export class Client {
...
+    console.error(
+      "The `appsignal.tracer()` function was called, but it has been removed in AppSignal for Node.js package version 3.x. Please read our migration guide to upgrade to this new version of our package: https://docs.appsignal.com/nodejs/3.x/migration-guide.html. It is also possible to downgrade to version 2.x, after which this code will work again."
```
unflxw, line 392: It might be good to mention that one of our other packages (`@appsignal/express` and the like) might be the one calling this function, and that they should remove them? The migration guide does say that, so it might be fine.

> It is also possible to downgrade to version 2.x, after which this code will work again.

This is true, but we may not want to point users in that direction.

unflxw (approved):
Looks good, left some nitpicks -- feel free to ignore.

luismiramirez (approved):

src/client.ts
tombruijn: You want to change it to a getter so it errors on `client.tracer` rather than the function call `client.tracer()`? Sure, that's fine. It breaks either way.

src/client.ts
tombruijn: When I see error as a user I usually want two options: a way to make it go a way by following a guide, or make it go away by downgrading. I can remove the last line, but that's the two options the reader has when reading this.

unflxw: Yeah, that makes sense. 👍

PR #823, opened 2023-01-05
Add Winston transport for logging

unflxw:
Fixes #821.

Implements a Winston transport for logging. The transport uses the global client instance to log messages via the extension. A best effort is made to preserve additional arguments added via splat, formatters and child loggers.

PR #826, opened 2023-01-05
Add Winston transport for logging

### LINKED ISSUE #821: Add Winston support to the logger
It should be possible to turn an AppSignal logger (from the Logging API beta) into a Winston transport. Something like:

```
const winstonLogger = winston.createLogger({
    transports: [
        Appsignal.logger("groupname").winstonTransport()
    ]
})
```

User-requested; [Intercom link](https://app.intercom.com/a/apps/yzor8gyw/inbox/inbox/conversation/16410700167053) (internal)

TO-DO:
- [x] Create documentation page -- [PR](https://github.com/appsignal/appsignal-docs/pull/1008)

unflxw:
Fixes #821 -- a re-do of #823.

Implements a Winston transport for logging. The transport uses the global client instance to log messages via the extension. A best effort is made to preserve additional arguments added via splat, formatters and child loggers.

Published as a beta version `@appsignal/nodejs@3.0.3-beta.1` for it to be tested by a customer ([Intercom link](https://app.intercom.com/a/apps/yzor8gyw/inbox/inbox/conversation/16410700167053))


unflxw:
Thinking of the child logger use case that is mentioned in the Intercom link: it probably would be nice if we extracted a `group` key from the info, if present, and used it instead of the group passed on to the transport on initialization.

Actually, we may want to make that change throughout the logging feature? Not sure if being forced to pass a group key every time is nice from the users' perspective -- many use cases will only have one group, and the application's logs are already their own log source in the UI. Perhaps we could give it some default value that can be overriden later.

thijsc (approved):
Looking good. Would be good to mention this https://docs.appsignal.com/logging/platforms/integrations/nodejs.html

src/winston_transport.ts
```diff
@@ -0,0 +1,126 @@
...
+      this.emit("logged", info)
+    })
+
+    const client = Client.client
+    if (!client || !client.isActive) {
+      return
+    }
+
+    const levelSeverity = severity(info[Symbol.for("level")])
+    if (levelSeverity == UNKNOWN_SEVERITY) {
```
tombruijn, line 68: We could also fall back on another log level, like trace, so it doesn't disappear completely.

tombruijn (approved):
Please add a to do item on the tracking issue that we need to update the docs as well for this.

src/winston_transport.ts
unflxw: Hmm, yeah, that makes sense. Thanks!

luismiramirez (approved):

kamilcglr:
Hello, thank you for the integration. I tried it but I don't have any logs coming back. I haven't found any documentation, I guess it hasn't been written yet. I wanted to know if this is the right way to proceed. We can discuss on Intercom if needed not to pollute this PR.

```
import { WinstonTransport } from '@appsignal/nodejs'
import { createLogger, transports } from 'winston'

this.logger = createLogger({
        transports: [new WinstonTransport({ group: 'app' }), new transports.Console()],
      })
```

Issue #822, opened 2023-01-05
Update `@opentelemetry/api`, remove `@opentelemetry/api-metrics`


unflxw:
On `npm i @appsignal/nodejs`:

```npm WARN deprecated @opentelemetry/api-metrics@0.33.0: Please use @opentelemetry/api >= 1.3.0```

We should update our dependencies.

tombruijn:
I think this is because of our [local overrides](https://github.com/appsignal/appsignal-nodejs/blob/dcf78f18d2a89cf29c051474ea9902713e474b8b/package.json#L43). Package `@opentelemetry/instrumentation` version 0.34.0 doesn't have this dependency anymore. Can we remove the override @unflxw ?

unflxw:
Probably! If the tests pass without it then we're good.

tombruijn:
@unflxw can you do a quick test of that?

unflxw:
It's not the overrides' fault (though I'll submit a PR to remove them anyway)

Prisma instrumentation has not updated its dependencies on `@opentelemetry/instrumentation` and `@opentelemetry/api` for a while -- the first one brings in the deprecated `@opentelemetry/api-metrics`.

Submitted a PR to Prisma: https://github.com/prisma/prisma/pull/17198

tombruijn:
Prisma PR and our PR merged. Closing.

unflxw:
@tombruijn Should we keep this open until there's a Prisma release that we can upgrade our dependency?

unflxw:
Closed by #842.

PR #829, opened 2023-01-09
Remove package.json overrides

unflxw:
We added these to ensure our tests passed, back when OpenTelemetry instrumentations wildly diverged on their dependencies. This should not be necessary anymore, as they are now fairly unified in that regard.

Part of #822, though not really directly related.

[skip changeset]


tombruijn (approved):
Please update your commit message to contain the PR description to pass the Git linter check.

PR #838, opened 2023-01-17
Add OpenTelemetry SDK start promise as a property

### LINKED ISSUE #837: sendError doesn't work outside routes
When using `sendError()` outside of a route handler in a Koa app, those errors get ignored.

The created spans are `NonRecordingSpan`s:

```
 NonRecordingSpan {
   _spanContext: {
     traceId: '00000000000000000000000000000000',
     spanId: '0000000000000000',
     traceFlags: 0
 }
```

This kind of span is returned when OpenTelemetry's tracing is suppresed: https://github.com/open-telemetry/opentelemetry-js/blob/51273714efd89edc21f69efb58085b98216659ff/packages/opentelemetry-sdk-trace-base/src/Tracer.ts#L81

## TODO

- [x] Check if this is a Koa instrumentation specific issue
- [x] Make `sendError()` always work with recording spans

luismiramirez:
There's a new property in the client called `instrumentationsLoaded`. It contains a promise that registers the instrumentations and the tracer provider. This fixes custom instrumentation in one-off scripts and in app top levels not being reported due to unregistered instrumentations.

## Usage instructions

If you're using `import` syntax

```js
import { Appsignal, sendError } from "@appsignal/nodejs"

await Appsignal.client.instrumentationsLoaded

sendError(new Error("Test error"))
```

If you're not using `import` syntax

```js
import { Appsignal, sendError } from "@appsignal/nodejs"

Appsignal.client.instrumentationsLoaded.then(() => {
  sendError(new Error("Test error"))
})
```

Fixes #837

### DIFF
--- a/.changesets/add-opentelemetry-sdk-initialization-to-client.md
+++ b/.changesets/add-opentelemetry-sdk-initialization-to-client.md
@@ -1,0 +1,9 @@
+---
+bump: "patch"
+type: "add"
+---
+
+Add OpenTelemetry SDK initialization to client
+
+The OpenTelemetry instrumentations are loaded async, this is OK when users rely on automatic instrumentations, but for custom instrumentations
+it might cause spans not to be reported. There's a new property in the client that contains the instrumentations registration that can be awaited for resolution before running any custom instrumentations.
--- a/src/client.ts
+++ b/src/client.ts
@@ -93,6 +93,7 @@ export class Client {
   config: Configuration
   readonly integrationLogger: IntegrationLogger
   extension: Extension
+  instrumentationsLoaded?: Promise<void>
 
   #metrics: Metrics
   #sdk?: NodeSDK
@@ -374,7 +375,7 @@ export class Client {
       spanProcessor
     })
 
-    sdk.start()
+    this.instrumentationsLoaded = sdk.start()
 
     return sdk
   }


unflxw:
Note for the documentation: top-level awaits are [only available in Node.js v14.8 onwards](https://www.stefanjudis.com/today-i-learned/top-level-await-is-available-in-node-js-modules/#top-level-%60await%60-is-available-%22unflagged%22-in-node.js-since-%60v14.8%60).

tombruijn (approved):
Should this promise be waited for in the appsignal.cjs file or the users app code?

luismiramirez:
> Should this promise be waited for in the appsignal.cjs file or the users app code?

It should be done in the users' app code. Top-level awaits are only supported when using ES modules as @unflxw linked. Also, `appsignal.cjs` is loaded using `--require`, which cannot be asynchronous.

evenfrost:
@luismiramirez how to use this in the project that doesn't use ES modules?

unflxw:
@luismiramirez I think the example in the PR (and in #840) that says "If you're not using import syntax" should use `require` instead:

```
const { Appsignal, sendError } = require("@appsignal/nodejs")
```

luismiramirez:
@unflxw that's correct, thank you

evenfrost:
@luismiramirez sorry, could you elaborate a bit on what you mean by `out of the app scope`?

Also, does that mean that now I should wrap every AppSignal method individually and not the whole application code once?

And what about native `throw`s? Will they be caught in this case (because wrapping these individually is not quite possible, as they can happen anywhere in the app)?

Sorry for a bunch of questions, just trying to understand how to use this in the most efficient way (considering it's used in the existing pretty large project in terms of code).

luismiramirez:
> could you elaborate a bit on what you mean by out of the app scope?

With "out of the app scope" I mean code that's not triggered by a web request to your app, as in one-off scripts or code that executes once when your app is starting up. For those, you don't have to do anything special because the instrumentations are automatically loaded.

> Also, does that mean that now I should wrap every AppSignal method individually and not the whole application code once?

No, your application code (Koa, Express, HTTP/S modules, etc.) will always be automatically instrumented without the need of wrapping it.

> And what about native throws? Will they be caught in this case (because wrapping these individually is not quite possible, as they can happen anywhere in the app)?

Native throws happening in the aforementioned scope must be controlled by yourself and sent to AppSignal using the `sendError()` helper. In order for that helper to work in that specific case, it must be wrapped with the `instrumentationsLoaded()` promise.

> Sorry for a bunch of questions, just trying to understand how to use this in the most efficient way (considering it's used in the existing pretty large project in terms of code).

No worries! We're here to help.

luismiramirez:
Fix published in [AppSignal for Node.js 3.0.4](https://www.npmjs.com/package/@appsignal/nodejs/v/3.0.4). 

Docs here: https://docs.appsignal.com/nodejs/3.x/instrumentation/instrumentations-load.html

PR #839, opened 2023-01-18
Initialize instrumentationsLoaded with a dummy promise

luismiramirez:
If AppSignal is not active or something prevents the OpenTelemetry sdk from starting, we provide an empty promise to avoid users' apps crashing.

[skip changeset]


luismiramirez:
Gonna try another approach to avoid relying on Appsignal being initialized

PR #840, opened 2023-01-18
Add instrumentationsLoaded as a helper

luismiramirez:
Instead of calling `instrumentationsLoaded()` from a client property, add it as an isolated helper function so users' apps keep working if they haven't initialized the AppSignal client.

## Usage instructions

If you're using `import` syntax

```js
import { instrumentationsLoaded } from "@appsignal/nodejs"

await instrumentationsLoaded()

sendError(new Error("Test error"))
```

If you're not using `import` syntax

```js
import { instrumentationsLoaded } from "@appsignal/nodejs"

instrumentationsLoaded().then(() => {
  sendError(new Error("Test error"))
})
```

### DIFF
--- a/.changesets/add-opentelemetry-sdk-initialization-to-client.md
+++ b/.changesets/add-opentelemetry-sdk-initialization-to-client.md
@@ -3,7 +3,7 @@ bump: "patch"
 type: "add"
 ---
 
-Add OpenTelemetry SDK initialization to client
+Add OpenTelemetry SDK initialization helper function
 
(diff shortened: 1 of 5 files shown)


unflxw (approved):
It's a function, right? So the examples should do `instrumentationsLoaded().then(...)`

luismiramirez:
> It's a function, right? So the examples should do instrumentationsLoaded().then(...)

True!

tombruijn (approved):

PR #841, opened 2023-01-23
Add NGINX metrics support

unflxw:
### [Add enableNginxMetrics config option](https://github.com/appsignal/appsignal-nodejs/pull/841/commits/6138ef8a57815ca82d8caceecddd9b06b9173ab4)

Set its default value to `false` and add it to all config mappings.
Update the diagnose submodule to expect the config option.

### [Type config mappings to ensure keys exist](https://github.com/appsignal/appsignal-nodejs/pull/841/commits/97f28a82619a5162473140cbace0797f675c3a29)

In the process, remove mappings for keys that no longer exist, and
add missing mappings for keys that do exist.

### [Bump agent version](https://github.com/appsignal/appsignal-nodejs/pull/841/commits/553a5b4b622c48e886db2587e8f9a3842ac8728e)

Update the agent to a version with support for NGINX metrics.


luismiramirez (approved):
OK for me when index type TS error from https://github.com/appsignal/appsignal-nodejs/blob/main/src/diagnose.ts is fixed.

src/__tests__/config.test.ts
```diff
@@ -39,7 +40,6 @@ describe("Configuration", () => {
       "content-length",
       "range"
     ],
-    sendEnvironmentMetadata: true,
```
tombruijn, line 42: Why is this config option removed as a default everywhere?

tombruijn:
I think the same span error data error as for the Ruby gem will occur in this repo, but I can't run it due to that type error at the moment https://github.com/appsignal/appsignal-ruby/pull/916/commits/3a79aac596dfe192f13254f9e2d45d88f804e642

src/__tests__/config.test.ts
unflxw: I was trying to ensure through typing that these mappings were consistent (that they mapped to config options that actually existed) and in the process I removed the "inconsistencies", the config options that were no longer declared in the `AppsignalOptions` type but still had a declared mapping. For example, there was a mapping for `instrumentPg` -- but `instrumentPg` is no longer a config option, so I removed the mapping.

For `sendEnvironmentMetadata`, I removed it because it was already undeclared on `AppsignalOptions` (meaning TS users already couldn't use it) and it is marked as deprecated in [our configuration options documentation page](https://docs.appsignal.com/nodejs/2.x/configuration/options.html#option-sendenvironmentmetadata), and I could not find any code in the integration that used it.

But looking at it now, it seems like [the agent does need the environment variable to be explicitly passed and set to true](https://github.com/appsignal/appsignal-agent/blob/99ae7938e880081ecfc252e7d3b87216143ba3b1/common/src/config.rs#L272-L275), or else [it won't schedule the environment metadata to be sent](https://github.com/appsignal/appsignal-agent/blob/99ae7938e880081ecfc252e7d3b87216143ba3b1/agent/src/worker/mod.rs#L105-L108).

So, what do we do? Options that come to mind:
- Un-deprecate the config option in the docs and add it back to the `AppsignalOptions` type
- Have the Node.js integration always pass it as true
- Have the agent default it to true

tombruijn: > it is marked as deprecated in [our configuration options documentation page](https://docs.appsignal.com/nodejs/2.x/configuration/options.html#option-sendenvironmentmetadata), and I could not find any code in the integration that used it.

@unflxw It shouldn't be deprecated. I don't know why it says that. It enables a thing in the agent and should be available as an option to turn it off.

> Un-deprecate the config option in the docs and add it back to the AppsignalOptions type

Yes.

tombruijn (approved):

tombruijn:
I've pushed the new agent update on the main branch, with the backtrace fixes. Please rebase :)

PR #867, opened 2023-02-21
Support formats in logger

thijsc:
Accept a third argument when creating a logger that sets the format that messages in logs will be parsed as. Bumps agent to 8d042e2

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -8,7 +8,7 @@ import { NoopMetrics, noopIntegrationLogger, noopLogger } from "./noops"
 import { demo } from "./demo"
 import { VERSION } from "./version"
 import { setParams, setSessionData } from "./helpers"
-import { BaseLogger, Logger, LoggerLevel } from "./logger"
+import { BaseLogger, Logger, LoggerFormat, LoggerLevel } from "./logger"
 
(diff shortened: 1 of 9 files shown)


src/client.ts
```diff
@@ -123,9 +123,13 @@ export class Client {
   /**
    * Global accessors for the AppSignal Logger API
    */
-  static logger(group: string, level?: LoggerLevel): Logger {
+  static logger(
+    group: string,
+    level: LoggerLevel = "info",
+    format: LoggerFormat = "plaintext"
```
tombruijn, line 129: We can turn this in a config object if that makes it easier to configure without having to specify defaults only to change the log format. Like keyword arguments in Ruby.

tombruijn (approved):

luismiramirez (approved):

src/client.ts
unflxw: I agree, I think we should do this. We need to keep compatibility with the existing signature, though. (It can be done with a [signature overload](https://www.typescriptlang.org/docs/handbook/2/functions.html#overload-signatures-and-the-implementation-signature))

thijsc: I will merge this as is, can you open an issue for this change if you want it?

PR #870, opened 2023-03-10
Add warnings to 2.x old packages

### LINKED ISSUE #869: Mark old packages as deprecated on npm
Release a new version, indentical to the last one, but with an update to the README, saying it's deprecated and point to the new Node.js docs.

## To do

- [x] See above description.

tombruijn:
Avoid confusion when people view our 2.x packages on npm when they're using 3.x or newer. Add a small warning to the top of the README to link people to the v3 docs.

This will require a patch release for all packages to have this update show up on npmjs.org.

Closes #869

The test suite fails for the latest Next.js, but that's irrelevant for this change.

### DIFF
--- a/packages/apollo-server/.changesets/add-warning-to-readme.md
+++ b/packages/apollo-server/.changesets/add-warning-to-readme.md
@@ -1,0 +1,6 @@
+---
+bump: "patch"
+type: "change"
+---
+
+Add warning to README about v2 and v3 compatibility.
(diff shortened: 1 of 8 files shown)


jeffkreeftmeijer (approved):

luismiramirez (approved):

unflxw (approved):
In addition, would it be valuable to [deprecate these packages](https://docs.npmjs.com/deprecating-and-undeprecating-packages-or-package-versions)?

tombruijn:
I don't know the exact impact of people still using a deprecated package. If they only see a message upon install, I'm okay with deprecating them on npm.

unflxw:
I think that's what happens -- I don't recall any other consequences when our package was using a deprecated package (#822)

PR #876, opened 2023-03-31
Update OpenTelemetry dependencies

unflxw:
The Next.js instrumentation requires more recent versions of the `@opentelemetry/api` and `@opentelemetry/core` libraries. The instrumentation libraries are often locked to the minor version, so they need to be bumped as well. Also we get bugfixes and such.


unflxw:
After having to upgrade to TypeScript 5.0 in order to solve some ~man-made horrors beyond my comprehension~ type declaration issues in our OpenTelemetry dependencies, the test failures that remain in the integration tests seem to be due to changes in the spans emitted by the instrumentation libraries. In other words, the integration test suite is working as expected!

I'll look into those on Monday. Marking this as a draft for now.

src/client.ts
```diff
@@ -376,7 +376,7 @@ export class Client {
       spanProcessor
     })
 
-    this.instrumentationsLoaded = sdk.start()
```
tombruijn, line 379: isn't this needed?
https://docs.appsignal.com/nodejs/3.x/instrumentation/instrumentations-load.html#main

unflxw: `sdk.start()` no longer returns a `Promise<void>`, just a good ol' `void`. I'm taking this to mean that the SDK is now loaded synchronously. I need to double-check that assumption before merging.

Before this function is called, the initialiser already sets `client.instrumentationsLoaded()` to `Promise.resolve()`. So `await this.instrumentationsLoaded()` now becomes an unnecessary no-op, but it won't break for users who have added it to their code.

If this is the case and there really is nothing that needs to be awaited on the SDK's initialisation, then we should probably remove the documentation for `instrumentationsLoaded` and deprecate it for removal in v4. We could also put it behind an accessor that emits a warning, telling users that they no longer need to call it.

tombruijn (approved):
Please remove the "[wip]" label from the commit if you intend to merge it as-is.

src/client.ts
unflxw: Looked at the change to the SDK start. Here's [the relevant PR](https://github.com/open-telemetry/opentelemetry-js/pull/3460) and [the issue it solved](https://github.com/open-telemetry/opentelemetry-js/issues/2912) -- the issue being the same one we were encountering, which we addressed by exposing the promise as `instrumentationsLoaded`. So the changes in this PR are correct.

I have therefore [created an issue for deprecating `instrumentationsLoaded`.](https://github.com/appsignal/appsignal-nodejs/issues/877)

PR #878, opened 2023-04-17
Do not crash on malformed error events

unflxw:
For some reason, if an error is thrown when rendering a Next.js template, the error will be forwarded to the back-end and reported to OpenTelemetry, but it will not have a type or a stacktrace. The extension, in turn, finds an `undefined` where a `string` should be, and throws an error accordingly.

I don't think we'd be able to do much with such an error further down the pipeline anyway, so if an exception event like that is encountered, consider it malformed and ignore it, not sending it to the extension.

### DIFF
--- a/src/span_processor.ts
+++ b/src/span_processor.ts
@@ -47,11 +47,17 @@ export class SpanProcessor implements OpenTelemetrySpanProcessor {
     errors.forEach(e => {
       const eventAttributes = e["attributes"] as any
 
-      opentelemetrySpan.setError(
-        eventAttributes["exception.type"],
-        eventAttributes["exception.message"],
-        eventAttributes["exception.stacktrace"]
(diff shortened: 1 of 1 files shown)


tombruijn (approved):
If it's only the stacktrace that's the problem, we could fall back on an empty value that doesn't error, but I don't know exactly what errors are thrown, so I'm trusting your judgement that we can ignore them.

unflxw:
@tombruijn Well, ideally we wouldn't ignore these errors. But without a stacktrace and an error type, they won't be very useful. I suppose we could default to sending "Error" as the type and an empty stacktrace.

tombruijn:
> But without a stacktrace and an error type, they won't be very useful.

With the error type also missing, two out of three things are not there, so let's ignore them.

Issue #882, opened 2023-04-26
Report on `appsignal.cjs` in the diagnose


unflxw:
The diagnose tool should look for this file at the locations where it's expected (`appsignal.cjs` or `src/appsignal.cjs`).

Ideally we'd like the configuration added to the `Appsignal` initialiser in this file to be reported in the diagnose report. To report on the configuration, the diagnose tool [creates a new configuration object](https://github.com/appsignal/appsignal-nodejs/blob/63b260268fdbb80eac83f8cfb53e77d7e54b53cf/src/diagnose.ts#L37), which does not have any values for the "initial" source.

Instead, the diagnose tool could attempt to require this `appsignal.cjs` file. When required, it will create an instance of `Appsignal`, which will write itself the global `Appsignal.client` object. We could then ask this global instance for its configuration object, and do the report based on that object.

Finally, since sometimes users don't have a plain `appsignal.cjs` file (for example, if they have an `appsignal.ts` file which ends up somewhere else as `dist/appsignal.js`) we could have a `--config` option so users can specify where this file actually is. Not sure if this is worth the effort.

## To do

- [x] Make the diagnose report on the presence/absence of `appsignal.cjs` and its contents https://github.com/appsignal/appsignal-nodejs/pull/884.
- [x] If the file exists, report on the config that is provided by it. https://github.com/appsignal/appsignal-nodejs/pull/885
- [ ] ~Add a `--config` option to the diagnose report to indicate the path to the config file.~

(This to do list is kind of a wish list -- if any of these items turn out to be too difficult, do the pragmatic thing and drop them)

tombruijn:
> Instead, the diagnose tool could attempt to require this appsignal.cjs file. When required, it will create an instance of Appsignal, which will write itself the global Appsignal.client object. We could then ask this global instance for its configuration object, and do the report based on that object.

We should avoid starting the AppSignal extension, which it does automatically in the client constructor.
We can set an env var like `APPSIGNAL_DIAGNOSE=true` and it won't start the extension, OpenTelemetry, minutely probes, etc. in that case. It only loads the config.

luismiramirez:
First PR checks for the presence of the file. Now I'm working on checking its contents and config values with Tom's proposal. If it gets too complicated/not-clean-enough I'll skip it and do the `--config` flag addition which should be more straightforward.

luismiramirez:
[...]

PR #884, opened 2023-05-09
Add the presence of the client file to diagnose

luismiramirez:
The diagnose report now checks on the presence of the `appsignal.cjs` file we require customers to initialize the AppSignal client when starting their apps.

Part of: https://github.com/appsignal/appsignal-nodejs/issues/882

### DIFF
--- a/src/__tests__/config.test.ts
+++ b/src/__tests__/config.test.ts
@@ -216,6 +216,42 @@ describe("Configuration", () => {
     })
   })
 
+  describe("clientFilePath", () => {
+    describe("when the client file exists in the root path", () => {
+      it("returns the path to the client file", () => {
+        jest.spyOn(fs, "existsSync").mockImplementation(givenPath => {
+          return givenPath === path.join(process.cwd(), "appsignal.cjs")
+        })
+        config = new Configuration({ name, pushApiKey })
+
+        expect(config.clientFilePath).toEqual(
+          path.join(process.cwd(), "appsignal.cjs")
+        )
+      })
+    })
+
+    describe("when the client file exists in the src path", () => {
+      it("returns the path to the client file", () => {
+        jest.spyOn(fs, "existsSync").mockImplementation(givenPath => {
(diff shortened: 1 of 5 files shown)


unflxw (approved):

src/diagnose.ts
```diff
@@ -157,6 +156,9 @@ export class DiagnoseTool {
       log_dir_path: {
         path: logFilePath ? path.dirname(logFilePath) : ""
       },
+      "appsignal.cjs": {
+        path: this.#config.clientFilePath || ""
```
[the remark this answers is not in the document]
luismiramirez: Yep! As we can't control what and how users write their `appsignal.cjs` code, I think it's more effective doing it by loading the code and then checking the config object.

src/__tests__/config.test.ts
```diff
@@ -216,6 +216,50 @@ describe("Configuration", () => {
...
+    })
+
+    describe("when the client file exists in the src path", () => {
+      it("returns the path to the client file", () => {
+        jest.spyOn(fs, "existsSync").mockImplementation(givenPath => {
+          if (givenPath === path.join(process.cwd(), "src", "appsignal.cjs")) {
+            return true
+          } else {
+            return false
+          }
```
tombruijn, line 244: I think this can be shorter to directly return the value from the if statement

src/__tests__/config.test.ts
```diff
@@ -216,6 +216,50 @@ describe("Configuration", () => {
...
 
+  describe("clientFilePath", () => {
+    describe("when the client file exists in the root path", () => {
+      it("returns the path to the client file", () => {
+        jest.spyOn(fs, "existsSync").mockImplementation(givenPath => {
+          if (givenPath === path.join(process.cwd(), "appsignal.cjs")) {
+            return true
+          } else {
+            return false
+          }
```
tombruijn, line 227: Same here as previous comment.

tombruijn (approved):

PR #885, opened 2023-05-10
Config client diagnose

luismiramirez:
### [Do not start extension if diagnose mode is enabled](https://github.com/appsignal/appsignal-nodejs/commit/b9fe462a2e8704c975a46b528123372ac5461dac)

When the `APPSIGNAL_DIAGNOSE` env var is present, don't start the extension from the client object. This will allow us to initialize the client during diagnose to check its configuration.

### [Load client file if present when running diagnose](https://github.com/appsignal/appsignal-nodejs/commit/90ea21a6402652cd552abbb48123cdc7a9422711)

The client file is loaded to take its values into account when running the diagnose script.

### [Make clientFilePath a static function](https://github.com/appsignal/appsignal-nodejs/pull/885/commits/26ee78f43e197297525e8161a5b9a154cd535dff) 

To know where the client file is stored, there's no need of
instantiating a Configuration object. This commit transforms the
`clientFilePath()` instance function into a static one.

Part of: https://github.com/appsignal/appsignal-nodejs/issues/882

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -146,11 +146,18 @@ export class Client {
     this.storeInGlobal()
 
     if (this.isActive) {
-      this.extension.start()
-      this.#metrics = new Metrics()
-      this.#sdk = this.initOpenTelemetry(
-        options.additionalInstrumentations || []
-      )
+      if (process.env._APPSIGNAL_DIAGNOSE === "true") {
+        Client.integrationLogger.info(
+          "Diagnose mode is enabled, not starting extension, SDK and probes"
+        )
+        this.#metrics = new NoopMetrics()
+      } else {
+        this.start()
+        this.#metrics = new Metrics()
+        this.#sdk = this.initOpenTelemetry(
+          options.additionalInstrumentations || []
+        )
+      }
     } else {
       this.#metrics = new NoopMetrics()
       console.error("AppSignal not starting, no valid configuration found")
@@ -181,11 +188,7 @@ export class Client {
    * directory.
    */
   public start(): void {
-    if (process.env["APPSIGNAL_DIAGNOSE"]) {
-      Client.integrationLogger.info(
-        "Diagnose mode is enabled, not starting extension"
-      )
-    } else if (this.config.isValid) {
+    if (this.config.isValid) {
       this.extension.start()
     } else {
       console.error("Not starting, no valid AppSignal configuration found")
--- a/src/client.ts
+++ b/src/client.ts
@@ -181,7 +181,11 @@ export class Client {
    * directory.
    */
   public start(): void {
-    if (this.config.isValid) {
+    if (process.env["APPSIGNAL_DIAGNOSE"]) {
+      Client.integrationLogger.info(
+        "Diagnose mode is enabled, not starting extension"
+      )
+    } else if (this.config.isValid) {
       this.extension.start()
     } else {
       console.error("Not starting, no valid AppSignal configuration found")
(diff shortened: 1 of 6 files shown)


src/client.ts
```diff
@@ -183,6 +183,8 @@ export class Client {
   public start(): void {
     if (this.config.isValid) {
       this.extension.start()
+    } else if (process.env["APPSIGNAL_DIAGNOSE"]) {
+      console.log("Diagnose mode is enabled, not starting extension")
```
unflxw, line 187: Should this maybe be logged through the integration logger?
```suggestion
      this.integrationLogger.log("Diagnose mode is enabled, not starting extension")
```
(not sure if this is how you access it, I don't remember)

unflxw (commented):
Does the test currently pass? 🤔 It seems to me that it should not be passing (because of the order of the conditions -- see the comment in the code)

luismiramirez:
>Does the test currently pass? 

Yes. I think something is not right with the agent in the test suite then. Rechecking

src/client.ts
```diff
@@ -183,6 +183,8 @@ export class Client {
   public start(): void {
     if (this.config.isValid) {
       this.extension.start()
+    } else if (process.env["APPSIGNAL_DIAGNOSE"]) {
```
tombruijn, line 186: I'm not sure where this env var gets set in the diagnose tool.

tombruijn (changes requested):

luismiramirez:
> Does the test currently pass? 🤔 It seems to me that it should not be passing (because of the order of the conditions -- see the comment in the code)

Well, turns out the `Client.start()` function is not used anywhere, the logic is never executed.

src/diagnose.ts
```diff
@@ -211,6 +214,21 @@ export class DiagnoseTool {
...
+  /**
+   * If it can load the client from the `appsignal.cjs` file, get the config
+   * object from the initialized client. Otherwise, return a default config object.
+   */
+  private getConfigObject(): Configuration {
+    // The file is required to execute the client initialization
+    // that stores the config object on the global object, making
+    // it available calling `Client.config` later.
+    if (Configuration.clientFilePath) {
+      require(Configuration.clientFilePath)
```
unflxw, line 226: This `require` could fail (for example, if there was a syntax error in the file) -- we should probably try-catch it.

src/diagnose.ts
```diff
@@ -34,7 +35,7 @@ export class DiagnoseTool {
   #extension: Extension
 
   constructor() {
-    this.#config = new Configuration({})
+    this.#config = this.getConfigObject()
```
unflxw, line 38: Calling `this.getConfigObject()` will in turn call `require(Configuration.clientFilePath)`, right? And it's for this require of the client file that we need `_APPSIGNAL_DIAGNOSE = true`, so that the client that we're requiring doesn't start the extension on initialisation.

But when the constructor is called here, `_APPSIGNAL_DIAGNOSE` hasn't been set yet -- that happens later, in the `generate()` method in line 48 of this file. So when the client file is required, the environment variable will not be set yet, and the extension will be started.

I think it would make sense to set the environment variable inside `getConfigObject` instead, also for readability -- that way the environment variable is set in the context where it's needed.

unflxw (approved):

src/client.ts
```diff
@@ -181,7 +181,11 @@ export class Client {
    * directory.
    */
   public start(): void {
-    if (this.config.isValid) {
+    if (process.env._APPSIGNAL_DIAGNOSE === "true") {
```
tombruijn, line 184: We can also move this check up to the constructor line 148. We don't need to start a lot of the other things there, like: the metrics, OpenTelemetry and the minutely probes.

tombruijn (approved):

PR #925, opened 2023-07-21
Add --config flag to diagnose command.

### LINKED ISSUE #887: Detect custom AppSignal client files from diagnose command
From a leftover of https://github.com/appsignal/appsignal-nodejs/issues/882

Even tho we don't recommend it, some users might initialize the AppSignal client with other methods instead of the recommended `appsignal.cjs`. If that happens, the diagnose command won't be able to detect the client initialization file.

We tried to use the `--config` flag, but with the current implementation, that means we need to share information between the diagnose command and the `Config` class which is not ideal. In order to make the `--config` flag to work, more significant changes are required.

## ToDo

- [ ] Provide users with the ability of telling where the custom AppSignal client initialization file is when using the diagnose command
- [ ] Check the presence of this file and load it as we do with `appsignal.cjs` or `src/appsignal.cjs`

luismiramirez:
Users can specify where they hold the AppSignal client config in their apps using the `--config` flag. This value if given will be taken into account to preload the file and to validate the config.

Closes #887

### DIFF
--- a/src/diagnose.ts
+++ b/src/diagnose.ts
@@ -282,11 +282,7 @@ export class DiagnoseTool {
     if (flagIndex !== -1 && flagIndex + 1 < process.argv.length) {
       const filePath = process.argv[flagIndex + 1]
 
-      if (path.isAbsolute(filePath)) {
-        return filePath
-      } else {
-        return path.resolve(filePath)
-      }
+      return path.resolve(filePath)
     }
   }
 
--- a/src/diagnose.ts
+++ b/src/diagnose.ts
@@ -158,8 +158,7 @@ export class DiagnoseTool {
         path: logFilePath ? path.dirname(logFilePath) : ""
       },
       "appsignal.cjs": {
-        path:
-          this.getCustomClientFilePath() || Configuration.clientFilePath || ""
+        path: this.clientFilePath() || ""
       },
       "appsignal.log": {
         path: logFilePath || "",
@@ -218,8 +217,7 @@ export class DiagnoseTool {
    * object from the initialized client. Otherwise, return a default config object.
    */
   private getConfigObject(): Configuration {
-    const clientFilePath =
-      this.getCustomClientFilePath() || Configuration.clientFilePath
+    const clientFilePath = this.clientFilePath()
     // The file is required to execute the client initialization
     // that stores the config object on the global object, making
     // it available calling `Client.config` later.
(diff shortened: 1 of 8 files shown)


tombruijn:
Made some additional fixes in #927

src/diagnose.ts
```diff
@@ -273,6 +276,20 @@ export class DiagnoseTool {
...
+    const flagIndex = process.argv.indexOf("--config")
+
+    if (flagIndex !== -1 && flagIndex + 1 < process.argv.length) {
+      const filePath = process.argv[flagIndex + 1]
+
+      if (!path.isAbsolute(filePath)) {
+        return filePath
+      } else {
+        return path.resolve(`../${filePath}`)
+      }
```
tombruijn, line 289:
```suggestion
      if (path.isAbsolute(filePath)) {
        return filePath
      } else {
        return path.resolve(filePath)
      }
```

If if-statement was in reverse: it making an absolute path even more absolute, and it not making a relative path an absolute path. I reversed the if-statement and made a change for the absolute path:

I'm not entirely sure why locally I had to specify `--config ../my_appsignal.cjs` for a custom relative file path, but here we don't. I guess because it's executed within the path of the script. But this change works.

tombruijn (changes requested):

src/diagnose.ts
luismiramirez: I could make it work in Docker using the redis-express app. I'll change it.

tombruijn (approved):

src/diagnose.ts
```diff
@@ -273,6 +276,20 @@ export class DiagnoseTool {
...
+    const flagIndex = process.argv.indexOf("--config")
+
+    if (flagIndex !== -1 && flagIndex + 1 < process.argv.length) {
+      const filePath = process.argv[flagIndex + 1]
+
+      if (path.isAbsolute(filePath)) {
+        return filePath
+      } else {
+        return path.resolve(filePath)
+      }
```
unflxw, line 289: I believe resolving an absolute path leaves it unchanged.

```suggestion
      return path.resolve(filePath)
```

unflxw (approved):

unflxw:
A thought: if a config file path is explicitly passed with the `--config` flag, but that path does not actually exist, perhaps we should exit with an error instead of going through with the diagnose?

tombruijn:
> A thought: if a config file path is explicitly passed with the --config flag, but that path does not actually exist, perhaps we should exit with an error instead of going through with the diagnose?

I guess that would make sense. It's a CLI option on for the diagnose so explicitly specifying a file that can't be loaded would be a CLI issue itself, not a config issue (where we detect it's not loaded but still make a report with it not loaded).

luismiramirez:
Added the existence check in the CLI -> https://github.com/appsignal/appsignal-javascript/pull/608

PR #939, opened 2023-08-31
Fix bogus platform mismatch error

### LINKED ISSUE #933: Bogus platform mismatch error on Alpine (musl)
When installing on Alpine, the install report writes that the platform that was installed is `linux-musl`. This is correct, as we're installing on Alpine.

But then, when the application is initialized, if the extension fails to load, the extension wrapper checks whether the install report's platform (`"linux-musl"`) matches the platform, as reported by Node's `process.platform` -- but Node's `process.platform` just reports `"linux"`.

Because of this, failures to load the extension will _always_ report a platform mismatch on Alpine, even if there is no such mismatch, misdirecting us and masking a more interesting failure to load the extension that would actually deserve looking into.

(A clue to this error not representing a true platform mismatch is that it is "backwards": it will say that the extension was installed for musl, but it's attempting to be ran on non-musl -- but the user will be attempting to run it on musl)

unflxw:
This seemed like it would be easy when I picked it up. Fixes #933.

---

### [Implement report path function only once](https://github.com/appsignal/appsignal-nodejs/commit/d0e8630e74ae1c0868309131741e0fa7f046a1ed)

Instead of having two implementations of the `reportPath` function
that must be kept in sync, implement it only once, in the JavaScript
scripts helpers (as it must be used by them) and then refer to it
from the TypeScript code by adding a `.d.ts` file that describes it.

### [Amend helper function descriptions](https://github.com/appsignal/appsignal-nodejs/commit/4092ebcdb7f87d8ea47e5844192f6516a16597e8)

I don't think that's what "and vice versa" means.

### [Fix musl platform comparison](https://github.com/appsignal/appsignal-nodejs/commit/537c15abfddeafd2f2b99436a7e9968f52def07e)

When the extension is installed in musl Linux, but the extension
fails to load for some other reason, the error displayed would
describe a bogus platform mismatch between musl and non-musl Linux.

Fix this by checking whether the current platform is musl Linux,
and expecting the corresponding platform target name to be found in
the install report. Do so by re-using the logic from the installation
report helpers.

### [Do not display unknown as architecture](https://github.com/appsignal/appsignal-nodejs/commit/85f422edf82bc1324445d349e62f6e74c905ca7b)

There is already an error that is emitted if the install report
could not be found. Let's use the generic error when that's the case
instead of hinting at a platform mismatch.

### [Fix build report tests](https://github.com/appsignal/appsignal-nodejs/commit/14e226f5e2c8f5e0b6511b26640056ac83e5cb32)

Since now `processTarget` does not call `hasMusl` across a module
boundary, it is not possible (to my knowledge) to mock the return
value of `hasMusl` when called by `processTarget`.

Instead, mock the `spawnSync` method of `child_process` to pretend
that `ldd` returned either `"musl"` or something else.

### DIFF
--- a/scripts/extension/__tests__/report.test.js
+++ b/scripts/extension/__tests__/report.test.js
@@ -1,8 +1,14 @@
 const originalProcess = process
 const originalProcessEnv = process.env
-const mockHelpers = { hasMusl: jest.fn() }
+const mockHelpers = { spawnSync: jest.fn() }
 
-jest.doMock("../support/helpers", () => mockHelpers)
+jest.doMock("child_process", () => {
+  const originalModule = jest.requireActual("child_process")
+  return {
+    ...originalModule,
+    ...mockHelpers
+  }
+})
 
 const { createBuildReport } = require("../report")
 
@@ -32,7 +38,8 @@ describe("muslOverride", () => {
   }
 
   function setHasMusl(value) {
-    mockHelpers.hasMusl.mockReturnValue(value)
+    const stderr = value ? "musl libc (arch)" : "something else"
+    mockHelpers.spawnSync.mockReturnValue({ stderr })
   }
 
   describe("with linux platform", () => {
--- a/src/__tests__/extension.failure.test.ts
+++ b/src/__tests__/extension.failure.test.ts
@@ -1,6 +1,9 @@
 import fs from "fs"
 import { Extension } from "../extension"
-import { reportPath } from "../../scripts/extension/support/helpers"
+import {
+  reportPath,
+  processTarget
+} from "../../scripts/extension/support/helpers"
 
 describe("Extension", () => {
   let ext: Extension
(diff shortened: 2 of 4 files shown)


scripts/extension/__tests__/report.test.js
```diff
@@ -32,7 +38,8 @@ describe("muslOverride", () => {
   }
 
   function setHasMusl(value) {
-    mockHelpers.hasMusl.mockReturnValue(value)
+    const stderr = value ? "musl" : "something else"
```
tombruijn, line 41:
```suggestion
    const stderr = value ? "musl libc (arch)" : "something else"
```
Make so that it reflects the real world closer and it needs the substring check, and not `==`, for testing completeness.

tombruijn (approved):

luismiramirez (approved):

PR #942, opened 2023-09-28
add disableOpenTelemetrySDK option

nirev:
When setting it to true, Appsignal will not start OpenTelemetry's NodeSDK
This is useful when there's a need to start and configure NodeSDK manually and plug
Appsignal SpanProcessor to it later.

Something like

```ts
import { ProxyTracerProvider, trace } from '@opentelemetry/api'
import { BasicTracerProvider } from '@opentelemetry/sdk-trace-node'
import { SpanProcessor as AppsignalSpanProcessor } from '@appsignal/nodejs'
import { Appsignal } from '@appsignal/nodejs'
import { SpanProcessor } from '@opentelemetry/sdk-trace-base'

  active: true,
  name: 'service',
  pushApiKey: process.env.APPSIGNAL_PUSH_API_KEY,
  environment: 'production',
  disableOpenTelemetrySDK: true,
})

if (appsignal.config.isValid) {
  const proxyTracerProvider = trace.getTracerProvider() as ProxyTracerProvider
  const nodeTracerProvider = proxyTracerProvider.getDelegate() as BasicTracerProvider

  const processor = new AppsignalSpanProcessor(appsignal) as unknown as SpanProcessor
  nodeTracerProvider.addSpanProcessor(processor)
}
```


nirev:
@unflxw Tagging you, as I asked about this is Appsignal's support chat

Thank you!

tombruijn (commented):
I don't know the context of this PR, but this would break some stuff, I think, if you expect it to work the same as our standard instrumentation.

We configure the [OpenTelemetry instrumentation libraries with additional config](https://github.com/appsignal/appsignal-nodejs/blob/e1659fe47d93054fd2276d3217e7e6aa093f6fb5/src/client.ts#L266-L342). Something we don't expose to be called separately later.

unflxw:
Thanks @nirev! I've continued the work over at #952. Closing.

PR #952, opened 2023-10-20
Add `initializeOpentelemetrySdk` config option

unflxw:
Taking over #942 to add the changes discussed there.

Add an `initializeOpentelemetrySdk` config option. Only initialize the OpenTelemetry SDK when this option is set to `true`. This allows experienced OpenTelemetry users to configure the SDK to their liking.


.changesets/add-initialize-opentelemetry-sdk-config-option.md
```diff
@@ -0,0 +1,31 @@
...
+
+```js
+import { SpanProcessor, Appsignal } from "@appsignal/nodejs";
+// or: const { SpanProcessor, Appsignal } = require("@appsignal/nodejs")
+
+const sdk = new NodeSDK({
+  spanProcessor: new SpanProcessor(Appsignal.client)
+});
+
+sdk.start()
```
tombruijn, line 25: I brought up a point in the other PR here: https://github.com/appsignal/appsignal-nodejs/pull/942#pullrequestreview-1652354521

> this would break some stuff, I think, if you expect it to work the same as our standard instrumentation.
>
> We configure the [OpenTelemetry instrumentation libraries with additional config](https://github.com/appsignal/appsignal-nodejs/blob/e1659fe47d93054fd2276d3217e7e6aa093f6fb5/src/client.ts#L266-L342). Something we don't expose to be called separately later.

This change would allow you to start the SDK and add AppSignal as a span processor, but doesn't add the instrumentations with their AppSignal specific config. Without these configs the reported data will arrive in ways we don't expect and could have missing or weird data reporting.

Do we expect people to configure this themselves that matches what we expect, or should we also expose a way to add the default instrumentations and their configs?

unflxw: I was expecting people to do it manually (hence the "look at the source code" at the bottom) but now that I think about it a bit more, it should be easy to expose an `.instrumentations` method in the client that gives you the same list of instrumentations that it would provide to the SDK. I'll have a go at implementing that.

unflxw: Done! Let me know what you think. I called it `.opentelemetryInstrumentations` for some additional "don't touch this if you don't know what OpenTelemetry is" vibes.

nirev:
Thank you, @unflxw! This is much better than my initial attempt :bow:

tombruijn (approved):

luismiramirez (approved):

jeffkreeftmeijer (approved):

PR #954, opened 2023-10-24
Fix Express error handler signature error

### LINKED ISSUE #953: `express-redis` integration test failure
[Something something types](https://appsignal.semaphoreci.com/jobs/58dde20c-7bc0-48d7-bc9c-db9c090761df). Might indicate an actual error, if our Express error handler's signature is no longer conformant with the expected type somehow. Note that the type declarations for Express are maintained as a separate `@types/express` package.

luismiramirez:
Fixed an error thrown when using the Express error handler that depended on which types package version for Express is used.

We've tried to fix the types but couldn't get to make it work due to how the interfaces in `@types/express-serve-static-core` are. We tried to replicate the exact same types allowed in the `IRouterHandler` interface without any success.

The return type of the error handler cannot be an `ExpressErrorHandler` anymore as it doesn't comply with the latest types defined by the packages.

This commit removes the return type of the error handler and makes the `req` and `res` argument to be of `any` type to stop the error from happening and to hopefully keep the handler compatible in future possible patches of the types packages.

Closes #953

### DIFF
--- a/.changesets/fix-types-compatibility-for-the-express-error-handler.md
+++ b/.changesets/fix-types-compatibility-for-the-express-error-handler.md
@@ -1,0 +1,6 @@
+---
+bump: "patch"
+type: "fix"
+---
+
+Fix a TypeScript types compatibility error upon app compilation when using the AppSignal Express error handler.
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@appsignal/nodejs",
-  "version": "3.0.23",
+  "version": "3.0.24",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@appsignal/nodejs",
-      "version": "3.0.23",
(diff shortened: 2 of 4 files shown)


.changesets/fix-types-compatibility-for-the-express-error-handler.md
```diff
@@ -0,0 +1,6 @@
+---
+bump: "patch"
+type: "fix"
+---
+
+Fix types compatibility for the Express error handler
```
tombruijn, line 6:
```suggestion
Fix a TypeScript types compatibility error upon app compilation when using the AppSignal Express error handler.
```

tombruijn (approved):

unflxw (approved):

tombruijn:
> Do we want to keep the changes to the `package.lock`? (Not opposed to it, just checking if it's intentional)

Yeah we should. That file always lags behind a bit.

PR #965, opened 2023-12-04
Clean up more extension files for macOS

tombruijn:
When running `npm run clean:ext` some files are left behind on macOS. Add these files to make sure it's clean after running this command.

[skip changeset]
[skip review]


tombruijn:
ci fixed in https://github.com/appsignal/appsignal-nodejs/pull/966

PR #978, opened 2024-01-31
Metric exporter

### LINKED ISSUE #909: Implement OpenTelemetry metrics exporter
~We could implement an exporter for OpenTelemetry metrics that calls the extension's metrics functions, similar to the existing OpenTelemetry span exporter.~ **EDIT:** We could also use the default OTel HTTP exporter with the existing HTTP metrics -- in fact we probably will want to do that, as a first step towards making the Node.js work the same way as Python.

 This would allow us to pick up metrics from the OpenTelemetry instrumentations, and in turn implement magic dashboards for those metrics. Ideally this would allow us to restore the Postgres dashboard that existed in 2.x -- see appsignal/public_config#49 for details.

## To do

- [x] Implement OpenTelemetry metrics exporter in Node.js
- [ ] Add back metrics previously lost for webvitals and postgres (create separate issues for it)
- [ ] Document the metrics support

luismiramirez:
[Add metric exporter](https://github.com/appsignal/appsignal-nodejs/commit/c22e3c236e91b92c2fa546df3731dc835e907e4a) 

Initialize OpenTelemetry's periodic metric exporter to send regular
metric data to the Agent OpenTelemetry metrics endpoint. It also enables
users to send their own OpenTelemetry metrics using compatible metrics
providers.

[Add enable OpenTelemetry HTTP config option](https://github.com/appsignal/appsignal-nodejs/commit/02d20103c6a6860dbd935a98be10929aea9d5b63) 

Make the OpenTelemetry HTTP config option available. It defaults to
`true`. This is required by the metrics endpoint to be able to send the
data somewhere.

[Enable metric exporter by config](https://github.com/appsignal/appsignal-nodejs/commit/2013815ec58b811a2979ab5d26b0d224a83bffa9) 

The OTel metric exporter is only enabled when the
`enableOpentelemetryHttp` config var is set to true. As metrics are sent
to the agent's HTTP server, if it's disabled, there's no point on
polluting the logs and sending unlimited requests to an unavailable
service.

[Fix type OTel dependency type issue](https://github.com/appsignal/appsignal-nodejs/commit/e7cee0fcacf1a958ea47f1f58c8d514dba31c93d) 

See: https://github.com/open-telemetry/opentelemetry-js/issues/3580#issuecomment-1701157270

[Fix boolean config options set to false](https://github.com/appsignal/appsignal-nodejs/commit/28f31bb30a0a04a8f5d7433c5a121f9c191e1489) 

When a boolean config option was set to false when initializing the
client and it wasn't its default value, it wouldn't work. This was due
to they way we checked the existence of values when setting the private
env vars. As it was a regular `if`, a `false` value would always be
`false` and the transformation skipped.

This commit checks if the present config value is of boolean type and
transforms it into a string nevermind its value.

---
Closes #909


luismiramirez:
Forgot about diagnose tests. Working on it.

src/config.ts
```diff
@@ -205,6 +206,10 @@ export class Configuration {
         process.env[k] = current.join(",")
       }
 
+      if (typeof current === "boolean") {
+        process.env[k] = String(current)
+      }
+
```
tombruijn, line 212: The check below this one would set it again if `true` (which wouldn't be a problem), or a non-boolean (which could be a problem), no?
We could make it an `else if (current)` statement part of the `if (typeof current === "boolean")` one.

src/client.ts
```diff
@@ -389,9 +391,20 @@ export class Client {
 
     const instrumentations = this.opentelemetryInstrumentations()
 
+    // Metrics initialization
+    let metricReader
+    if (this.config.data["enableOpentelemetryHttp"]) {
+      metricReader = new PeriodicExportingMetricReader({
+        exporter: new OTLPMetricExporter({
+          url: "http://localhost:8099/v1/metrics"
```
tombruijn, line 399: It could be useful to implement the [`opentelemetry_port` option](https://docs.appsignal.com/python/configuration/options.html#option-opentelemetry_port) for this, so it is customizable, but it's not required.

.changesets/add-metric-exporter.md
```diff
@@ -0,0 +1,7 @@
+---
+bump: "minor"
+type: "add"
+---
+
+With the OpenTelemetry metric exporter initialized by default, the integration will send default metrics from the instrumentations,
+and also allows you to send your custom metrics to AppSignal to build your own dashboards.
```
tombruijn, line 7:
```suggestion
The OpenTelemetry metric exporter is now initialized by default. The integration will send default metrics from the instrumentations, and also allows you to send your custom metrics using OpenTelemetry to AppSignal to build your own dashboards.
```

.changesets/fix-boolean-config-options.md
```diff
@@ -0,0 +1,6 @@
+---
+bump: "patch"
+type: "fix"
+---
+
+Boolean config options are now set to `false` if required when setting them up from the client initialization.
```
tombruijn, line 6:
```suggestion
Fix issue with boolean config options not being set to `false` properly for the agent configuration.
```

tombruijn (approved):

src/config/options.ts
```diff
@@ -28,6 +28,7 @@ export type AppsignalOptions = {
   logLevel: string
   loggingEndpoint: string
   name: string
+  opentelemetryPort: string | number
```
unflxw, line 31: I was wondering why this was declared as `string | number`, because I thought `number` would make sense for this, but then I saw that it's also declared as `str | int` [in the Python integration](https://github.com/appsignal/appsignal-python/blob/main/src/appsignal/config.py#L51).

Then I kept looking. 🙃 A few lines below in this file, `statsdPort` is declared as `string` only. But the [stand-alone agent TOML configuration file](https://github.com/appsignal/appsignal-agent/blob/e6dc07f98a2de165e0bf5a40ecaba9b7044bc88f/agent/src/standalone_config.rs#L70-L79) expects a number, not a string. And [the Ruby configuration option docs](https://docs.appsignal.com/ruby/configuration/options.html#option-statsd_port) claims it should be a number.

I think we should aim to unify this across integrations -- it makes for a confusing experience for it to be typed differently in each one.

But for this PR, I think we should make `opentelemetryPort` and `statsdPort` have the same type, whether that's `string` or `number` or `string | number`.

package.json
```diff
@@ -29,7 +29,8 @@
     "@opentelemetry/instrumentation-redis": ">= 0.35.4 < 0.36",
     "@opentelemetry/instrumentation-redis-4": ">= 0.35.5 < 0.36",
     "@opentelemetry/instrumentation-restify": ">= 0.34.2 < 0.35",
-    "@opentelemetry/sdk-node": ">= 0.45.0 < 0.46",
+    "@opentelemetry/sdk-node": "^0.47",
+    "@opentelemetry/exporter-metrics-otlp-proto": "^0.47.0",
     "@prisma/instrumentation": ">= 5.4.0 < 5.6.0",
```
unflxw, line 34: And we're back in OpenTelemetry dependency hell! It feels like our second home at this point.

```
> npm list @opentelemetry/api
@appsignal/nodejs@3.0.30 /Users/nlapresta/Code/appsignal-nodejs
├── @opentelemetry/api@1.7.0
...
└─┬ @prisma/instrumentation@5.5.2
  ├── @opentelemetry/api@1.6.0
```

Reminder that, due to some mind-blowing OpenTelemetry API overengineering, these two versions of `@opentelemetry/api` won't be able to share global state with each other -- or one of them will read the other one's global state, but the other one will refuse to read it, or something like that. I don't remember the details. Anyway, it results in trace context being lost for instrumentations using one of the two versions. Fun!

Updating the version bounds on `@prisma/instrumentation` to [a version that includes this commit](https://github.com/prisma/prisma/commit/249341f6cec6e0edec187d1b28cef0999e223f30) should fix the issue:

```suggestion
"@prisma/instrumentation": ">= 5.6.0 < 5.10.0",
```

unflxw: By the way @luismiramirez, this is why the other bounds were specified so strictly -- to make sure we didn't accidentally install a version with a different `@opentelemetry/api` version. 

Could you change the bounds on `@opentelemetry/api`, `@opentelemetry/instrumentation-http` and `@opentelemetry/sdk-node` back to a strict version range on the current minor version? I know it sucks, but it's the solution we have to this problem at the moment. Otherwise, the moment that `@opentelemetry/api@1.8.0` is released, all hell breaks loose.

luismiramirez: I think I got them all right. The rebase was a bit of a mess. Could you re-check, please?

unflxw: The `@prisma/instrumentation` bound didn't change -- maybe that got lost on the rebase?

After changing the bound, the `package-lock.json` needs to be updated using `npm update`.

jeffkreeftmeijer (approved):

unflxw (approved):
Dependency hell averted!

PR #1004, opened 2024-03-19
Explicitly parse configuration variables by type

jeffkreeftmeijer:
This patch adds explicit configuration types, which are looped over when reading the environment variables to make sure each configuration value gets a type of value it expects.


src/__tests__/config.test.ts
```diff
@@ -143,6 +145,59 @@ describe("Configuration", () => {
...
+          ...expectedDefaultConfig,
+          ...envOptions
+        }
+        config = new Configuration({})
+
+        expect(config.data).toEqual(expectedConfig)
+      })
+    })
+
+    describe("withAPPSIGNAL_DISABLE_DEFAULT_INSTRUMENTATIONS set to true", () => {
```
unflxw, line 170: nit: typo

```suggestion
    describe("with APPSIGNAL_DISABLE_DEFAULT_INSTRUMENTATIONS set to true", () => {
```

src/config.ts
```diff
@@ -169,16 +176,39 @@ export class Configuration {
...
+        conf[ENV_TO_KEY_MAPPING[k]] = current
+      }
+    })
+
+    BOOL_KEYS.forEach(k => {
+      const current = process.env[k]
+
+      if (current) {
+        conf[ENV_TO_KEY_MAPPING[k]] = current == "true"
+      }
```
unflxw, line 192: Perhaps use the same logic here as in `LIST_OR_BOOL_KEYS`? Meaning, if it's not either `true` or `false`, don't set it.

That is what we do [in the Python instrumentation](https://github.com/appsignal/appsignal-python/blob/main/src/appsignal/config.py#L311-L321).

```suggestion
      if (current == "true") {
        conf[ENV_TO_KEY_MAPPING[k]] = true
      } else if (current == "false") {
        conf[ENV_TO_KEY_MAPPING[k]] = false
      }
```

luismiramirez (approved):

src/config/configmap.ts
```diff
@@ -114,3 +114,52 @@ export const JS_TO_RUBY_MAPPING: Record<keyof AppsignalOptions, string> = {
   statsdPort: "statsd_port",
   workingDirectoryPath: "working_directory_path"
 }
+
+export const BOOL_KEYS: string[] = [
```
unflxw, line 118: Nit: with a bit of TypeScript magic, we can ensure the values here are existing keys of `ENV_TO_KEY_MAPPING`, preventing typos.

First, we change `ENV_TO_KEY_MAPPING` from being declared as a specific type:

```ts
export const ENV_TO_KEY_MAPPING: Record<string, keyof AppsignalOptions> = {...}
```

To being constrained to that type, but its actual type being inferred by TypeScript:

```ts
export const ENV_TO_KEY_MAPPING = {...} satisfies Record<string, keyof AppsignalOptions>
```

This means that its type is inferred as having the specific keys in the `ENV_TO_KEY_MAPPING` literal.

And then, we can declare `BOOL_KEYS` (and `STRING_KEYS` and so on) as follows:

```ts
export const BOOL_KEYS: (keyof typeof ENV_TO_KEY_MAPPING)[] = [...]

unflxw (approved):
All look good! I left some comments to consider but none of them are deal-breakers.

One last nit: I really like [this test in the Ruby integration](https://github.com/appsignal/appsignal-ruby/blob/5819cfdf3c44d00693e781b5d16fda44eecfd2b2/spec/lib/appsignal/config_spec.rb#L3-L13), where we ensure that every known environment variable from `ENV_TO_KEY_MAPPING` is present in one (and only one) list of `STRING_KEYS`, `BOOL_KEYS` and so on.

src/config/configmap.ts
jeffkreeftmeijer: Woudn't doing this negate the need of having a separate test?

I'm not against doing this, but let's have a separate pull request for it. I'd like to push this patch out. :)

unflxw: Go ahead! As I said, none of this is a blocker.

I don't think this negates the need for the test because this only checks that the keys are in the list, not that each key is only in one list.

PR #1029, opened 2024-04-12
Add OpenTelemetry RabbitMQ support

jeffkreeftmeijer:
This patch adds support for `@opentelemetry/instrumentation-amqplib` by including it in the default instrumentation list. It comes with a rest case named `express-rabbitmq`, which is based on the already existing `express-redis`.

To do
===

- [x] Add instrumentation
- [x] Add integration test
- [x] Add to an appsignal/test-setups test setup (https://github.com/appsignal/test-setups/pull/202)
- [x] Add extractor in the agent (https://github.com/appsignal/appsignal-agent/pull/1132)
- [x] Write docs for it (https://github.com/appsignal/appsignal-docs/pull/1883)


tombruijn:
Unassigning myself as a reviewer until I can test it in a test app.

src/client.ts
```diff
@@ -45,6 +46,7 @@ import {
 import { SpanProcessor, TestModeSpanProcessor } from "./span_processor"
 
 const DefaultInstrumentations = {
+  "@opentelemetry/instrumentation-amqp": AmqplibInstrumentation,
```
unflxw, line 49: Let's make this match the name of the actual instrumentation package.

```suggestion
  "@opentelemetry/instrumentation-amqplib": AmqplibInstrumentation,
```

test/express-rabbitmq/tests/spec/app_spec.rb
```diff
@@ -0,0 +1,82 @@
+# frozen_string_literal: true
+
+RSpec.describe "Rabbitm app" do
```
unflxw, line 3:
```suggestion
RSpec.describe "RabbitMQ app" do
```

(is this what it was supposed to say?)

unflxw (approved):
As mentioned before on conversations about this task, I really think we should amend the extractor to add `appsignal.ignore_trace_if_root` before releasing it. Otherwise these spans will become performance actions in uninstrumented contexts.

PR #1050, opened 2024-05-07
Add missing required transitive dependencies

### LINKED ISSUE #1037: Fix Yarn and pnpm support by declaring required transitive dependencies
We mostly test the AppSignal for Node.js integration with npm, which lets you do _whatever_ in terms of dependency use.

When `@appsignal/nodejs` attempts to `require("@opentelemetry/sdk-metrics")`, which is not a declared dependency of `@appsignal/nodejs`, but it is a transitive dependency (via `@opentelemetry/sdk-node`), npm complies by picking _any version of `@opentelemetry/sdk-node` it can find in the dependency tree_.

This behaviour can be problematic. Less chaos-friendly package managers, such as Yarn and pnpm, do not allow requiring transitive dependencies, and throw an error instead. Unfortunately, because we do the above, that means that AppSignal is not compatible with Yarn and pnpm.

Fortunately, there's no reason why we have to require undeclared transitive dependencies. We should declare `@opentelemetry/sdk-metrics` (and any other undeclared-but-required dependencies) as dependencies or peer dependencies of our package.

Related docs change: https://github.com/appsignal/appsignal-docs/pull/1860

unflxw:
Declare the transitive dependencies that our code already imports as _de facto_ dependencies in our `package.json`. Update the `package-lock.json` accordingly.

For `@opentelemetry/instrumentation`, which is only required for a type annotation, use `import type` instead.

Fixes #1037. [skip changeset]

### DIFF
--- a/.changesets/improve-compatibility-with-alternative-package-managers.md
+++ b/.changesets/improve-compatibility-with-alternative-package-managers.md
@@ -1,0 +1,6 @@
+---
+bump: "patch"
+type: "fix"
+---
+
+Improve compatibility with alternative package managers. Fix an issue where AppSignal would fail to install with Yarn when using package managers that require transitive dependencies to be declared.
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,16 +1,17 @@
 {
   "name": "@appsignal/nodejs",
-  "version": "3.4.1",
-  "lockfileVersion": 3,
+  "version": "3.4.2",
+  "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@appsignal/nodejs",
-      "version": "3.4.1",
+      "version": "3.4.2",
       "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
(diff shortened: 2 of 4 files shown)


luismiramirez (approved):

unflxw:
I can confirm that this works with the latest Yarn version (in that it removes the error reported by the customer during the AppSignal build, allowing the build to succeed with a reasonable looking -- I have not tested further)

tombruijn (approved):
You sure you don't want to add a changeset so you can ship it and people know we support it again?

unflxw:
@tombruijn I think you're right, I'll add a changeset of some sort, yes.

The thing is, I don't want us to proclaim in the docs again that we support Yarn or pnpm, though, like we used to do before this issue. Realistically, we're not doing any testing with those tools.

But it would be nice to say we've made some improvements to the compatibility of our package with alternative package managers like Yarn or pnpm. Give them the same treatment we currently give to WSL: we don't _explicitly_ support it, but if it works for you, that's great.

Issue #960, opened 2023-11-13
AppSignal errors on Windows when set to active


unflxw:
AppSignal should print an error message when on Windows, warning the user that AppSignal is not active -- but it should not look like it crashed. The error seen here about the install report, while not really a crash in itself, could be mistaken for one.

```
[appsignal][ERROR] Unable to fetch install report: Error: ENOENT: no such file or directory, open 'C:\dev\predium\apps\backend\node_modules\@appsignal\nodejs\ext\install.report'
at Object.openSync (node:fs:590:3)
at Object.readFileSync (node:fs:458:35)
at fetchInstalledArch (C:\dev\predium\apps\backend\node_modules\@appsignal\nodejs\dist\extension_wrapper.js:8:40)
at Object.<anonymous> (C:\dev\predium\apps\backend\node_modules\@appsignal\nodejs\dist\extension_wrapper.js:22:42)
at Module._compile (node:internal/modules/cjs/loader:1165:14)
at Object.Module._extensions..js (node:internal/modules/cjs/loader:1219:10)
at Module.load (node:internal/modules/cjs/loader:1043:32)
at Function.Module._load (node:internal/modules/cjs/loader:878:12)
at Module.require (node:internal/modules/cjs/loader:1067:19)
at require (node:internal/modules/cjs/helpers:103:18) {
errno: -4058,
syscall: 'open',
code: 'ENOENT',
path: 'C:\\dev\\predium\\apps\\backend\\node_modules\\@appsignal\\nodejs\\ext\\install.report'
}
[appsignal][ERROR] The AppSignal extension was installed for architecture 'unknown-unknown', but the current architecture is 'x64-win32'. Please reinstall the AppSignal package on the host the app is started.
```

PR #1051, opened 2024-05-07
Only report extension load errors on start

unflxw:
Before this change, extension load errors would be reported whenever
`@appsignal/nodejs` or the `client.ts` file was imported. Now they
will only be reported whenever `Client` fails to initialise.

This fixes an issue where the extension load errors would be reported
when running `npm install` in the `appsignal-nodejs` folder.

Rearrange the way that the client is initialised. Remove the
`client.start()` method, as starting is done automatically during
initialisation. Use early returns to clearly distinguish the client
initialisation failure scenarios. Emit distinct error messages for
different configuration-related failures.

Change the `Client.isActive` method to reflect whether the client
was actually initialised, rather than whether several factors in
the configuration should mean that the client was initialised.

This is an improvement to https://github.com/appsignal/appsignal-nodejs/issues/960, as AppSignal will no longer show an error about the extension not loading when `active` is set to `false`. It would still be nice to have a dedicated, non-scary Windows error.

### DIFF
--- a/.changesets/improve-error-reporting-during-initialisation.md
+++ b/.changesets/improve-error-reporting-during-initialisation.md
@@ -1,0 +1,10 @@
+---
+bump: "patch"
+type: "change"
+---
+
+### Improve error reporting during initialisation
+
+Do not report an error with the extension installation when AppSignal is imported -- instead, report it when attempting to initialise AppSignal. Do not report an error with the extension if AppSignal is not configured to be active.
+
+When AppSignal does not start due to its configuration (`active` is set to `false`, or the push API key is missing) report the specific reason why.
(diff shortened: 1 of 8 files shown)


tombruijn:
I'm okay with this change but you need to update the tests for this.

unflxw:
@tombruijn Yeah, the tests are correct to be failing -- now the message is not emitted at all. I'll see what I can do to fix it.

unflxw:
@tombruijn It's been fixed now -- but I've also changed a lot of other stuff in the process. I couldn't understand what the right place to emit these warnings in the client initialisation was, so I ended up rearranging it significantly to figure it out.

tombruijn (approved):

tombruijn:
Worth it to add a changeset .

PR #1064, opened 2024-06-14
Pass timestamp on OpenTelemetry span close

unflxw:
When the SpanProcessor closes a span, before this commit, it would call `span.close()`, which would call the `appsignal_close_span` function in the extension, which would override the end time passed as an argument when calling `createOpenTelemetrySpan`.

To fix this, pass the end time provided by OpenTelemetry _again_ to `span.close()` when closing the span, re-wiring it so that, if a timestamp is passed, it calls `appsignal_close_span_with_timestamp` instead. This function already existed in the extension, but a NAPI bridge for it needed to be created.

### DIFF
--- a/.changesets/fix-incorrect-span-durations.md
+++ b/.changesets/fix-incorrect-span-durations.md
@@ -1,0 +1,6 @@
+---
+bump: patch
+type: fix
+---
+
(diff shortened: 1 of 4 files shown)


luismiramirez (approved):

luismiramirez:
Related intercom conversation: https://app.intercom.com/a/inbox/yzor8gyw/inbox/conversation/16410700333561

PR #1123, opened 2024-09-16
Implement heartbeat check-ins

unflxw:
Part of https://github.com/appsignal/integration-guide/issues/169.

Add an `Appsignal.checkIn.heartbeat` helper that emits a single heartbeat for the check-in identifier given.

When given `{continuous: true}` as the second argument, it starts an interval that emits a heartbeat every thirty seconds. This is a convenience method for the use case where the heartbeat is only meant as a check that the process is alive.

Split functions that deal with different event kinds out of the scheduler and test them independently.

### DIFF
--- a/src/__tests__/check_in.test.ts
+++ b/src/__tests__/check_in.test.ts
@@ -1,21 +1,20 @@
 import nock from "nock"
-import { cron, Cron, Event, EventKind } from "../check_in"
-import {
-  scheduler,
-  resetScheduler,
-  setDebounceTime,
-  resetDebounceTime,
-  debounceTime
-} from "../check_in/scheduler"
+import { cron, Cron, heartbeat } from "../check_in"
+import { Event, EventKind } from "../check_in/event"
+import { scheduler, resetScheduler, debounceTime } from "../check_in/scheduler"
 import { Client, Options } from "../client"
 import {
-  heartbeat,
+  heartbeat as deprecatedHeartbeat,
   Heartbeat,
   heartbeatClassWarnOnce,
   heartbeatHelperWarnOnce
 } from "../heartbeat"
 import { ndjsonParse } from "../utils"
 import type { InternalLogger } from "../internal_logger"
+import {
+  heartbeatInterval,
+  killContinuousHeartbeats
+} from "../check_in/heartbeat"
 
 const DEFAULT_CLIENT_CONFIG: Partial<Options> = {
   active: true,
@@ -107,7 +106,7 @@ function spyOnInternalLogger(
   return spies as Required<typeof spies>
 }
 
-describe("checkIn.Cron", () => {
+describe("Check-ins", () => {
   let client: Client
   let theCron: Cron
   let requests: Request[]
@@ -122,7 +121,7 @@ describe("checkIn.Cron", () => {
 
   beforeEach(async () => {
     await resetScheduler()
-    resetDebounceTime()
+    debounceTime.reset()
 
     client = new Client(DEFAULT_CLIENT_CONFIG)
 
@@ -137,7 +136,7 @@ describe("checkIn.Cron", () => {
 
   afterAll(async () => {
     await resetScheduler()
-    resetDebounceTime()
+    debounceTime.reset()
     nock.restore()
   })
 
@@ -229,7 +228,7 @@ describe("checkIn.Cron", () => {
       // These tests will mock requests one by one, so that
       // they can await their responses.
       nock.cleanAll()
-      setDebounceTime(() => 20)
+      debounceTime.set(() => 20)
     })
 
     it("transmits events close to each other in time in a single request", async () => {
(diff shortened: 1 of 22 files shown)


src/check_in/heartbeat.ts
```diff
@@ -0,0 +1,18 @@
...
+  () => HEARTBEAT_INTERVAL_MILLISECONDS
+)
+
+let continuousHeartbeats: NodeJS.Timeout[] = []
+
+export function addContinuousHeartbeat(interval: NodeJS.Timeout) {
+  continuousHeartbeats.push(interval)
+}
+
+export function killContinuousHeartbeats() {
```
tombruijn, line 15: Can we mark them as private with a comment so that when people see them being auto suggested by their editor they understand they shouldn't use it?

src/__tests__/check_in.test.ts
```diff
@@ -493,13 +492,63 @@ describe("checkIn.Cron", () => {
...
+      ])
+    })
+
+    describe("with the `continuous: true` option", () => {
+      it("sends heartbeat events continuously", async () => {
+        debounceTime.set(() => 20)
+
+        heartbeat("test-heartbeat", { continuous: true })
+
+        await sleep(60)
```
tombruijn, line 528: Could we speed this up by mocking the sleep time value in the scheduler?

tombruijn (approved):

src/__tests__/check_in.test.ts
unflxw: We do that ([for the heartbeat here](https://github.com/appsignal/appsignal-nodejs/pull/1123/files/839073e8c3a13956e7ec5019039095e15a7cf402#diff-2ef3a10248c9f7c4a782cdefe5f5f5ccbef615d8a55e2303a170bc8e47279f82R496-R498) and [for the scheduler's debounce here](https://github.com/appsignal/appsignal-nodejs/pull/1123/files/839073e8c3a13956e7ec5019039095e15a7cf402#diff-2ef3a10248c9f7c4a782cdefe5f5f5ccbef615d8a55e2303a170bc8e47279f82R524))

Note that this is 60 milliseconds, not 60 seconds!

src/check_in/heartbeat.ts
unflxw: I'll take a look and see if that's possible, as that would be ideal. They are in this separate `heartbeat.ts` file for that very reason, to reduce the chances of them being accidentally imported.

unflxw: This was a bit of a rabbit hole -- will continue looking into it tomorrow.

First, I just checked and the editor does not seem auto-suggest these specific functions (do let me know if it does auto-complete them for you!) -- even if it does auto-suggest other internal paths, like importing `cron` from `@appsignal/nodejs/dist/check_in` (which isn't great...)

For reference, the way I'm testing this is by modifying the `express-redis` test setup to trigger autocompletions, after manually modifying its `node_modules/@appsignal/nodejs` to point to the correct location (`test-setups/nodejs/integration`) outside Docker.

It seems like whether a path and its members are suggested by the autocomplete has some relation to whether members of the path are exported (transitively? unclear) from the module's entry point (in our case, `index.ts`)

It also seems that TypeScript's autocomplete has no built-in mechanism for marking symbols not to be suggested by the autocomplete, nor does it interpret JSDoc's semantically similar markers (`@ignore`, `@internal`, `@private`) to mean that, either. Doing a little import-then-export dance doesn't seem to work either.

After reaching "ask ChatGPT about it" levels of confusion, it suggested I put the private symbols in a namespace. This actually works! But it's a big loss in readability and ergonomics.

Instead of exporting symbols, we export a namespace named `internal`, and then we need to unpack that namespace whenever we want to use it, or to re-export it at the right location. It makes quite a mess all together, and we'd need to sprinkle this all over our code.

I've done the changes needed to hide check-in related functions as a proof of concept (did not fix the tests, just the application code) so that they can only be imported in the desired way: as `import { checkIn } from "@appsignal/nodejs"`, and as `import { heartbeat } from "@appsignal/nodejs"` for the deprecated heartbeat helpers, which we should really remove before this confuses someone.

https://github.com/appsignal/appsignal-nodejs/pull/new/wip-implement-heartbeat-checkins-hide-internal

unflxw: I googled the right words this time and found [this StackOverflow answer](https://stackoverflow.com/a/65185125), which might be the correct incantation. Will give this a try later.

tombruijn: Wow that's way more work that I thought you would need to put into it. I had hoped a Ruby YARD equivalent like `@api private` would exist. Hopefully `@internal` is just that.

unflxw: Will do. While I'm at it, I'd rather get it done for the whole library (there's a lot of `NoopWhatever`s and the like that should not be suggested by autocomplete either) so I think it's best to do it in a separate PR.

unflxw: https://github.com/appsignal/appsignal-nodejs/pull/1130

PR #1130, opened 2024-09-19
Improve TypeScript suggestions

unflxw:
This is probably incomplete, as I have attempted to, in doubt, err on the side of caution and avoid removing a symbol that is potentially useful to the customer. Please point out any symbols I have missed.

### [Use @internal to improve suggestions](https://github.com/appsignal/appsignal-nodejs/commit/5a112d45cb2e25b07e41aa92dd6ef27cb23c2e9f)

TypeScript's autocomplete will suggest pretty much any symbol that
is transitively exported in your package. This includes a lot of
internal functions and objects that exist for no-op or testing
purposes and should not be used by customers.

Use the `@internal` JSDoc annotation and the `stripInternal` option
in `tsconfig.json` to avoid internal functions from being declared
in the `.d.ts` files, preventing TypeScript's autocomplete from
suggesting them.

Added on top of the changes in https://github.com/appsignal/appsignal-nodejs/pull/1123, which should be merged first. See https://github.com/appsignal/appsignal-nodejs/compare/implement-heartbeat-checkins...improve-typescript-suggestions for a good diff.

### DIFF
--- a/src/check_in/event.ts
+++ b/src/check_in/event.ts
@@ -1,8 +1,11 @@
 import type { Cron } from "./cron"
 
+/** @internal */
 export type EventKind = "start" | "finish"
+/** @internal */
 export type EventCheckInType = "cron" | "heartbeat"
 
+/** @internal */
 export class Event {
   identifier: string
   digest?: string
--- a/src/check_in/heartbeat.ts
+++ b/src/check_in/heartbeat.ts
@@ -2,16 +2,19 @@ import { stubbable } from "../utils"
 
 const HEARTBEAT_INTERVAL_MILLISECONDS = 30_000
 
+/** @internal */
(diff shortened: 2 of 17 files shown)


tombruijn (approved):
The docs for stripInternal says:

> This is an internal compiler option; use at your own risk, because the compiler does not check that the result is valid.

It's probably fine for now.

It suggest using [api-extractor](https://api-extractor.com/). Is that something we want to look at?

unflxw:
@tombruijn It's probably a good idea, yes. I did accidentally (in a now-rebased-out commit) break the `.d.ts` signatures, by marking as internal symbols some files that were re-exported in other files. The build itself does not complain, but fortunately the integration tests caught it.

PR #1134, opened 2024-10-15
Add Pino transport

luismiramirez:
A Pino transport is now exposed from the package
(`AppsignalPinoTransport`). It requires an initialized AppSignal client to work, and also accepts a group as an optional parameter.

Attributes are flattened so no data is lost.

**Usage**:

```js
import pino from "pino"
import { Appsignal, AppsignalPinoTransport } from "@appsignal/nodejs"

const logger = pino(
  AppsignalPinoTransport({
    client: Appsignal.client,
    group: "my-group",
  })
)
```

Samples from a Fastify test app:

**Simple request:**

image

**Error:**

image

_Part of #828_

### DIFF
--- a/src/pino_transport.ts
+++ b/src/pino_transport.ts
@@ -1,0 +1,78 @@
+import { LOGGER_LEVEL_SEVERITY } from "./logger"
+import build from "pino-abstract-transport"
+import { Client } from "./client"
+
+type PinoTransportOptions = {
+  client: Client
+  group: string
+}
+
+const appsignalPinoTransport = ({
+  client,
+  group = "app"
+}: PinoTransportOptions) => {
+  return build(async (source: any) => {
+    for await (const obj of source) {
+      sendLogs(parseInfo(obj, group), client)
+    }
+  })
+
+  async function sendLogs(data: Record<string, any>, client: Client) {
+    client.extension.log(
+      data.group || "app",
+      data.severity,
+      0,
+      data.msg,
+      data.attributes
+    )
+  }
+}
+
+function parseInfo(
+  obj: Record<string, any>,
+  group: string
+): Record<string, any> {
+  const { hostname, level, msg, ...attributes } = obj
+
+  return {
+    severity: getSeverity(level),
+    hostname,
+    group,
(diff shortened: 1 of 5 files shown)


luismiramirez:
Thanks @duailibe! I took inspiration from your https://github.com/duailibe/pino-appsignal-transport

src/pino_transport.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  return result
+}
+
+function getSeverity(level: number): number {
+  if (level >= 50) return 6
+  if (level >= 40) return 5
+  if (level >= 30) return 3
+  if (level >= 20) return 2
+  return 1
```
unflxw, line 97: It might help for readability to refer to our [logger level constants](https://github.com/appsignal/appsignal-nodejs/blob/main/src/logger.ts#L7-L14) -- so, instead of `return 6`, `return LOGGER_LEVEL_SEVERITY.error`, and so on.

src/pino_transport.ts
```diff
@@ -0,0 +1,100 @@
...
+  client,
+  group = "app"
+}: PinoTransportOptions) => {
+  let buffer: Record<string, any>[] = []
+  let timeout: NodeJS.Timeout | null = null
+
+  return build(
+    async (source: any) => {
+      for await (const obj of source) {
+        buffer.push(parseInfo(obj, group))
```
unflxw, line 19: Do we need to buffer the lines before sending them? I see that https://github.com/duailibe/pino-appsignal-transport does it, because it's sending an HTTP request.

But we're sending them to the extension, where the agent will already do its own buffering before sending them over the network. I think we could simplify the implementation here by removing the buffering logic and sending them to the agent straight away.

unflxw (approved):

src/pino_transport.ts
luismiramirez: I'll try, and if it works, we'll keep it. The simpler the better.

duailibe:
Glad it could be of some help!

Great work! I'll migrate from my own thing once this gets released

PR #1151, opened 2024-11-21
Fix Pino transport

unflxw:
See [Intercom conversation](https://app.intercom.com/a/inbox/yzor8gyw/inbox/admin/5246522/conversation/16410700382271#part_id=comment-16410700382271-21909579802), test setup, [docs change](https://github.com/appsignal/appsignal-docs/pull/2269/files). 

The Pino transport could not be used as a transport (that is, as initialised by `pino.transport()`, but only as a destination, and only in the main thread.

This commit makes the Pino transport communicate with the extension directly, which should work from any worker thread -- the assumption being that the extension has been started by the main thread.

This works around the fact that the client cannot be serialised and sent across threads.

This allows for the Pino transport to be used like other transports, and _alongside_ other transports:

```js
const logger = pino({
  transport: {
    targets: [
      { target: "@appsignal/nodejs/pino", options: { group: "pino" } },
      { target: ... }
    ]
  }
})
```

The `pino/index.js` file has been added as a convenience to allow for the target to be referred to as `@appsignal/nodejs/pino`, which is nicer-looking than the transport's real location inside `dist/`.


.changesets/allow-pino-transport-to-be-used-as-a-transport.md
```diff
@@ -0,0 +1,50 @@
...
+Now, it is also possible to use it as a Pino transport, with the `transport` Pino config option or the `pino.transport()` function:
+
+```js
+import pino from "pino";
+
+pino({
+  transport: {
+    target: "@appsignal/nodejs/pino",
+    options: {
+      group: "pino"
```
tombruijn, line 29:
```suggestion
      group: "application"
```

The test app uses this. If it's the log group, we use "application" for app logs.

src/pino_transport.ts
```diff
@@ -1,44 +1,47 @@
...
-  client,
-  group = "app"
-}: PinoTransportOptions) => {
+type LogData = {
+  severity: number
+  message: string
+  attributes: Record<string, any>
+}
+
+const appsignalPinoTransport = ({ group = "app" }: PinoTransportOptions) => {
```
tombruijn, line 15:
```suggestion
const appsignalPinoTransport = ({ group = "application" }: PinoTransportOptions) => {
```
Is this the default group value and should it be "application" then?

tombruijn (commented):
Will review again when the test app works.

.changesets/allow-pino-transport-to-be-used-as-a-transport.md
unflxw: Yes, it's the log group. This is just an example value in the changeset snippet -- the customer can set it to whatever value they want.

src/pino_transport.ts
unflxw: Yes, it's the default group value. I'm not sure what it should be -- [the logger itself](https://github.com/appsignal/appsignal-nodejs/blob/main/src/client.ts#L137-L147) doesn't really have a default value for the group. `application` is the default _source_. I can't find any other AppSignal loggers that have default group values.

Note that I'm preserving [existing behaviour](https://github.com/appsignal/appsignal-nodejs/pull/1151/files/11b789d924788abf97460727da18ab75c611b361#diff-4e22a816fc6ab0d9f33925bbcd1c8116a17341f0e0232931bf69f0bc281a543bL22) here -- the default value was already `"app"`, I just moved where it's being set.

luismiramirez (approved):

tombruijn (approved):

PR #1207, opened 2025-08-18
Disable distributed OpenTelemetry propagation

tombruijn:
Prevent OpenTelemetry from sending tracing headers to other applications through HTTP requests.

Use an unconfigured `CompositePropagator` to disable HTTP propagation headers while maintaining local trace correlation. It is not configured with any other propagators, so it will not add the headers to the outgoing requests.

I hope this fixes the issue with app data being reported to the wrong app, if one app makes a request to another app and they're both instrumented with OpenTelemetry.

## To do

- [ ] Inform [this customer](https://app.intercom.com/a/inbox/yzor8gyw/inbox/conversation/215469972929521?view=List) upon release.

## Resources

- [Internal private dev discussion about the change](https://appsignal.slack.com/archives/C7XHXQ7N0/p1753337899481699)

### DIFF
--- a/.changesets/disable-opentelemetry-propagation.md
+++ b/.changesets/disable-opentelemetry-propagation.md
@@ -1,0 +1,6 @@
+---
+bump: minor
+type: change
+---
+
+Disable the OpenTelemetry propagation across applications. If multiple applications are instrumented with AppSignal for Node.js using OpenTelemetry, the propagation across applications could lead to application data being reported the wrong application in AppSignal.
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@appsignal/nodejs",
-  "version": "3.6.5",
+  "version": "3.6.7",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@appsignal/nodejs",
-      "version": "3.6.5",
+      "version": "3.6.7",
       "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
--- a/src/client.ts
+++ b/src/client.ts
@@ -15,6 +15,7 @@ import {
   AggregationTemporality,
   InstrumentType
 } from "@opentelemetry/sdk-metrics"
+import { CompositePropagator } from "@opentelemetry/core"
 import { OTLPMetricExporter } from "@opentelemetry/exporter-metrics-otlp-proto"
 
 import type { Instrumentation } from "@opentelemetry/instrumentation"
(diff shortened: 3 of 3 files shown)


unflxw (approved):
I wonder/worry this might break instrumentations that use propagation for other reasons -- for example, [our own BullMQ instrumentation](https://github.com/search?q=repo%3Aappsignal%2Fopentelemetry-instrumentation-bullmq%20propagation&type=code), but also others.

It would be worth running the BullMQ instrumentation's tests (for older BullMQ versions, as the instrumentation is broken on `latest`) with this propagator configuration and seeing if it works.

unflxw:
It would seem this breaks propagation at large: adding `provider.register({propagator: new CompositePropagator()});` (which is what passing a `textMapPropagator` to the `NodeSDK` [ultimately does](https://github.com/open-telemetry/opentelemetry-js/blob/a8a641a1c7979217aa80fdcd693857e4c901d26d/experimental/packages/opentelemetry-sdk-node/src/sdk.ts#L377-L379)) [at the top of the `beforeEach` in the BullMQ instrumentation's tests](https://github.com/appsignal/opentelemetry-instrumentation-bullmq/blob/72a9b0962214735341d764bd9b18e4eb86031ccc/test/instrumentation.test.ts#L167) causes the tests that rely on propagation to fail.

This by itself could be an acceptable breakage, but other instrumentations might rely on propagation as well. We may wish to disable propagation specifically for the HTTP integration, if that's possible.

tombruijn:
> It would seem this breaks propagation at large: adding `provider.register({propagator: new CompositePropagator()});` (which is what passing a `textMapPropagator` to the `NodeSDK` [ultimately does](https://github.com/open-telemetry/opentelemetry-js/blob/a8a641a1c7979217aa80fdcd693857e4c901d26d/experimental/packages/opentelemetry-sdk-node/src/sdk.ts#L377-L379)) [at the top of the `beforeEach` in the BullMQ instrumentation's tests](https://github.com/appsignal/opentelemetry-instrumentation-bullmq/blob/72a9b0962214735341d764bd9b18e4eb86031ccc/test/instrumentation.test.ts#L167) causes the tests that rely on propagation to fail.
> 
> This by itself could be an acceptable breakage, but other instrumentations might rely on propagation as well. We may wish to disable propagation specifically for the HTTP integration, if that's possible.

That's weird. If I run the test app we have it does work with the changes in this PR.

Request timeline: 

2025-08-22 at 11 20 20 - Incident #2 - Sample - AppSignal

Worker timeline:

2025-08-22 at 11 20 36 - Incident #1 - Sample - AppSignal

PR #1228, opened 2025-09-25
Auto detect log format

tombruijn:
Change the default log format from 'plaintext' to 'autodetect'. Our processor will automatically try to detect the correct log format, which should create a better experience than the default being plaintext.

Part of https://github.com/appsignal/integration-guide/issues/174

### DIFF
--- a/src/__tests__/logger.test.ts
+++ b/src/__tests__/logger.test.ts
@@ -52,7 +52,7 @@ describe("BaseLogger", () => {
     ).toEqual(6)
   })
 
-  it("defaults to a plaintext logger format", () => {
+  it("defaults to the autodetect logger format", () => {
     expect(logger.format).toEqual(3)
     expect(client.internalLogger.warn).not.toHaveBeenCalled()
   })
--- a/src/__tests__/logger.test.ts
+++ b/src/__tests__/logger.test.ts
@@ -53,18 +53,18 @@ describe("BaseLogger", () => {
   })
 
   it("defaults to a plaintext logger format", () => {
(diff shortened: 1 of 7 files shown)


src/__tests__/logger.test.ts
```diff
@@ -53,18 +53,18 @@ describe("BaseLogger", () => {
   })
 
   it("defaults to a plaintext logger format", () => {
```
lipskis, line 55: After this change, I was a bit confused by the test name, because the expected format is `autodetect`. But then again, `autodetect` is not really a format.

lipskis (approved):

src/__tests__/logger.test.ts
tombruijn: No, this is a copy-paste mistake from me. The format autodetect is the default, even though that's not really a format as you say 😅

unflxw (commented):
Looks good so far. Also needs to be updated in the Pino and Winston transports:
- https://github.com/appsignal/appsignal-nodejs/blob/main/src/pino_transport.ts#L33
- https://github.com/appsignal/appsignal-nodejs/blob/main/src/winston_transport.ts#L79

tombruijn:
> Like in the Ruby PR: has the agent been updated here as well?

Yes. It's been updated in #1227

tombruijn:
I've updated the Pino and Winston transports. I saw no tests for the Pino transport, so I made an issue here: https://github.com/appsignal/appsignal-nodejs/issues/1243

PR #1436, opened 2026-05-24
Clear stale private config environment variables

SapphireSmith:
## Summary

Fix an issue where private AppSignal environment variables could keep stale values when a later configuration omits them.

`Configuration#writePrivateConfig` writes internal `_APPSIGNAL_*` environment variables from the merged config. Before this change, it only updated values when a current value existed, but it did not clear previously written values when a later configuration no longer provided them.

This could leave old private env vars behind across repeated initializations in the same process.

## What changed

- Clear mapped private `_APPSIGNAL_*` environment variables when the current merged config no longer has a value for them
- Clear `_APPSIGNAL_LOG_FILE_PATH` when no current log file path is resolved
- Add a regression test covering repeated initialization where the first config sets `name` and the second config omits it

## Reproduction

Before this change, this kind of sequence could leave stale state behind:

1. Initialize AppSignal with a config that sets `name`
2. Initialize AppSignal again in the same process with a valid partial config that omits `name`
3. `config.data.name` is `undefined` on the second config, but `_APPSIGNAL_APP_NAME` still keeps the old value

The new test covers this behavior and verifies the stale private env var is cleared.

## Testing

- `npx jest src/__tests__/config.test.ts --runInBand -t "clears private env values when a later config omits them"`
- `npx jest src/__tests__/config.test.ts --runInBand`

### DIFF
--- a/.changesets/clear-stale-private-env.md
+++ b/.changesets/clear-stale-private-env.md
@@ -1,0 +1,7 @@
+---
+bump: patch
+type: fix
+integrations: all
+---
+
+Clear stale private environment variables when later configurations omit them.
--- a/src/__tests__/config.test.ts
+++ b/src/__tests__/config.test.ts
@@ -392,6 +392,24 @@ describe("Configuration", () => {
       expect(env("_APPSIGNAL_APP_PATH")).toEqual(process.cwd())
     })
 
+    it("clears private env values when a later config omits them", () => {
+      new Configuration({
+        name,
+        active: true,
+        pushApiKey
+      })
+
+      expect(env("_APPSIGNAL_APP_NAME")).toEqual(name)
+
+      config = new Configuration({
+        active: true,
+        pushApiKey
+      })
+
+      expect(config.data.name).toBeUndefined()
+      expect(env("_APPSIGNAL_APP_NAME")).toBeUndefined()
+    })
+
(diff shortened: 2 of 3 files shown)


unflxw:
Hi @SapphireSmith, thank you for bringing this up. What is the scenario in which you encountered the underlying issue? Generally speaking, you're not supposed to call `Appsignal.start` more than once.

SapphireSmith:
Thanks for the question. I didn’t encounter this from a normal single-initialization application flow directly. I found it while looking through the configuration lifecycle, then reproduced it with repeated initialization in the same process.

What stood out to me is that `Configuration#writePrivateConfig` mutates global private `_APPSIGNAL_*` env vars, but before this change it only updated values when present and never cleared previously written values when a later initialization omitted them. That meant a later initialization with a valid partial config could leave stale private env state behind from an earlier one.

I agree this depends on repeated initialization, so if that is considered unsupported usage, I understand if this is not something you want to handle. I opened the PR because it looked like a small way to make the config-to-env synchronization more robust.

unflxw:
Hi @SapphireSmith, thanks for proposing this. Unfortunately we will not be merging this change at this time.

SapphireSmith:
Thanks for taking a look and for the clarification.

PR #1449, opened 2026-06-07
Ensure sendError closes spans when callbacks throw

SapphireSmith:
## Summary

Fix an issue where `sendError(error, fn)` could leave the created error span unfinished if the callback threw.

Before this change, `sendError` recorded the error, ran the callback, and only then ended the span. If the callback threw while attaching extra data, `span.end()` was never reached, so the error span could be lost.

## What changed

* Wrap the `sendError` callback execution in `try/finally`
* Always call `span.end()`, even if the callback throws
* Add a regression test covering the callback-throws case

## Reproduction

Before this change, a flow like this could leave the error span unfinished:

```ts
try {
  throw new Error("Boom")
} catch (err) {
  sendError(err as Error, () => {
    throw new Error("Callback failure")
  })
}
```

In that case, the callback error would bubble out, but the created error span would never be ended.

## Testing

```bash
npx jest src/__tests__/helpers.test.ts --runInBand -t "still ends the error span if the callback throws"

npx jest src/__tests__/helpers.test.ts --runInBand
```

### DIFF
--- a/.changesets/fix-send-error-callback-span.md
+++ b/.changesets/fix-send-error-callback-span.md
@@ -1,0 +1,7 @@
+---
+bump: patch
+type: fix
+integrations: all
+---
+
+Ensure `sendError` ends its span even when the callback throws.
--- a/src/__tests__/helpers.test.ts
+++ b/src/__tests__/helpers.test.ts
@@ -219,6 +219,31 @@ describe("Helpers", () => {
       })
     })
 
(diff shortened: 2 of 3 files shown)


unflxw (approved):

unflxw:
Thanks for the contribution, @SapphireSmith! Merged. 🎉

PR #1455, opened 2026-06-12
fix: handle BigInt serialization in helper functions

SapphireSmith:
## Summary

Fix an issue where serializing objects containing `BigInt` values crashed the host application with a `TypeError: Do not know how to serialize a BigInt`.

Before this change, helper functions like `setCustomData`, `setParams`, or `setSessionData` (which use `JSON.stringify` internally via `setSerialisedAttribute`) would crash if any property in the object was a `BigInt` because `JSON.stringify` does not natively support `BigInt` serialization.

## What changed

- Check for `bigint` type in `circularReplacer` inside `src/helpers.ts` and return its string representation using `.toString()`
- Add a regression test to `src/__tests__/helpers.test.ts` to cover objects containing `BigInt` values
- Create a patch changeset to document the fix in the changelog

## Reproduction

Before this change, a flow like this would crash the application with a `TypeError`:

```javascript
const Appsignal = require("@appsignal/nodejs");

const client = new Appsignal.Client({
  active: true,
  pushApiKey: "TEST"
});

// This throws: TypeError: Do not know how to serialize a BigInt
client.setCustomData({
  userId: BigInt("123456789012345678")
});
```

## Testing

```bash
npx jest src/__tests__/helpers.test.ts -t "handles BigInt values without throwing a TypeError"

npx jest src/__tests__/helpers.test.ts
```

### DIFF
--- a/.changesets/fix-bigint-serialization.md
+++ b/.changesets/fix-bigint-serialization.md
@@ -1,0 +1,7 @@
+---
+bump: patch
+type: fix
+integrations: all
+---
+
+Fix BigInt serialization crash in helper functions. BigInt values passed to `setCustomData`, `setParams`, or `setSessionData` no longer throw a TypeError.
--- a/src/__tests__/helpers.test.ts
+++ b/src/__tests__/helpers.test.ts
@@ -125,6 +125,18 @@ describe("Helpers", () => {
     })
   })
 
+  it("handles BigInt values without throwing a TypeError", () => {
+    trace.getTracer("test").startActiveSpan("Some span", span => {
+      setCustomData({ userId: BigInt("123456789012345678") })
+      span.end()
+    })
+
+    expect(spans.length).toEqual(1)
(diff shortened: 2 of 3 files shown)


tombruijn (approved):

SapphireSmith:
Hi @tombruijn, thank you for the approval! Just checking in to see if there's anything else needed on my end to get this merged, or if we are good to go?

tombruijn:
> Hi @tombruijn, thank you for the approval! Just checking in to see if there's anything else needed on my end to get this merged, or if we are good to go?

We need to take a moment on our end to see if this is the way we want to store this and not reuse any existing system we already have for this big int problem.

SapphireSmith:
Ah got it, yeah I saw `src/internal/data.ts` formats it as `bigint:${value}`. I can change it in the helpers to do the same if you want to keep it consistent?

PR #1497, opened 2026-08-20
Update the OpenTelemetry dependencies

unflxw:
Stacked on #1496.

Also depends on https://github.com/appsignal/appsignal-agent/pull/1294
being released. This pull request should be merged after that release.

### [Update the OpenTelemetry dependencies](https://github.com/appsignal/appsignal-nodejs/commit/6c7be872035d7f74090336f23f507b045717365d)

The `@opentelemetry/propagator-jaeger` version that resolved here
carries CVE-2026-59892, and the `@opentelemetry/core` version carries
CVE-2026-54285, which applies to the baggage propagation we do use.

From version 0.221.0 only the stable HTTP, network and database
attributes are emitted, and the agent reads the stable name first for
every attribute it needs, so these spans report as before.
`@opentelemetry/instrumentation-mongodb` also renamed its spans to
`<operation> <collection>`, which the agent read the operation from, so
this needs appsignal/appsignal-agent#1294. `@opentelemetry/core`,
`sdk-trace-base` and `sdk-metrics` are pinned exactly because `sdk-node`
pins them exactly, and any other range resolves a second copy.


package.json
```diff
@@ -12,31 +12,31 @@
...
-    "@opentelemetry/instrumentation-nestjs-core": ">= 0.59.0 < 0.60.0",
-    "@opentelemetry/instrumentation-pg": ">= 0.65.0 < 0.66.0",
-    "@opentelemetry/instrumentation-redis": ">= 0.61.0 < 0.67.0",
-    "@opentelemetry/instrumentation-restify": ">= 0.58.0 < 0.59.0",
-    "@opentelemetry/instrumentation-undici": ">= 0.23.0 < 0.24.0",
-    "@opentelemetry/sdk-metrics": ">= 2.6.0 < 2.7.0",
-    "@opentelemetry/sdk-node": ">= 0.213.0 < 0.222.0",
-    "@opentelemetry/sdk-trace-base": ">= 2.6.0 < 2.11.0",
-    "@prisma/instrumentation": ">= 7.0.0 < 7.5.0",
+    "@opentelemetry/instrumentation-graphql": ">= 0.69.0 < 0.70.0",
```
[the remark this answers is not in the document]
unflxw: This is deliberate and recorded. `@opentelemetry/instrumentation-fastify` is deprecated in favour of `@fastify/otel`, and moving to that is its own piece of work because it changes which Fastify versions are supported. The duplicate it causes has an entry in `otel_duplicates_allowed.json` in #1503, which names the reason and what would remove it, and the check fails on any duplicate that is not listed there.

.changesets/update-the-opentelemetry-dependencies.md
```diff
@@ -0,0 +1,11 @@
...
+bump: minor
+type: change
+---
+
+Update the bundled OpenTelemetry packages to their current versions. This
+resolves two reported vulnerabilities: CVE-2026-59892 in
+`@opentelemetry/propagator-jaeger`, which this integration does not use but did
+bundle through `@opentelemetry/sdk-node`, and CVE-2026-54285 in
+`@opentelemetry/core`, which applies to the W3C baggage propagation that this
+integration does use.
```
unflxw: Make each of the CVEs a bullet point for readability.

script/dump_test_app_spans
```diff
@@ -0,0 +1,67 @@
...
+# Usage: script/dump_test_app_spans <test app name> <output directory>
+# Example: script/dump_test_app_spans express-redis tmp/spans/before
+#
+# Dump each app twice, once on either side of the change, then compare the two
+# directories:
+#
+#   script/dump_test_app_spans express-redis tmp/spans/before
+#   # change the dependencies
+#   script/dump_test_app_spans express-redis tmp/spans/after
+#   node scripts/compare_span_dumps.js tmp/spans/before tmp/spans/after
```
unflxw: I'm not sure what to do with these scripts. On the one hand, the bobot did find them useful, to be able to compare before/after behaviour and confidently upgrade the OpenTelemetry integrations. On the other hand, that's what our tests are _supposed_ to be for, and right now these scripts are dead code.

This whole workflow needs to be rethought beyond these scripts. These test apps check only a fraction of the span attributes and shape expectations that we have. For span attributes, it's mostly checking them _on behalf of the agent_, which has specific logic in its extractor that depends on them. Then there's the parent-child relationships, what constitutes a trace versus what is nested into another trace, which is arguably more important to AppSignal than what span attributes are present.

There's many libraries we support for which we don't have a test app here, or we don't have one on `test-setups`, or both. It's also unclear to me how much of this would even matter in a future collector mode version of the integration, though the collector also has extractors of its own.

I'm thinking to drop this whole thing for now -- the scripts and the `ARCHIVE_SPANS` logic they add to the helpers -- and come back to the whole problem space with fresh eyes later on.

.changesets/update-the-opentelemetry-dependencies.md
unflxw: Done. Each CVE is its own bullet now.

script/dump_test_app_spans
unflxw: Dropped, all of it. The `Add a way to compare the spans a test app produces` commit is gone from this pull request, which takes `script/dump_test_app_spans`, `scripts/compare_span_dumps.js`, `Span.archive_all`, the `after(:suite)` hook and the `ARCHIVE_SPANS` passthrough in all sixteen compose files with it. The dependency bump is a separate commit and is untouched.

Agreed on the reasoning. The scripts caught the MongoDB span rename during this upgrade, but the suite caught that too, by failing on `Span.find_by_name!("mongodb.find")`. What the suite did not catch was the consequence, which is that the agent derived the operation from that name prefix. That assumption lives in a Rust file in another repository and is written down nowhere on this side, so a green suite here and a green suite there were both compatible with every MongoDB event being named wrong. That is the gap worth designing for, and it is not one these scripts close.

unflxw:
Collapsed into #1511, which carries every commit from this stack unchanged. Closing this one. The discussion here stays readable, and anything still open is worth raising again on #1511.

PR #1504, opened 2026-08-24
Warn when two @opentelemetry/api copies load

unflxw:
Stacked on #1503.

### [Warn when two @opentelemetry/api copies load](https://github.com/appsignal/appsignal-nodejs/commit/f51828146012f8f8b9443bd018910963a9f8e35b)

An application loading two incompatible copies reports only some of its
spans, because the API hands a tracer that drops every span to whichever
copy loses the version comparison, which looks the same as an
application that produces none. The check reads the module cache rather
than walking `node_modules`, since a copy nothing loaded cannot cause
this, and says it once per process because which copies are loaded does
not change while one is running.

### DIFF
--- a/.changesets/warn-about-duplicate-opentelemetry-api-copies.md
+++ b/.changesets/warn-about-duplicate-opentelemetry-api-copies.md
@@ -1,0 +1,9 @@
+---
+bump: patch
+type: add
+---
+
+Warn when your application loads more than one version of
+`@opentelemetry/api`, warning about potential data loss, as older versions
+fail to send data through global values configured by newer versions.
+The warning names the versions and where they were loaded from.
--- a/src/client.ts
+++ b/src/client.ts
@@ -9,6 +9,7 @@ import { demo } from "./demo"
 import { VERSION } from "./version"
 import { setParams, setSessionData } from "./helpers"
 import { BaseLogger, Logger, LoggerFormat, LoggerLevel } from "./logger"
+import { duplicateOpenTelemetryApiWarningOnce } from "./otel_api_copies"
 import { diag, DiagConsoleLogger, DiagLogLevel } from "@opentelemetry/api"
 import {
   PeriodicExportingMetricReader,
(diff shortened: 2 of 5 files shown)


src/client.ts
```diff
@@ -199,11 +200,28 @@ export class Client {
...
+   * handed a tracer that drops every span, which looks the same as an
+   * application that produces no spans.
+   */
+  private warnAboutDuplicateOpenTelemetryApi() {
+    const warning = duplicateOpenTelemetryApiWarning()
+    if (!warning) return
+
+    console.warn(`appsignal WARNING: ${warning}`)
+    Client.internalLogger.warn(warning)
+  }
```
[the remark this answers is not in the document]
unflxw: Leaving these separate. `consoleAndLoggerWarn` in `heartbeat.ts` is private to that module, and `heartbeat.ts` imports `Client`, so sharing it means a third module and a change to heartbeat code that is otherwise unrelated to this one. That is more indirection than two lines of duplication is worth.

.changesets/warn-about-duplicate-opentelemetry-api-copies.md
```diff
@@ -0,0 +1,9 @@
+---
+bump: patch
+type: add
+---
+
+Warn when your application loads more than one version of
+`@opentelemetry/api`. Only one of them can report spans, so some of your data
+goes missing, and until now nothing said why. The warning names the versions
+and where they were loaded from.
```
unflxw:
```suggestion
`@opentelemetry/api`, warning about potential data loss, as older versions
fail to send data through global values configured by newer versions.
The warning names the versions and where they were loaded from.
```

unflxw: Applied verbatim.

This one had slipped past me: my earlier sweep read the submitted review comments, and this thread was not among them at the time, so it only showed up when I went looking for unsubmitted reviews just now.

unflxw:
Collapsed into #1511, which carries every commit from this stack unchanged. Closing this one. The discussion here stays readable, and anything still open is worth raising again on #1511.

Issue #1494, opened 2026-08-20
Verify whether any dependencies need upgrading.


Caick-P:
We may have dependencies that need to be upgraded to avoid vulnerabilities or for general system improvement.

Context: A customer reached out concerned about vulnerabilities in our Node.js integration, especially because we pin minor versions.

[Slack thread](https://appsignal.slack.com/archives/C7XHXQ7N0/p1787230258048399).

tombruijn:
I think @unflxw is picking this up under #1511

PR #1511, opened 2026-08-31
Update the OpenTelemetry dependencies

unflxw:
**Do not merge before the next agent release.** The updated
`@opentelemetry/instrumentation-mongodb` renames its spans from
`mongodb.<operation>` to `<operation> <collection>`, and the agent read the
operation from that prefix. appsignal/appsignal-agent#1294 fixes that and is
merged, but the newest agent release is 0.36.12 from June, and this package
pins it. Merging before an agent release that contains it would leave MongoDB
events unnamed.

This replaces the stack of #1495, #1496, #1497, #1502, #1503 and #1504, which
are closed in favour of this one. The commits and their messages are unchanged;
review them one at a time and the order still reads as a single argument.

### [Remove the global `BlobPropertyBag` declaration](https://github.com/appsignal/appsignal-nodejs/commit/bf6d15cea811c5302b50cf0477b3a9a2920d767b)

The declaration stood in for a DOM type the `undici` typings in
`@types/node` no longer refer to. From `@types/node` 25 it breaks the
build, because that version declares `BlobPropertyBag` itself and
TypeScript reports the two as a conflict. The committed lock file pins
`@types/node` 22, so only an install that resolves versions afresh runs
into it.

### [Stop sending the trace context in B3 headers](https://github.com/appsignal/appsignal-nodejs/commit/3c70393ad34b1a608862a906cef172fdc7cc8f70)

Two applications that both report to AppSignal join the same trace when
the trace context travels between them, and the second one's data is
reported against the first one's application. The B3 propagators were
sending it in the `b3` and `x-b3-*` headers. Baggage still propagates,
and the only loss is the span link between a BullMQ publish and the job
that processes it, which the agent does not read.

### [Update two test apps' `@opentelemetry/api` bound](https://github.com/appsignal/appsignal-nodejs/commit/20f9a27eec1034baf28ce7200095f38621265ff4)

`express-redis` and `express-rabbitmq` are the only test apps whose own
code creates spans, so they are the only ones that declare
`@opentelemetry/api` themselves, and their bound was still 1.6. Each
resolved one copy for itself and a second inside the linked integration.
Their spans arrived anyway, because an older copy of the API may read a
global registered by a newer one, so the wrong way round would have
failed both suites with no indication of why.

### [Check for duplicate `@opentelemetry/api` copies](https://github.com/appsignal/appsignal-nodejs/commit/ef892eb2ab149c13cc4d0d994916f7a5e67e2a6d)

Two copies of `@opentelemetry/api` in one project make the integration
report nothing and log nothing, because the API only reads its global
tracer provider from a copy whose version is compatible, and the copy
that loses is handed a tracer that discards every span. Whether that
happens is decided by npm's resolution rather than by anything in the
source, so the check installs the packed package into a throwaway
project with an empty npm cache: `npm link` and our own lock file both
change the answer, and a warm cache can serve a resolution a fresh
install would not pick.

### [Stop Dependabot from widening version ranges](https://github.com/appsignal/appsignal-nodejs/commit/8c4f7bd7090718052719dddecdd5be775be7c92f)

Dependabot widens the range for a library by default, so it turned the
`@opentelemetry/sdk-node` bound into `>= 0.213.0 < 0.219.0` and let npm
resolve a different `@opentelemetry/core` for `sdk-node` than for the
rest of the tree. These bounds are narrow so that packages released
together resolve to one version of each other, and they are grouped
because they are only correct as a set.

### [Widen the `@opentelemetry/api` version bound](https://github.com/appsignal/appsignal-nodejs/commit/04bfbe91eb622e089fc0ec296282154de659c6a4)

Every OpenTelemetry package depends on the API as a peer dependency and
none as a regular one, so the installed version is whatever they all
agree on, and a bound of ours can only narrow that agreement. It becomes
the binding one as soon as we move to an SDK release accepting a newer
API, and a bound left behind then nests a second copy in applications
that resolve the newer version, which the resolution check cannot see
because it installs this package on its own.

### [Update the OpenTelemetry dependencies](https://github.com/appsignal/appsignal-nodejs/commit/5a8d444b6964ac309711dfd35b2f98def55d8964)

The `@opentelemetry/propagator-jaeger` version that resolved here
carries CVE-2026-59892, and the `@opentelemetry/core` version carries
CVE-2026-54285, which applies to the baggage propagation we do use.

From version 0.221.0 only the stable HTTP, network and database
attributes are emitted, and the agent reads the stable name first for
every attribute it needs, so these spans report as before.
`@opentelemetry/instrumentation-mongodb` also renamed its spans to
`<operation> <collection>`, which the agent read the operation from, so
this needs appsignal/appsignal-agent#1294. `@opentelemetry/core`,
`sdk-trace-base` and `sdk-metrics` are pinned exactly because `sdk-node`
pins them exactly, and any other range resolves a second copy.

### [Drop support for Node.js 18](https://github.com/appsignal/appsignal-nodejs/commit/e8d6e866f4df28191ea40a71103a60cf9a4a290d)

Node.js 18 reached end of life in April 2025 and CI stopped testing it
in January 2026, while the `engines` range still claimed to support it.
The floor is 20.6.0 rather than 20 because that is what the bundled
OpenTelemetry packages require, and a plain `>= 20` would leave the same
silent gap between 20.0 and 20.5 that `>= 18` left below 18.19.

### [Require `pino-abstract-transport` v3](https://github.com/appsignal/appsignal-nodejs/commit/cd5ddd17b00a982696fd5265d673360f77b77a5a)

Version 2 was in the accepted range only so that Node.js 18 users had a
resolvable version, and the floor is now 20.6.0. Accepting a single
major also lets npm share one copy with the rest of an application's
tree instead of installing a duplicate.

### [Fail the check on unexplained duplicates](https://github.com/appsignal/appsignal-nodejs/commit/a7e7dad0c0d1931454f05c3ccd4d23f9e2c558e9)

Duplicates of every OpenTelemetry package other than the API were listed
and not acted on, which is where a mistake in the version bounds shows
up first. They now fail unless `otel_duplicates_allowed.json` accounts
for them, and an entry that stops matching is reported so the file does
not keep reasons that have stopped being true.

Which dependency causes a duplicate is read from where npm put it, since
npm nests a copy inside whichever dependency asked for a version the
hoisted copy cannot satisfy. The check runs on Node.js 20 and 24 because
they ship npm 10 and npm 11, whose peer dependency resolution differs.

### [Check the test apps for duplicate API copies](https://github.com/appsignal/appsignal-nodejs/commit/0c6639f8daceeeb3e852c0446d8b89ea35bfc3a1)

A test app resolving two incompatible copies of `@opentelemetry/api`
reports nothing, and its suite fails on an empty spans file without
saying why. The apps link the integration rather than installing it, so
`check_otel_resolution.js` cannot stand in for this, and copies that
differ only in their patch version are accepted because the API compares
and minor versions and linking produces such copies routinely.

### [Warn when two @opentelemetry/api copies load](https://github.com/appsignal/appsignal-nodejs/commit/af89a092ad6556e5badcd4760ae3a4ef3520d0c8)

An application loading two incompatible copies reports only some of its
spans, because the API hands a tracer that drops every span to whichever
copy loses the version comparison, which looks the same as an
application that produces none. The check reads the module cache rather
than walking `node_modules`, since a copy nothing loaded cannot cause
this, and says it once per process because which copies are loaded does
not change while one is running.

### DIFF
--- a/.changesets/stop-sending-b3-trace-context-headers.md
+++ b/.changesets/stop-sending-b3-trace-context-headers.md
@@ -1,0 +1,8 @@
+---
+bump: patch
+type: fix
+---
+
+Do not emit B3 trace context propagation headers. If you relied on AppSignal to
+send the trace context in the `b3` or `x-b3-*` headers, you can configure that
+yourself by passing your own `textMapPropagator` to the OpenTelemetry SDK.
--- a/global.d.ts
+++ b/global.d.ts
@@ -3,5 +3,4 @@ import { Client } from "./src/client"
 declare global {
   // eslint-disable-next-line no-var
   var __APPSIGNAL__: Client
-  type BlobPropertyBag = unknown
 }
--- a/src/__tests__/client.test.ts
+++ b/src/__tests__/client.test.ts
@@ -7,6 +7,12 @@ import { Client } from "../client"
 import { Metrics } from "../metrics"
 import { NoopInternalLogger, NoopLogger, NoopMetrics } from "../noops"
 import { Instrumentation } from "@opentelemetry/instrumentation"
+import {
+  propagation,
+  trace,
+  ROOT_CONTEXT,
+  TraceFlags
+} from "@opentelemetry/api"
 import { BaseInternalLogger } from "../internal_logger"
 import { BaseLogger } from "../logger"
 
@@ -31,6 +37,31 @@ describe("Client", () => {
     const startSpy = jest.spyOn(Extension.prototype, "start")
     client = new Client({ ...DEFAULT_OPTS, active: true })
     expect(startSpy).toHaveBeenCalled()
+  })
+
+  it("propagates baggage, and nothing else", () => {
+    client = new Client({ ...DEFAULT_OPTS, active: true })
+
+    const context = trace.setSpanContext(
+      propagation.setBaggage(
+        ROOT_CONTEXT,
+        propagation.createBaggage({ user_id: { value: "123" } })
+      ),
+      {
+        traceId: "d4cda95b652f4a1592b449d5929fda1b",
+        spanId: "6e0c63257de34c92",
+        traceFlags: TraceFlags.SAMPLED
+      }
+    )
+
+    const carrier: Record<string, string> = {}
+    propagation.inject(context, carrier)
+
+    // A trace context header here means two applications that both report to
+    // AppSignal can end up sharing a trace. `traceparent` was disabled on
+    // purpose; the B3 headers went on sending the trace context anyway, for a
+    // year, because they were configured by accident.
+    expect(Object.keys(carrier).sort()).toEqual(["baggage"])
   })
 
   it("stops the extension when the client is stopped", async () => {
--- a/src/client.ts
+++ b/src/client.ts
@@ -16,7 +16,6 @@ import {
   InstrumentType
 } from "@opentelemetry/sdk-metrics"
 import { CompositePropagator } from "@opentelemetry/core"
-import { B3Propagator, B3InjectEncoding } from "@opentelemetry/propagator-b3"
 import { W3CBaggagePropagator } from "@opentelemetry/core"
 import { OTLPMetricExporter } from "@opentelemetry/exporter-metrics-otlp-proto"
 
@@ -426,12 +425,11 @@ export class Client {
       instrumentations,
       spanProcessor,
       metricReader,
+      // Propagate baggage, but not the trace context. Two applications that
+      // both report to AppSignal must not end up sharing a trace, or the
+      // second one's data is reported against the first one's application.
       textMapPropagator: new CompositePropagator({
-        propagators: [
-          new W3CBaggagePropagator(),
-          new B3Propagator(),
-          new B3Propagator({ injectEncoding: B3InjectEncoding.MULTI_HEADER })
-        ]
+        propagators: [new W3CBaggagePropagator()]
       })
     })
 
--- a/test/express-rabbitmq/app/package-lock.json
+++ b/test/express-rabbitmq/app/package-lock.json
@@ -10,7 +10,7 @@
       "license": "ISC",
       "dependencies": {
         "@appsignal/nodejs": "^3.7.3",
-        "@opentelemetry/api": ">= 1.6.0 < 1.7",
+        "@opentelemetry/api": "^1.9.0",
         "amqplib": "^0.10.3",
         "cookie-parser": "^1.4.7",
         "express": "^4.22.1",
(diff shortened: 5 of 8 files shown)


src/otel_api_copies.ts
```diff
@@ -0,0 +1,111 @@
...
+    .join("\n")
+
+  return (
+    "More than one version of @opentelemetry/api is loaded, so some spans " +
+    "will not be reported and nothing else will say so:\n" +
+    `${listed}\n` +
+    "Make your application and its dependencies agree on one version of " +
+    "@opentelemetry/api. If a dependency asks for a version that cannot be " +
+    "shared, updating it may be enough, and your package manager can force a " +
+    "single version if it is not."
```
unflxw:
```suggestion
    "Your application and its dependencies must agree on a single version of " +
    "@opentelemetry/api. If a dependency asks for a version that cannot be " +
    "shared, updating it may be enough. Your package manager can also" +
    "override it to a single version."
```

Might be nice to mention `npm list @opentelemetry/api` as a tool to debug this,
though it's not package-manager neutral. In practice this advice doesn't need to
be package-manager neutral because this shouldn't happen in pnpm/yarn.

src/otel_api_copies.ts
```diff
@@ -0,0 +1,111 @@
...
+  if (seen.size < 2) return undefined
+
+  const listed = copies
+    .map(copy => `  ${copy.version} in ${copy.path}`)
+    .sort()
+    .join("\n")
+
+  return (
+    "More than one version of @opentelemetry/api is loaded, so some spans " +
+    "will not be reported and nothing else will say so:\n" +
```
unflxw:
```suggestion
    "More than one version of @opentelemetry/api is loaded. This will cause " +
    "some spans to not be reported:\n" +
```

Less narrative.

tombruijn (approved):

script/check_otel_duplicates.js
tombruijn, line 1: These scripts are included in the package we ship. Is that intentional? 

https://github.com/appsignal/appsignal-nodejs/blob/f51828146012f8f8b9443bd018910963a9f8e35b/.npmignore#L4-L4

unflxw: Thanks for spotting that -- that's a mistake carried from a previous iteration, in which I wanted to ship these as post-install npm checks.

Turns out that, on failure, post-install npm checks just undo _everything_ that npm was doing, and it's a pretty poor user experience if it ever gets triggered, so I went with the runtime check at initialization time (see `src/otel_api_copies.ts`) instead.

Will fix.

unflxw:
Fixes #1494.

src/otel_api_copies.ts
unflxw: Applied, with one space added: `"...can also" + "override it..."` concatenated to `alsooverride`.

I took up the `npm list @opentelemetry/api` idea as its own sentence, between the requirement and the two remedies. It does name npm specifically, so it is easy to drop if we would rather stay neutral.

src/otel_api_copies.ts
unflxw: Applied as suggested.

script/check_otel_duplicates.js
unflxw: Fixed by moving the four OpenTelemetry check files from `scripts/` to `script/`, which the package does not ship. Only their path changed.

`scripts/extension/` stays, because it does have to ship. `package.json` runs it from `preinstall` and `install`, and `src/extension_wrapper.ts`, `src/diagnose.ts` and `src/version.ts` import from it. That is what the `!/scripts/**/*` on line 4 is for.

`npm pack --dry-run` now lists only `scripts/extension/*`.

